PR #5, opened 2026-04-22
MLIR Frontend PR 1 — Decouple parser from dialect-specific logic

### LINKED ISSUE #4: [Refactor] Decouple parser.py from dialect-specific logic
## Refactor

Remove all hardcoded dialect names from `parser.py` so that adding new
region-opening or multi-line ops requires no changes to the parser.

## Motivation

`parser.py` hardwires knowledge of specific dialects in two places:

**`_line_opens_region` (parser.py:364–367)** — hardwires `scf.for` and
`scf.if` as the only ops that introduce a `{ ... }` region body. Adding any
new region-opening op requires editing the parser.

**`_is_new_operation` (parser.py:320–323)** — hardwires continuation-line
prefixes for `ktdp` and `linalg` ops:

```python
if stripped_line.startswith((
    'sizes:', 'strides:', 'base_map', 'access_tile_set',   # ktdp
    'ins(', 'outs(', 'dimensions', 'permutation',           # linalg
)):
    return False
```

Adding any multi-line op from a new dialect requires editing this tuple. Both
violations contradict the open/closed principle already embodied by
`dialects/registry.py`.

## What can be implemented

Both hardcoded checks were replaced with **structural detection** that requires
zero dialect-specific knowledge — no registry metadata, no per-op annotations.

**Multi-line operation grouping** — `_is_new_operation` is replaced by
`_is_op_complete`, which asks whether the accumulated op is *complete* rather
than whether the next line looks like a continuation. Three structural signals
drive flushing:

- **Type-terminal completeness**: an op is complete when its accumulated text
  ends with a type annotation (`: <type>` or `-> <type>`) or a void terminator
  (`return`, `scf.yield`).
- **Blank lines**: a blank line after balanced-brace content always flushes.
- **SSA assignment**: an incoming `%name = ` line unambiguously starts a new op.

This eliminates the continuation-prefix list entirely. A new multi-line op like
`construct_indirect_access_tile` with `intermediate_variables(...)` continuation
lines just works — the accumulated text has no type annotation yet, so lines keep
accumulating until it does.

**Region detection** — `_line_opens_region` is replaced by a structural
`%`-peek: scan inside the `{ }` block and return `True` if any inner line
contains a `%` SSA reference. Attribute blocks (`{ key = value, ... }`) never
contain `%`; region blocks (`{ %x = arith.addf ... }`) always do. This handles
`scf.for`, `scf.if`, `linalg.generic`, and `linalg.reduce` (explicit-region
form) without any op-specific knowledge.

## Changes

| File | Change |
|------|--------|
| `ktir_cpu/parser.py` | Replace `_is_new_operation` with `_is_op_complete`; replace hardcoded `_line_opens_region` with structural `%` peek; add KTIRParser design docstring |
| `ktir_cpu/dialects/linalg_ops.py` | Document linalg.reduce shorthand vs explicit-region forms; executor resolves combiner from `op.regions` when present |
| `ktir_cpu/interpreter.py` | Unknown operations raise `ValueError` instead of silent skip |
| `tests/test_interpreter.py` | Update unknown-op test to expect `ValueError` |

## Pitch

- [x] Replace `_is_new_operation` with `_is_op_complete` (type-terminal + blank-line + SSA signals)
- [x] Replace hardcoded `_line_opens_region` with structural `%`-peek inside `{ }`
- [x] Add KTIRParser design docstring to `parser.py`
- [x] All existing tests pass; new ops with arbitrary continuation structure or region bodies work without parser changes

## Additional context

No dependency cycle is introduced. The dependency graph stays:

```
parser.py  →  dialects/registry.py  ←  scf_ops.py
                                    ←  linalg_ops.py
                                    ←  ktdp_ops.py
```

Dialect modules never import from `parser.py`.

cc @mudhakar @raghukiran1224

### LINKED ISSUE #6: [Feature] MLIR-backed parser frontend
## Feature

Add `MLIRFrontendParser` as a swappable alternative to the regex-based
`KTIRParser`, using the MLIR Python bindings for parsing and verification.

## Motivation

The regex parser works for shipped examples but is costly to maintain: every
new MLIR syntax variation requires a regex update, and parse errors are
hard to diagnose. The MLIR Python bindings provide authoritative
parse/verify/walk — if MLIR accepts it, the frontend parser handles it.

This also lays groundwork for consuming the upstream C++ dialect extension
at https://github.com/torch-spyre/ktir-mlir-frontend as a Python package.

## Pitch

- [x] Decouple parser tokenizer from dialect-specific logic (structural
  detection instead of hardcoded keyword lists)
- [x] Introduce `KTIRParserBase` ABC and `parser:` parameter on
  `KTIRInterpreter` for dependency injection of alternative backends
- [x] Refactor test suites into mixin architecture (`ParseTestMixin`,
  `InterpreterTestMixin`) so any backend can reuse all existing tests
- [ ] Implement `MLIRFrontendParser` in `ktir_cpu/mlir_frontend/` with
  `MLIRTypeAdapter` (30+ per-op handlers) using `mlir_ktdp.ir`
- [ ] Add adapter test suites: 5 parse-adapter + 8 execution-adapter
  classes that inherit all existing tests
- [ ] Consume https://github.com/torch-spyre/ktir-mlir-frontend as a
  Python package (provides `mlir_ktdp` and `tools_ktdp`)

## Open questions

- **Auto mode**: should `KTIRInterpreter()` auto-detect and prefer the
  MLIR frontend when `mlir_ktdp` is available, or should the caller
  always choose explicitly?
- **Package dependency**: how should `ktir-mlir-frontend` be wired in —
  as an optional extra (`pip install ktir_cpu[mlir]`) or a hard
  dependency?

## Additional context

- Implementation branches on https://github.com/fabianlim/ktir-mlir-frontend:
  `decouple-parser-from-dialects`, `parser-test-mixins`, `mlir-bindings-parser`
- First two PRs are made, other PR temporarily here https://github.com/fabianlim/ktir-cpu/pull/2
- MLIR frontend is optional — tests auto-skip when `mlir_ktdp` is not
  installed

cc: @mudhakar @raghukiran1224 @kiszk

fabianlim:
Closes #4  and addresses #6   cc: @mudhakar @raghukiran1224  @kiszk 

## Roadmap

> MLIR frontend parser series: add `MLIRFrontendParser` as a swappable
> alternative to the regex-based `KTIRParser`.

**PR 1 — Decouple parser**: Make the tokenizer and region detector fully
structural so they require zero dialect-specific metadata. Pure
refactoring — no new API, no new features.

**PR 2 — Test mixins + KTIRParserBase ABC**: Introduce the
`KTIRParserBase` abstract base class and refactor both test suites to use
mixins that decouple test assertions from parser implementation. This is
the infrastructure that enables any future parser backend to reuse all
existing tests.

**PR 3 — MLIRFrontendParser**: Implement `MLIRFrontendParser` in
`ktir_cpu/mlir_frontend/` using `mlir_ktdp.ir` structural walk with
`MLIRTypeAdapter` (30+ per-op handlers). Add adapter test suites that
reuse all existing tests via the mixin infrastructure from PR 2.

---

- **→ PR 1 — Decouple parser from dialect-specific logic**
- PR 2 — Test mixins + KTIRParserBase ABC
- PR 3 — MLIRFrontendParser + adapter test suite

## Goal

Make the KTIR parser fully dialect-agnostic so that adding new ops never
requires touching the tokenizer or region detector. This is a prerequisite
for the MLIR frontend parser: both backends must agree on how to split MLIR text
into ops and regions, and structural detection eliminates the need for
dialect-specific metadata.

## Background

The tokenizer previously used `_is_new_operation` with a hardcoded list of
continuation prefixes (`sizes:`, `strides:`, `ins(`, `outs(`, `dimensions`,
...). Every new multi-line op required adding its continuation keyword.
Similarly, `_line_opens_region` hardcoded op names that introduce regions.

### `ktir_cpu/parser.py` — structural tokenization and region detection

Replace keyword-based mechanisms with structural detection:

**Multi-line operation grouping** — flip the question from "does the next
line look like a continuation?" to "is the accumulated op complete?". Three
structural signals drive flushing:

- **Type-terminal completeness** (`_is_op_complete`): an op is complete
  when its accumulated text ends with `: <type>` or `-> <type>`, or is a
  void terminator (`return`, `scf.yield`)
- **Blank lines**: a blank line after balanced-brace content always flushes
- **SSA assignment**: an incoming `%name = ` line unambiguously starts a
  new op

This eliminates the hardcoded continuation-prefix list (`sizes:`,
`strides:`, `ins(`, `outs(`, `dimensions`, ...). A new multi-line op like
`construct_indirect_access_tile` with `intermediate_variables(...)`
continuation lines just works — the accumulated text has no type annotation
yet, so lines keep accumulating until it does.

**Region detection** — peek into `{ }` blocks and check for `%` SSA
references. Attribute blocks (`{ key = value }`) never contain `%`; region
blocks (`{ %x = arith.addf ... }`) always do. This handles `scf.for`,
`scf.if`, `linalg.generic`, and `linalg.reduce` (explicit-region form)
without any op-specific knowledge.

**Design docstring** — add KTIRParser 3-phase design overview with explicit
extension rules:

1. Do **not** add operation-specific keywords to the tokenizer
2. Do **not** hardcode op names for region detection

| Old (dialect-coupled) | New (structural) |
|---|---|
| New multi-line op → add continuation prefix | Just works (type-terminal detection) |
| New region-opening op → add to hardcoded list | Just works (`%` heuristic) |

### `ktir_cpu/dialects/linalg_ops.py` — combiner resolution from regions

The structural region detector now correctly extracts the `{ arith.maxnumf }`
body of explicit-region `linalg.reduce` as an `op.regions` entry. Update the
executor to resolve the combiner from `op.regions` when present, instead of
relying solely on the shorthand `{ arith.maxnumf }` attribute string.

This is a direct consequence of the decoupling: the old code treated
`{ arith.maxnumf }` as an inline attribute because `_line_opens_region`
did not know about `linalg.reduce`. The structural `%`-peek correctly
classifies it as a region when the body contains SSA references, and as an
attribute when it does not (shorthand form). The executor must handle both.

## Side-fixes included in this PR

These are small fixes found during testing of the structural changes. They
are included here because they exercise the new tokenizer paths and confirm
the decoupling works end-to-end.

### `ktir_cpu/interpreter.py` — unknown ops raise `ValueError`

Previously, unknown ops printed a warning and continued silently. This
masked spec-gap failures. Changed to `ValueError` so that `xfail(strict=True)`
tests in `test_spec_gaps.py` fail deterministically. Included here because
the structural region detector changes which ops are "seen" by the
interpreter (e.g. region bodies are now correctly extracted), so the error
behaviour must be well-defined.

### `ktir_cpu/dialects/ktdp_ops.py` + `examples/triton-ktir/indexed_add.mlir` — whitespace fix

The `construct_indirect_access_tile` parser required
`intermediate_variables(` with no space before the paren. Compiler output
sometimes emits `intermediate_variables (` with a space. Fixed the regex
to `intermediate_variables\s*\(`. Included here because the new
type-terminal tokenizer correctly joins these lines (validating the
decoupling), but the dialect parser then failed on the space.

### `ktir_cpu/dialects/registry.py` — remove unused import

Dead code cleanup.

### `tests/test_interpreter.py` — update unknown-op tests

Follows from the `ValueError` change above.

## Changes per file

| File | Category | Change |
|------|----------|--------|
| `ktir_cpu/parser.py` | **core** | Replace `_is_new_operation` with `_is_op_complete`; replace hardcoded `_line_opens_region` with structural `%` peek; add 3-phase design docstring |
| `ktir_cpu/dialects/linalg_ops.py` | **core** | Combiner resolution from `op.regions`; document two reduce forms |
| `ktir_cpu/interpreter.py` | side-fix | Unknown ops raise `ValueError` |
| `ktir_cpu/dialects/ktdp_ops.py` | side-fix | `intermediate_variables\s*\(` whitespace fix |
| `examples/triton-ktir/indexed_add.mlir` | side-fix | Whitespace fix to match compiler output |
| `ktir_cpu/dialects/registry.py` | side-fix | Remove unused import |
| `tests/test_interpreter.py` | side-fix | Update unknown-op tests for `ValueError` |

## Test plan

- [ ] `indexed_add.mlir` (construct_indirect_access_tile with `intermediate_variables (...)`) parses correctly
- [ ] `paged_attention.mlir` regression check (indirect access inside scf.for)
- [ ] `reduce_generic.mlir` explicit-region linalg.reduce parses and executes
- [ ] Unknown ops raise `ValueError` (test_scf_parallel xfail strict)

### DIFF
--- a/ktir_cpu/parser.py
+++ b/ktir_cpu/parser.py
@@ -29,10 +29,35 @@ from .parser_utils import parse_attr_block, parse_tensor_type, parse_numeric
 class KTIRParser:
     """KTIR MLIR parser.
 
-    Parses compiler-generated KTIR including multi-line operations
-    (e.g. ktdp.construct_memory_view spanning 3-4 lines), nested
-    regions (scf.for loop bodies), and compiler-specific constant
-    syntax like {1823 : index} or {dense<0.0>}.
+    Parsing proceeds in three phases:
+
+    1. **Module-level pre-scan** — collects named attribute aliases
+       (``#name = value``) so that dialect parsers can resolve ``#name``
+       references during op parsing.
+
+    2. **Tokenization** (``_tokenize_operations``) — splits the body
+       text of a function or region into complete operation strings.
+       Multi-line ops are joined, and ``{ }`` blocks are classified
+       structurally: blocks whose content contains ``%`` SSA
+       references are extracted as **regions** (recursively parsed);
+       all others are kept as inline **attribute blocks**.
+
+    3. **Op dispatch** — each operation string is dispatched to a
+       dialect-registered parser (``register_parser`` in
+       ``dialects/registry.py``).  Unrecognised ops fall through to
+       the general-purpose parser.
+
+    To add support for a new operation, register a parser in the
+    appropriate dialect module — the tokenizer and dispatcher are
+    dialect-agnostic and should not require changes.  Specifically:
+
+    - Do **not** add operation-specific keywords to the tokenizer
+      (e.g. continuation-line prefixes).  Multi-line grouping relies
+      on type-terminal detection, blank-line boundaries, and SSA
+      assignment signals — all of which are structural.
+    - Do **not** hardcode op names for region detection.  Regions are
+      detected by the ``%`` heuristic; any ``{ }`` block that
+      contains SSA references is automatically treated as a region.
     """
 
     def parse_file(self, filepath: str) -> IRModule:
--- a/ktir_cpu/parser.py
+++ b/ktir_cpu/parser.py
@@ -250,38 +250,56 @@ class KTIRParser:
             line = lines[i]
             stripped = line.strip()
 
-            # Skip empty lines and comments
+            # A blank line (or comment) separates operations.  Flush
+            # the accumulated op if braces are balanced.
             if not stripped or stripped.startswith('//'):
-                # If we have accumulated lines for an operation, a blank line
-                # might separate it. But we should only flush if the accumulated
-                # op looks complete.
+                if current_op_lines:
+                    accumulated = ' '.join(current_op_lines)
+                    open_braces = accumulated.count('{') - accumulated.count('}')
+                    if open_braces == 0:
+                        results.append((accumulated, current_regions))
+                        current_op_lines = []
+                        current_regions = []
                 i += 1
                 continue
 
-            # Only flush when the accumulated op text has balanced braces —
-            # an open { ... } attribute block must stay on one logical op.
-            # parse_attr_block uses the same depth-counting loop internally,
-            # so we reuse it here: a balanced block means it found a closing '}'.
-            # NOTE: underlying assumption is that the KTIR is well-formed with no 
-            #       stray braces.
+            # Flush when the accumulated op is structurally complete
+            # (type annotation or void terminator) with balanced braces,
+            # OR when the incoming line is an SSA assignment (%name = )
+            # which unambiguously starts a new operation.
+            #
+            # Type-terminal flush (adjacent ops without blank line):
+            #   %c0 = arith.constant 0 : index   <- complete (: index)
+            #   %c1 = arith.constant 1 : index   <- new op
+            #
+            # SSA flush (prev op has no type terminal, e.g. dimensions=):
+            #   linalg.broadcast ... dimensions = [1]   <- no type
+            #   %next = arith.subf ...                  <- SSA triggers flush
+            #
+            # Multi-line ops stay together until complete:
+            #   %t = construct_indirect_access_tile   <- no type yet
+            #       intermediate_variables(...)        <- continuation
+            #       %view[...] { ... } : ... -> type  <- complete
             accumulated = ' '.join(current_op_lines)
             open_braces = accumulated.count('{') - accumulated.count('}')
-            # Don't flush when accumulated ends with '=' — the next line is
-            # a continuation (e.g. multi-result: %a, %b, %c =\n  scf.for ...)
-            ends_with_eq = accumulated.rstrip().endswith('=')
-            if self._is_new_operation(stripped) and current_op_lines and open_braces == 0 and not ends_with_eq:
-                # Flush the previous operation
-                results.append((accumulated, current_regions))
-                current_op_lines = []
-                current_regions = []
+            # A line starting with '->' is always a result-type
+            # continuation (e.g. `: input_types\n  -> result_type`).
+            if current_op_lines and open_braces == 0 and not stripped.startswith('->'):
+                starts_ssa = re.match(
+                    r'(?:%\w+\s*,\s*)*%\w+\s*=\s', stripped
+                )
+                if self._is_op_complete(accumulated) or starts_ssa:
+                    results.append((accumulated, current_regions))
+                    current_op_lines = []
+                    current_regions = []
 
             current_op_lines.append(stripped)
 
-            # Check if this line opens a region body (scf.for, scf.if).
-            # A region starts when a line ends with '{' and the operation
-            # is a control flow op (scf.for, scf.if).
-            # We need to find the matching '}' across subsequent lines.
-            if self._line_opens_region(stripped, current_op_lines):
+            # Check if this line opens a region body.
+            # A region { } is distinguished from an attribute { } by
+            # peeking inside: regions contain %SSA references,
+            # attribute blocks do not.
+            if self._line_opens_region(stripped, lines, i):
                 # The '{' at the end of this line opens a region.
                 # Remove the trailing '{' from the op text.
                 current_op_lines[-1] = stripped.rstrip('{').rstrip()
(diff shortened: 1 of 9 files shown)


ktir_cpu/parser.py
```diff
@@ -305,76 +348,54 @@ def _tokenize_operations(self, body_text: str) -> List[Tuple[str, List[str]]]:
...
-        # linalg.generic: the region opens on the outs(...) { line, not on
-        # the first { which is the inline attribute block.
-        if 'linalg.generic' in op_text and 'outs(' in op_text:
-            return True
+        # Peek into the { } block — if any line contains %, it's a region.
+        depth = 1
+        for j in range(line_idx + 1, len(lines)):
+            inner = lines[j]
+            depth += inner.count('{') - inner.count('}')
+            if '%' in inner:
```
kiszk: Does it work if `%` in comment (e.g. `// 90% of execution time is consumed`)

ktir_cpu/parser.py
```diff
@@ -305,76 +348,54 @@ def _tokenize_operations(self, body_text: str) -> List[Tuple[str, List[str]]]:
...
-        - scf.for, scf.if, scf.yield
-
-        Continuation lines start with:
-        - sizes:, strides:, base_map, access_tile_set
-        - : (type annotation on its own line)
-        - } (closing brace of attribute block)
+    # Regex matching a terminal type token at end of op text.
+    # Covers: tensor<...>, memref<...>, !ktdp.access_tile<...>,
+    #         index, i32, f16, f32, etc.
+    _TYPE_TERMINAL_RE = re.compile(
```
kiszk, line 370: Does this regular expression support nested `<>`? For example,
```
%t = tensor.empty() : tensor<128x256xf16, #sparse_tensor.encoding<{map = (d0, d1) -> (d0 : dense, d1 : compressed)}>>
```

kiszk (dismissed):
Leaving two minor comments, but lgtm in general.

ktir_cpu/parser.py
fabianlim: @kiszk for the catch. It turns out you are right, the regex was not robust enough to detect this edge case, to address this we did 2 things:
1. add a preprocessing step upfront to strip the whole text for any comments
2 improve the regex to capture SSA's better by also matching alphanumeric context after `%`

ktir_cpu/parser.py
fabianlim: Yes it turns out it was robust against this, we have added new tests such as `test_type_terminal_nested_angle_brackets` to prove its robustness. These tests are added in a new section covering other cases for matching the terminal type

fabianlim:
@kiszk thanks for the review and catching one bug. I have added a new suite of tests for the `_is_op_complete` heuristic, now covering all 3 sub-heuristics used by the regex parser to determine if an op has terminated. Hope to get another approval

kiszk (approved):
lgtm, thank you.

PR #8, opened 2026-04-22
[Fix] Centralize dtype handling in dtypes.py; fix byte-size bugs for non-f16 dtypes

WarningRan:
## Problem

`TileRef.size_bytes()` was hardcoding `2 if dtype == "f16" else 1`, returning 1 byte/elem
for i32, i64, f32, and mxfp8. Investigating the fix revealed a broader issue: dtype
handling was duplicated across 8+ locations, with several containing real bugs:

| Location | Bug |
|---|---|
| `ops/memory_ops.py` tile_access | i64 → 4 bytes instead of 8 (wrong pointer offset) |
| `dialects/arith_ops.py` arith.constant | f32 → `np.int32` (wrong array dtype) |
| `dialects/tensor_ops.py` tensor.splat | f32 → `np.float16` (precision narrowing) |
| `dialects/tensor_ops.py` tensor.empty | substring match `"f16" in dtype_str` (fragile) |
| `interpreter.py` _ktir_dtype | missing i64 fallback → returned "f16" |
| `ir_types.py` TileRef.size_bytes | original bug: all non-f16 → 1 byte |

## Solution

Add `ktir_cpu/dtypes.py` as the single source of truth for all dtype mappings.
All six locations above now delegate here. No more silent fallbacks — unknown dtypes
raise `ValueError`; placeholder dtypes raise `NotImplementedError` (see below).

## Changes

- **new** `ktir_cpu/dtypes.py` — `SUPPORTED_DTYPES`, `to_np_dtype`, `bytes_per_elem`,
  `to_ktir_dtype`; private `_PLACEHOLDER_DTYPES` for fp8/mxfp8
- `ktir_cpu/ir_types.py` — `TileRef.size_bytes()` imports from `dtypes` (fixes layering)
- `ktir_cpu/memory.py` — `_get_np_dtype`/`_bytes_per_elem` re-exported from `dtypes`;
  `HBMSimulator.read_element` fixed (i64 was 4 bytes)
- `ktir_cpu/interpreter.py` — `_ktir_dtype` re-exported from `dtypes` (adds i64)
- `ktir_cpu/ops/memory_ops.py` — `tile_access` fixed (i64 pointer offset)
- `ktir_cpu/dialects/arith_ops.py` — `arith.constant` fixed (f32 → correct dtype)
- `ktir_cpu/dialects/tensor_ops.py` — `tensor.empty` and `tensor.splat` fixed
- **new** `tests/test_dtypes.py` — parametrized coverage for all supported dtypes
- `tests/test_ktir_cpu.py` — updated dtype assertions; added `test_tileref_size_bytes`

## Placeholder dtypes: fp8 / mxfp8

These do not appear in any `examples/` file, so they are registered as
`_PLACEHOLDER_DTYPES`. `to_np_dtype` raises `NotImplementedError` for them —
any future example that uses them will fail `test_examples` immediately, forcing a
proper implementation before merge.

Simulation semantics for fp8/mxfp8 need confirmation from @kiszk and
@Prasanth-Chatarasi before they can be promoted to full support.

## Tests

- Main baseline: 442 passed, 6 xfailed
- This branch: 467 passed, 6 xfailed (+25 new tests, zero regressions)

### DIFF
--- a/ktir_cpu/dialects/tensor_ops.py
+++ b/ktir_cpu/dialects/tensor_ops.py
@@ -20,6 +20,7 @@ from typing import Optional, Tuple
 import numpy as np
 
 from ..grid import CoreContext
+from ..dtypes import to_np_dtype
 from ..ir_types import Operation, Tile
 from .registry import register, register_parser
 
@@ -47,7 +48,7 @@ def tensor__empty(op, context, env):
     """Create an uninitialized tensor of the given shape."""
     shape = op.attributes.get("shape", (1,))
     dtype_str = op.attributes.get("dtype", "f16")
-    dtype = np.float16 if "f16" in dtype_str else np.float32
+    dtype = to_np_dtype(dtype_str)
     data = np.zeros(shape, dtype=dtype)
     return Tile(data, dtype_str, shape)
 
@@ -77,9 +78,9 @@ def tensor__splat(op, context, env):
         np_dtype = np.int32
         dtype_str = "i32"
     else:
-        np_dtype = np.float16 if dtype_str in ("f16", "f32") else np.int32
+        np_dtype = to_np_dtype(dtype_str)
 
-    scalar_val = float(scalar) if np_dtype == np.float16 else int(scalar)
+    scalar_val = float(scalar) if np.issubdtype(np_dtype, np.floating) else int(scalar)
     data = np.full(shape, scalar_val, dtype=np_dtype)
     return Tile(data, dtype_str, shape)
 
--- a/ktir_cpu/dialects/tensor_ops.py
+++ b/ktir_cpu/dialects/tensor_ops.py
@@ -80,8 +80,7 @@ def tensor__splat(op, context, env):
     else:
         np_dtype = to_np_dtype(dtype_str)
 
-    scalar_val = float(scalar) if np.issubdtype(np_dtype, np.floating) else int(scalar)
-    data = np.full(shape, scalar_val, dtype=np_dtype)
+    data = np.full(shape, scalar, dtype=np_dtype)
     return Tile(data, dtype_str, shape)
 
 
--- a/ktir_cpu/dtypes.py
+++ b/ktir_cpu/dtypes.py
@@ -1,0 +1,84 @@
+# Copyright 2025 The Torch-Spyre Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Canonical KTIR dtype mappings.
+
+Single source of truth for all dtype-related logic.  Every module that
+needs to convert between KTIR dtype strings and NumPy dtypes should
+import from here.
+"""
+
+import numpy as np
+
+SUPPORTED_DTYPES: dict[str, np.dtype] = {
+    # floating point
+    "f16":     np.dtype(np.float16),
+    "fp16":    np.dtype(np.float16),
+    "float16": np.dtype(np.float16),
+    "f32":     np.dtype(np.float32),
+    "float32": np.dtype(np.float32),
+    # integer
+    "i32":     np.dtype(np.int32),
+    "si32":    np.dtype(np.int32),
+    "index":   np.dtype(np.int32),
+    "i64":     np.dtype(np.int64),
+    "si64":    np.dtype(np.int64),
+}
+
+# Placeholder dtypes: not yet exercised by any example.  to_np_dtype raises
+# NotImplementedError for these so that any future example that uses them
+# fails test_examples immediately, forcing a proper implementation first.
+# Pending confirmation from @kiszk @Prasanth-Chatarasi on simulation semantics.
+_PLACEHOLDER_DTYPES: frozenset[str] = frozenset({"fp8", "mxfp8"})
+
+_REVERSE_MAP: dict[np.dtype, str] = {
+    np.dtype(np.float16): "f16",
+    np.dtype(np.float32): "f32",
+    np.dtype(np.int32):   "i32",
+    np.dtype(np.int64):   "i64",
+}
+
+
+def to_np_dtype(dtype: str) -> np.dtype:
+    """Convert a KTIR dtype string to a NumPy dtype.
+
+    Raises NotImplementedError for placeholder dtypes (fp8, mxfp8).
+    Raises ValueError for unrecognised strings.
+    """
+    if dtype in _PLACEHOLDER_DTYPES:
+        raise NotImplementedError(
+            f"dtype {dtype!r} is a placeholder pending hardware confirmation; "
+            "update SUPPORTED_DTYPES before adding examples that use it"
+        )
+    try:
+        return SUPPORTED_DTYPES[dtype]
+    except KeyError:
+        raise ValueError(f"Unsupported KTIR dtype: {dtype!r}")
+
+
+def bytes_per_elem(dtype: str) -> int:
+    """Return element size in bytes for a KTIR dtype string."""
+    return int(to_np_dtype(dtype).itemsize)
+
+
+def to_ktir_dtype(np_dtype: np.dtype) -> str:
+    """Map a NumPy dtype to a canonical KTIR dtype string.
+
+    Raises ValueError for unrecognised NumPy dtypes.
+    """
+    np_dtype = np.dtype(np_dtype)
+    try:
+        return _REVERSE_MAP[np_dtype]
+    except KeyError:
+        raise ValueError(f"No KTIR dtype for NumPy dtype: {np_dtype!r}")
--- a/ktir_cpu/dtypes.py
+++ b/ktir_cpu/dtypes.py
@@ -39,7 +39,9 @@ SUPPORTED_DTYPES: dict[str, np.dtype] = {
 # Placeholder dtypes: not yet exercised by any example.  to_np_dtype raises
 # NotImplementedError for these so that any future example that uses them
 # fails test_examples immediately, forcing a proper implementation first.
-# Pending confirmation from @kiszk @Prasanth-Chatarasi on simulation semantics.
+# fp8/mxfp8 are registered as placeholders that raise NotImplementedError. 
+# RFC 0682 doesn't define them; torch-spyre uses "fp8" at 1 byte/elem on hardware. 
+# NOTE: What should the simulation representation be?
 _PLACEHOLDER_DTYPES: frozenset[str] = frozenset({"fp8", "mxfp8"})
 
 _REVERSE_MAP: dict[np.dtype, str] = {
--- a/ktir_cpu/ir_types.py
+++ b/ktir_cpu/ir_types.py
@@ -72,9 +72,9 @@ class TileRef:
 
     def size_bytes(self) -> int:
         """Calculate size in bytes."""
+        from .memory import _bytes_per_elem
         total_elements = np.prod(self.shape)
-        bytes_per_element = 2 if self.dtype == "f16" else 1  # f16=2, mxfp8=1
-        return int(total_elements * bytes_per_element)
+        return int(total_elements * _bytes_per_elem(self.dtype))
 
 
 @dataclass
--- a/ktir_cpu/ir_types.py
+++ b/ktir_cpu/ir_types.py
@@ -37,7 +37,7 @@ class Tile:
     in MLIR).  Produced by load operations and consumed by compute ops.
     """
     data: np.ndarray
-    dtype: str  # "f16" or "mxfp8"
+    dtype: str  # see dtypes.SUPPORTED_DTYPES
     shape: Tuple[int, ...]
 
     def copy(self) -> 'Tile':
@@ -72,9 +72,8 @@ class TileRef:
 
     def size_bytes(self) -> int:
         """Calculate size in bytes."""
-        from .memory import _bytes_per_elem
-        total_elements = np.prod(self.shape)
-        return int(total_elements * _bytes_per_elem(self.dtype))
+        from .dtypes import bytes_per_elem
+        return int(np.prod(self.shape) * bytes_per_elem(self.dtype))
 
 
 @dataclass
--- a/ktir_cpu/memory.py
+++ b/ktir_cpu/memory.py
@@ -91,20 +91,7 @@ import numpy as np
 # Module-level helpers shared by HBMSimulator and LXScratchpad
 # ---------------------------------------------------------------------------
 
-def _get_np_dtype(dtype: str) -> np.dtype:
-    """Convert a KTIR dtype string to a NumPy dtype."""
-    if dtype in ("f16", "fp16", "float16"):
-        return np.float16
-    if dtype in ("i32", "si32", "index"):
-        return np.int32
-    if dtype in ("i64", "si64"):
-        return np.int64
-    return np.float32  # f8 / mxfp8 / others approximated as float32
-
-
-def _bytes_per_elem(dtype: str) -> int:
-    """Return element size in bytes for a KTIR dtype string."""
-    return int(np.dtype(_get_np_dtype(dtype)).itemsize)
+from .dtypes import to_np_dtype as _get_np_dtype, bytes_per_elem as _bytes_per_elem
 
 
 def _find_allocation(
@@ -294,7 +281,7 @@ class HBMSimulator:
             Use ``read(ptr, 1, dtype)[0]`` instead.  This method will be
             removed when the tt dialect is updated.
         """
-        bytes_per_elem = 2 if dtype in ("f16", "fp16", "float16") else 4
+        bytes_per_elem = _bytes_per_elem(dtype)
         alloc = _find_allocation(self.memory, addr, bytes_per_elem)
         if alloc is None:
             return np.float16(0.0)
--- a/ktir_cpu/memory.py
+++ b/ktir_cpu/memory.py
@@ -91,13 +91,13 @@ import numpy as np
 # Module-level helpers shared by HBMSimulator and LXScratchpad
 # ---------------------------------------------------------------------------
 
-from .dtypes import to_np_dtype as _get_np_dtype, bytes_per_elem as _bytes_per_elem
+from .dtypes import to_np_dtype, bytes_per_elem
 
 
 def _find_allocation(
     memory: Dict[int, np.ndarray],
     ptr: int,
-    bytes_per_elem: int,
+    elem_size: int,
 ) -> Optional[Tuple[int, np.ndarray, int]]:
     """Find the allocation containing byte address *ptr*.
 
@@ -113,17 +113,17 @@ def _find_allocation(
         return (ptr, memory[ptr], 0)
     for base_ptr, data in memory.items():
         # Use the allocation's own itemsize to compute its byte span,
-        # not the caller's bytes_per_elem (which reflects the access dtype
+        # not the caller's elem_size (which reflects the access dtype
         # and may differ from the stored dtype).
         end_ptr = base_ptr + data.size * data.itemsize
- 
+
         # NOTE: the ptr in memory check in the first return
         #       actually handles the ptr == base_ptr case.
         # - therefore the strict equality base_ptr < ptr
-        #   is meant to emphasize that equality checks will 
+        #   is meant to emphasize that equality checks will
         #   not come to this branch of logic.
         if base_ptr < ptr < end_ptr:
-            elem_offset = (ptr - base_ptr) // bytes_per_elem
+            elem_offset = (ptr - base_ptr) // elem_size
             return (base_ptr, data, elem_offset)
     return None
 
@@ -133,7 +133,7 @@ def _read_flat(
     ptr: int,
     n_elements: int,
     np_dtype: np.dtype,
-    bytes_per_elem: int,
+    elem_size: int,
 ) -> np.ndarray:
     """Read *n_elements* elements starting at byte address *ptr*.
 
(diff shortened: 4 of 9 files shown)


fabianlim (changes requested):
# PR #8 Review — Fix `TileRef.size_bytes()` element size calculation

**Reviewer:** Claude
**PR:** https://github.com/torch-spyre/ktir-cpu/pull/8

---

## Overview

The PR replaces hardcoded byte-width logic in `TileRef.size_bytes()` with a call to
`_bytes_per_elem()` from `memory.py`. The old code treated everything except f16 as
1 byte, which under-counted i32 (4×), i64 (8×), and f32 (4×).

---

## Correctness

The fix is arithmetically correct for i32/i64/f32. One semantic subtlety:

**mxfp8 handling changes semantics.** The old code returned 1 byte/elem
(hardware-accurate). The new code returns 4 bytes/elem because `_get_np_dtype("mxfp8")`
falls through to `np.float32` — consistent with how mxfp8 is stored in the simulator,
but divergent from the actual hardware element width. This is already the established
pattern in the codebase (`memory.py:102` and the two tests in `test_ktir_cpu.py:108,157`
all treat mxfp8 as float32), so the change aligns `TileRef` with that convention.
However, the call site in `TileRef.size_bytes()` should carry a comment explaining the
discrepancy so future readers aren't surprised.

---

## Architectural Concern — Layering Violation

`ir_types.py` is the foundational types module (currently no upstream project imports).
The fix implicitly adds:

```python
from .memory import _bytes_per_elem
```

This creates a **downward dependency** — `ir_types.py` (core data model) importing
from `memory.py` (memory simulation layer). The current import graph is:

```
ir_types.py  ←  memory.py  ←  latency.py
                            ←  grid.py
                            ←  interpreter.py
```

Adding the reverse edge makes `ir_types` depend on `memory`, and risks an import cycle
if `memory.py` ever imports `ir_types` (a natural future direction).

It also exposes a private `_`-prefixed function across module boundaries, which leaks
an implementation detail of `memory.py`.

**Recommended alternative:** inline the logic directly in `ir_types.py` (it's three
lines) or extract both `_get_np_dtype` and `_bytes_per_elem` to a small shared
`dtypes.py` utility module that both `ir_types.py` and `memory.py` can import from.

---

## Impact Scope — PR Description Is Misleading

The PR states: *"The fix corrects memory tracking in the latency model's `track_lx()`
function."*

This is not accurate. `TileRef.size_bytes()` is never called from `track_lx()`.
All three `track_lx()` call sites use `Tile.size_bytes()` (which calls
`self.data.nbytes` — already correct via NumPy):

- `interpreter.py:217` — `result` is a `Tile`
- `control_ops.py:118,161` — `val` is a `Tile`

The actual impact is limited to direct callers of `TileRef.size_bytes()`, of which
only one exists in the codebase today: `test_ktir_cpu.py:425` (f16 only — the case
that was already correct before this PR).

The PR description should be corrected to accurately reflect the real scope.

---

## Missing Tests

The existing test at `test_ktir_cpu.py:425` only covers `dtype="f16"`, which was
correct before this PR. No tests are added for the broken cases being fixed.
Suggested additions:

```python
# i32: was 4 bytes total (wrong), now 16
ref_i32 = TileRef(base_ptr=0, shape=(4,), strides=[1], memory_space="HBM", dtype="i32")
assert ref_i32.size_bytes() == 16

# i64: was 4 bytes total (wrong), now 32
ref_i64 = TileRef(base_ptr=0, shape=(4,), strides=[1], memory_space="HBM", dtype="i64")
assert ref_i64.size_bytes() == 32

# mxfp8: was 4 bytes total (hardware-accurate), now 16 (simulation representation)
ref_mxfp8 = TileRef(base_ptr=0, shape=(4,), strides=[1], memory_space="HBM", dtype="mxfp8")
assert ref_mxfp8.size_bytes() == 16
```

Without these, the fix has no regression guard.

---

## Broader Issue — Dtype Handling Is Fragmented

This PR is a symptom of a wider problem: there is no single authoritative
dtype→byte-width mapping in the codebase. The same logic is duplicated in at least
three places, with three slightly different implementations:

| Location | Logic | Dtypes covered |
|---|---|---|
| `memory.py:94–107` | `_get_np_dtype` + `_bytes_per_elem` | f16, i32, i64, mxfp8/others→float32 |
| `memory_ops.py:94` | `2 if dtype in ("f16","fp16","float16") else 4` | f16, everything else→4 |
| `memory_ops.py:297` | same inline expression | f16, everything else→4 |
| `ir_types.py:76` | `2 if dtype == "f16" else 1` | f16, everything else→1 (the bug) |

The `Tile.dtype` docstring (`ir_types.py:40`) still says `"f16" or "mxfp8"`, omitting
i32 and i64 entirely, which suggests the dtype surface has grown without a coordinated
update to any of the byte-width sites.

The recommended fix — a shared `dtypes.py` with a single `bytes_per_elem(dtype)` —
would eliminate all of the above duplication and make future dtype additions (e.g. bf16,
i8) a one-line change in one place. As long as each module rolls its own mapping, this
class of bug will recur.

---

## Summary

| Concern | Severity |
|---|---|
| Importing private `_bytes_per_elem` from a higher-level module into `ir_types.py` | Medium — layering violation |
| mxfp8 silently changes from 1 byte (hardware) to 4 bytes (simulation) with no comment | Low — undocumented semantic shift |
| No tests for i32, i64, mxfp8 cases | Medium — zero regression guard |
| PR description incorrectly attributes fix to `track_lx()` | Low — misleading, should be corrected |
| Dtype→byte-width logic duplicated in 4 places with inconsistent coverage | Medium — root cause of this bug class |

**Verdict:** The core bug fix is valid and consistent with the existing mxfp8/float32
convention in `memory.py`. Before merging: add tests for the fixed dtypes, add a
comment in `size_bytes()` about the mxfp8 simulation-vs-hardware distinction, correct
the PR description, and resolve the layering concern. A follow-up to consolidate dtype
handling into a shared `dtypes.py` is strongly recommended to prevent recurrence.

ktir_cpu/ir_types.py
```diff
@@ -72,9 +72,9 @@ def __post_init__(self):
 
     def size_bytes(self) -> int:
         """Calculate size in bytes."""
+        from .memory import _bytes_per_elem
```
fabianlim: in general im concerned about how the dtypes are being handled. According to the review, other than `_bytes_per_elem`, there seems to be other adhoc ways of handling dtypes. The best is to
- have a centralized way of handling them
- figure out a set of dtypes we want to handle, and reject all others


@kiszk @Prasanth-Chatarasi i think we basically only need to support f16, f32, i32 tensors?

WarningRan:
### Re: mxfp8 semantic change (1 byte → 4 bytes)

The review noted that this PR changes mxfp8 from 1 byte/elem to 4 bytes/elem — not just a bug fix, but a semantic change. I want to explain why this is intentional, and check if the approach is acceptable.

**Short version:** The old code (1 byte) was the only place in the codebase that treated mxfp8 as 1 byte. Everywhere else already uses 4 bytes. This PR makes `TileRef.size_bytes()` consistent with the rest.

**Evidence:**

NumPy has no native fp8 type, so the interpreter approximates mxfp8 as `np.float32` (4 bytes/elem):
- `_get_np_dtype("mxfp8")` → `np.float32` (`memory.py:102`, fallback path)
- `Tile.size_bytes()` → `data.nbytes` on a float32 array → 4 bytes/elem
- `_bytes_per_elem("mxfp8")` → 4
- Tests confirm: `test_ktir_cpu.py:108,157` both assert `_get_np_dtype("mxfp8") == np.float32`

The old `TileRef.size_bytes()` returning 1 byte was the only outlier.

**The trade-off I'm aware of:**

4 bytes = simulation size, not hardware size. If `TileRef.size_bytes()` were used for LX capacity tracking with mxfp8, it would overestimate by 4x vs. real hardware. But using 1 byte would make `TileRef` and `Tile` disagree on the same tensor — which seems worse.

**What I checked outside ktir-cpu:**

- RFC 0682: no mention of mxfp8 or any fp8 variant
- torch-spyre: supports fp8 (E4M3/E5M2) at 1 byte on hardware, but uses the name "fp8", not "mxfp8" — the relationship is unclear

This is also connected to the dtype set question — if we only need f16/f32/i32 (per @fabianlim's suggestion), the mxfp8 trade-off goes away entirely.

**Does this reasoning look correct? If so, I'll add a comment in `size_bytes()` noting the distinction.**

WarningRan:
### Dtype centralization: audit findings and proposal

@fabianlim Following up on your point about centralizing dtype handling. I audited the full codebase and want to share the findings and a concrete proposal.

#### 1. Current state: dtype logic is scattered across 8+ locations

Every place that needs "KTIR dtype string → numpy type" or "dtype → byte size" has its own inline implementation:

| # | Mechanism | Location | What it does | Dtypes covered | Issue |
|---|-----------|----------|-------------|----------------|-------|
| 1 | `_get_np_dtype()` | `memory.py:94-102` | KTIR string → numpy dtype | f16 (3 aliases), i32 (3 aliases), i64 (2 aliases), others → float32 | f32 has no explicit branch (correct by fallback accident); unknown dtypes silently become float32 |
| 2 | `_bytes_per_elem()` | `memory.py:105-107` | KTIR string → byte size | Delegates to `_get_np_dtype` above | Correct, but lives in `memory.py` (upper-layer module) |
| 3 | `_ktir_dtype()` | `interpreter.py:34-43` | numpy dtype → KTIR string | f16, f32, i32, others → "f16" | i64 missing — int64 arrays get tagged "f16" (8 bytes → 2) |
| 4a | inline bytes | `memory_ops.py:94` (tile_access) | bytes per element | f16 → 2, others → 4 | i64 gets 4 instead of 8 — wrong pointer offsets. This is a hot path |
| 4b | inline bytes | `memory.py:297` (read_element) | same | same | same (deprecated method) |
| 5a | inline np_dtype | `arith_ops.py:153` (arith.constant) | KTIR string → numpy dtype | f16 → float16, others → int32 | f32 gets mapped to int32 |
| 5b | inline np_dtype | `tensor_ops.py:77-80` (tensor.splat) | same | f16/f32 → float16, others → int32 | f32 gets mapped to float16 (narrowing) |
| 5c | inline np_dtype | `tensor_ops.py:50` (tensor.empty) | same | "f16" substring → float16, others → float32 | Substring match; no integer types |

Additionally, parsers (`ktdp_ops.py:223`, `parser_utils.py:25-42`) extract dtype strings from MLIR without any validation — typos flow through silently.

#### 2. Inconsistency matrix

These independent implementations disagree on several dtypes:

| dtype | `_get_np_dtype` | `_ktir_dtype` | tile_access bytes | arith.constant | tensor.splat |
|-------|-----------------|---------------|-------------------|----------------|--------------|
| f16 | ✓ float16 | ✓ "f16" | ✓ 2 | ✓ float16 | ✓ float16 |
| f32 | ✓ float32 (fallback) | ✓ "f32" | ✓ 4 | ✗ **int32** | ✗ **float16** |
| i32 | ✓ int32 | ✓ "i32" | ✓ 4 | ✓ int32 | ✓ int32 |
| i64 | ✓ int64 | ✗ **"f16"** | ✗ **4 (should be 8)** | ✗ int32 | ✗ int32 |
| mxfp8 | ✓ float32 (fallback) | ✗ "f16" | ✓ 4 | ✗ int32 | ✗ float16 |

**Note:** The ✗ entries may not be triggered today (depends on whether MLIR programs with those dtypes reach that code path), but they would produce silently incorrect results if those dtypes are used in the future.

#### 3. Proposal: `ktir_cpu/dtypes.py`

Create a new bottom-layer module as the single source of truth for dtype definitions.

`dtypes.py` only imports numpy — it becomes the lowest-level module in the project. Dependency graph:

```
dtypes.py (new) ← ir_types.py ← memory.py ← all other modules
```

No circular dependency risk. `ir_types.py` imports `dtypes.py` instead of `memory.py`, resolving the layering issue from the review.

Three public APIs:

```python
SUPPORTED_DTYPES: dict[str, np.dtype] = {
    "f16": np.float16,  "fp16": np.float16,  "float16": np.float16,  # 2 bytes
    "f32": np.float32,                                                 # 4 bytes — explicit, no longer fallback
    "i32": np.int32,    "si32": np.int32,     "index": np.int32,      # 4 bytes
    "i64": np.int64,    "si64": np.int64,                             # 8 bytes  ← needs confirmation
    # fp8 — numpy has no native fp8; approximated as float32
    # (hardware: 1 byte/elem, simulation: 4 bytes/elem)
    "fp8": np.float32,  "mxfp8": np.float32,                          #          ← needs confirmation
}

def to_np_dtype(dtype: str) -> np.dtype:
    """Raises ValueError on unknown dtype instead of silent fallback."""

def bytes_per_elem(dtype: str) -> int:
    """Delegates to to_np_dtype."""

def to_ktir_dtype(np_dtype: np.dtype) -> str:
    """Reverse mapping. Raises ValueError on unknown numpy dtype."""
```

**@kiszk @Prasanth-Chatarasi** — Two groups of dtypes in `SUPPORTED_DTYPES` need your confirmation:

- **i64/si64**: Present in `_get_np_dtype` but missing from `_ktir_dtype`. Does KTIR need to support i64 tensors? If yes, it stays and all code paths get fixed. If no, we remove it from the mapping.
- **mxfp8/fp8**: Not defined in RFC 0682. torch-spyre uses "fp8" (E4M3/E5M2, 1 byte on hardware), ktir-cpu uses "mxfp8" (approximated as np.float32, 4 bytes in simulation). The relationship between the two names is undocumented. Should we keep mxfp8 support? Is "mxfp8" the same as torch-spyre's "fp8"?

We will build `dtypes.py` with all currently-used dtypes and annotate the uncertain ones in the mapping table. Removing a dtype after confirmation is a one-line change — this does not block the work.

#### 4. Changes

After creating `dtypes.py`, replace all scattered inline logic:

| File | Current code | Replace with | What it fixes |
|------|-------------|--------------|---------------|
| `ir_types.py:75` | `from .memory import _bytes_per_elem` | `from .dtypes import bytes_per_elem` | Layering issue |
| `memory.py:94-107` | `_get_np_dtype` / `_bytes_per_elem` standalone impl | Delegate to `dtypes` | f32 explicit + unknown dtype raises |
| `interpreter.py:34-43` | `_ktir_dtype` missing i64, fallback "f16" | `dtypes.to_ktir_dtype` | i64 reverse mapping |
| `memory_ops.py:94` | `2 if f16 else 4` | `dtypes.bytes_per_elem(parent_ref.dtype)` | i64 byte size (4 → 8) |
| `arith_ops.py:153` | `float16 if "f16" else int32` | `dtypes.to_np_dtype(dtype_str)` | f32 no longer becomes int32 |
| `tensor_ops.py:77-80` | `float16 if in ("f16","f32") else int32` | `dtypes.to_np_dtype(dtype_str)` | f32 no longer narrowed to float16 |
| `tensor_ops.py:50` | `"f16" in dtype_str` substring match | `dtypes.to_np_dtype(dtype_str)` | Exact matching |
| `memory.py:297` | `2 if f16 else 4` | `dtypes.bytes_per_elem(dtype)` | i64 byte size |
| `ir_types.py:40` | docstring says `"f16" or "mxfp8"` | Update to list all supported dtypes | Documentation accuracy |

#### 5. Tests

- New `tests/test_dtypes.py`: parametrized tests covering forward mapping, reverse mapping, and `ValueError` on unknown dtypes
- New `test_tileref_size_bytes`: parametrized over all supported dtypes
- Existing `test_hbm_get_np_dtype` / `test_lx_get_np_dtype` remain compatible

fabianlim:
@WarningRan 

 Good design. dtypes.py as a single source of truth is the right direction.
 
  One thing to nail down: to_np_dtype and bytes_per_elem should only accept keys in SUPPORTED_DTYPES — not freeform strings. Both should raise ValueError consistently for anything not in the map. The current silent fallback to float32 for unknown dtypes is exactly what hides bugs like this one; don't carry that forward.
 
  For the initial implementation, keep SUPPORTED_DTYPES minimal — only what the tests actually exercise: f16/fp16/float16, f32/float32, i32/si32/index. Add fp8 and mxfp8 as explicit placeholders mapping to float32 in simulation, with a comment that they're pending confirmation from @kiszk @Prasanth-Chatarasi . Update the "unknown" assertions in test_hbm_get_np_dtype / test_lx_get_np_dtype to expect ValueError.

WarningRan:
@fabianlim Thanks for the clear guidance. One question before I start coding:

`indexed_add.mlir` uses `memref<2xi64>`, and `test_examples.py:311` seeds it with `np.int64`. Should i64/si64 be included in the initial `SUPPORTED_DTYPES`, or should indexed_add be left for a follow-up?

fabianlim:
@WarningRan 

>indexed_add.mlir uses memref<2xi64>, and test_examples.py:311 seeds it with np.int64. Should i64/si64 be included in the initial SUPPORTED_DTYPES, or should indexed_add be left for a follow-up?

Sorry i missed out on this point. Yes then lets include it. In general any dtype that appears in any of examples lets include, the rest lets put placeholders so they get tracked. And we want to put the tests such that, in the future if anybody puts those any of those placeholder data types, there will be some test in `test_examples` that fails that example

ktir_cpu/memory.py
```diff
@@ -91,20 +91,7 @@
...
-        return np.int32
-    if dtype in ("i64", "si64"):
-        return np.int64
-    return np.float32  # f8 / mxfp8 / others approximated as float32
-
-
-def _bytes_per_elem(dtype: str) -> int:
-    """Return element size in bytes for a KTIR dtype string."""
-    return int(np.dtype(_get_np_dtype(dtype)).itemsize)
+from .dtypes import to_np_dtype as _get_np_dtype, bytes_per_elem as _bytes_per_elem
```
fabianlim: > Suggestion 

> These are imported under the old private names, making callsites look like local helpers when they are the canonical public API. The fix is to import under original names and rename the bytes_per_elem parameter in _find_allocation / _read_flat to avoid the name collision (e.g. elem_size):

```python
from .dtypes import to_np_dtype, bytes_per_elem

def _find_allocation(memory, ptr, elem_size):
    ...
    elem_offset = (ptr - base_ptr) // elem_size

def _read_flat(memory, ptr, n_elements, np_dtype, elem_size):
    ...
    alloc = _find_allocation(memory, ptr, elem_size)


# callsites in HBMSimulator.read / LXScratchpad.read
np_dtype = to_np_dtype(dtype)
return _read_flat(self.memory, ptr, n_elements, np_dtype, bytes_per_elem(dtype))
```

fabianlim: Also there seems to be dead instance methods on `HBMSimulator` and `LXScratchpad`

Both classes still carry:

```python
def _get_np_dtype(self, dtype: str) -> np.dtype:
    return _get_np_dtype(dtype)
```

Nothing calls these through `self` — all internal callers use the module-level alias
directly. They should be removed.

WarningRan: Done in the latest commit: imported `to_np_dtype/bytes_per_elem` under their original names, renamed the bytes_per_elem parameter to `elem_size` in `_find_allocation/_read_flat`, and removed the dead `_get_np_dtype` instance methods along with their now-redundant tests.

ktir_cpu/dtypes.py
```diff
@@ -0,0 +1,84 @@
...
+    # floating point
+    "f16":     np.dtype(np.float16),
+    "fp16":    np.dtype(np.float16),
+    "float16": np.dtype(np.float16),
+    "f32":     np.dtype(np.float32),
+    "float32": np.dtype(np.float32),
+    # integer
+    "i32":     np.dtype(np.int32),
+    "si32":    np.dtype(np.int32),
+    "index":   np.dtype(np.int32),
```
WarningRan, line 34: MLIR index is pointer-sized (64-bit on a 64-bit target). Pinned to int32 here to match what the parser currently emits. Should this be int64 for Spyre simulation?

ktir_cpu/dtypes.py
```diff
@@ -0,0 +1,84 @@
...
+    "i64":     np.dtype(np.int64),
+    "si64":    np.dtype(np.int64),
+}
+
+# Placeholder dtypes: not yet exercised by any example.  to_np_dtype raises
+# NotImplementedError for these so that any future example that uses them
+# fails test_examples immediately, forcing a proper implementation first.
+# Pending confirmation from @kiszk @Prasanth-Chatarasi on simulation semantics.
+_PLACEHOLDER_DTYPES: frozenset[str] = frozenset({"fp8", "mxfp8"})
```
WarningRan, line 45: @kiszk @Prasanth-Chatarasi fp8/mxfp8 are registered as placeholders that raise NotImplementedError. RFC 0682 doesn't define them; torch-spyre uses "fp8" at 1 byte/elem on hardware. What should the simulation representation be?

ktir_cpu/dialects/tensor_ops.py
```diff
@@ -77,9 +78,9 @@ def tensor__splat(op, context, env):
         np_dtype = np.int32
         dtype_str = "i32"
     else:
-        np_dtype = np.float16 if dtype_str in ("f16", "f32") else np.int32
+        np_dtype = to_np_dtype(dtype_str)
 
-    scalar_val = float(scalar) if np_dtype == np.float16 else int(scalar)
+    scalar_val = float(scalar) if np.issubdtype(np_dtype, np.floating) else int(scalar)
```
WarningRan: Changed from `== np.float16` so that f32 scalars also cast via `float() `instead of `int()`. Is this the right policy for all float dtypes?

fabianlim: a little bit confused here. if its already float, why do we need to do a recast? I feel recasting is dangerous and takes away the design that we are now centralizing the dtype management in `to_np_dtype`

ktir_cpu/dtypes.py
fabianlim: @WarningRan can you clarify what "Pinned to int32 here to match what the parser currently emits." specifically means?

WarningRan: The comment was imprecise — "what the parser currently emits" doesn't accurately describe the reason. The mapping follows an existing convention throughout the codebase: the old memory.py grouped "i32", "si32", and "index" together, and linalg_ops produces Tile(..., "index", ...). dtypes.py inherits that convention. The open question stands: should index be int64 for a 64-bit Spyre target?

ktir_cpu/dialects/tensor_ops.py
WarningRan: You're right. `np.full` handles the conversion itself - the pre-cast was redundant and inconsistent with centralizing dtype logic in `to_np_dtype`. Removed in the latest commit.

ktir_cpu/dtypes.py
fabianlim: @mudhakar do you know the answer to this questions?

fabianlim: @WarningRan ok it turns out that for `index` is a more complicated problem. The answer is that index should be modeled by `int32`, but the reason is very subtle. I documented the reason in this follow up issue and I assigned you https://github.com/torch-spyre/ktir-cpu/issues/14

For now I can approve this PR.

ktir_cpu/dtypes.py
fabianlim: @WarningRan lets not put names in the codebase, all you need to do ist to just document down the discussion

fabianlim (dismissed):
I approved as we have achieved a consolidation of the dtype logic, which is I think the scope of this PR. There is a more subtle-issue with how the HBM memory should be indexed, that is broken off into https://github.com/torch-spyre/ktir-cpu/issues/14.

fabianlim (approved):

PR #9, opened 2026-04-23
MLIR Frontend PR 2 — Test mixins + KTIRParserBase ABC

### LINKED ISSUE #6: [Feature] MLIR-backed parser frontend

fabianlim:
> [!IMPORTANT]
  >  To be merged after #5 

## Roadmap

> MLIR frontend parser series: add `MLIRFrontendParser` as a swappable
> alternative to the regex-based `KTIRParser`.

**PR 1 — Decouple parser**: Make the tokenizer and region detector fully
structural so they require zero dialect-specific metadata. Pure
refactoring — no new API, no new features.

**PR 2 — Test mixins + KTIRParserBase ABC**: Introduce the
`KTIRParserBase` abstract base class and refactor both test suites to use
mixins that decouple test assertions from parser implementation. This is
the infrastructure that enables any future parser backend to reuse all
existing tests.

**PR 3 — MLIRFrontendParser**: Implement `MLIRFrontendParser` in
`ktir_cpu/mlir_frontend/` using `mlir_ktdp.ir` structural walk with
`MLIRTypeAdapter` (30+ per-op handlers). Add adapter test suites that
reuse all existing tests via the mixin infrastructure from PR 2.

---

- PR 1 — Decouple parser from dialect-specific logic
- **→ PR 2 — Test mixins + KTIRParserBase ABC**
- PR 3 — MLIRFrontendParser + adapter test suite

## Goal

Enable any future parser backend to reuse all existing tests without
modification. Introduce the `KTIRParserBase` contract and refactor both
test suites (`test_dialects_parse.py`, `test_examples.py`) into a mixin
architecture that separates test logic from parser implementation.

## Background

To support alternative parser backends (e.g. MLIR frontend parser, C++
dialect extensions), the interpreter needs a swappable parser contract,
and tests must not be coupled to the regex parser's behaviour. In
particular, tests must not hardcode SSA argument names from MLIR source
because different backends may normalise them differently (e.g. the MLIR
frontend parser normalises to positional `arg0`, `arg1`, ...).

### `ktir_cpu/parser.py` — `KTIRParserBase` ABC

```python
class KTIRParserBase(ABC):
    @abstractmethod
    def parse_module(self, mlir_text: str) -> IRModule: ...

    def parse_file(self, filepath: str) -> IRModule:
        with open(filepath) as f:
            return self.parse_module(f.read())
```

Any object satisfying this interface can be injected into the interpreter.
`KTIRParser` inherits from it — existing usage unchanged.

### `ktir_cpu/interpreter.py` — injectable parser

Add `parser: Optional[KTIRParserBase]` to `KTIRInterpreter.__init__()`.
When `None` (default), falls back to `KTIRParser()`. Existing call sites
(`KTIRInterpreter()`, `KTIRInterpreter(latency_config=cfg)`) unchanged.

### `tests/test_dialects_parse.py` — `ParseTestMixin`

Portable assertion API that decouples test logic from parser implementation:

```
ParseTestMixin
├── _parse(op_text, parse_ctx, args)   # override per backend
├── assert_op_type(op, expected)       # parser-agnostic
├── assert_num_operands(op, n)         # parser-agnostic
├── assert_result_type(op, expected)   # parser-agnostic
├── assert_attribute(op, key, value)   # parser-agnostic
└── assert_operand_names(op, *names)   # parser-specific — override in adapters
```

All 5 parse test classes (`TestArithParsers`, `TestLinalgParsers`,
`TestTensorParsers`, `TestKtdpParsers`, `TestScfParsers`) refactored to
inherit `ParseTestMixin` and use the assertion API.

All op text examples fixed to be valid MLIR (parseable by both regex and
MLIR frontend backends).

### `tests/test_examples.py` — `InterpreterTestMixin`

Factory mixin for execution tests:

```python
class InterpreterTestMixin:
    def _make_interp(self):
        return KTIRInterpreter()
```

All 8 execution test classes refactored to use `self._make_interp()` and
`interp.arg_names(func_name)` instead of hardcoded MLIR argument names.

### Design methodology: extending tests for a new parser backend

Documented in `test_dialects_parse.py` module docstring:

1. Create a mixin that overrides `_parse()` to construct ops via your
   backend
2. Override `assert_operand_names()` if your backend normalises SSA names
3. Override `assert_attribute()` for keys whose representation differs
4. For each `TestXxxParsers` class, create:
   `class TestXxxMyBackend(MyMixin, TestXxxParsers): pass`
5. Run the full suite — inherited tests are the correctness oracle

### Other file changes

| File | Change |
|------|--------|
| `ktir_cpu/ir_types.py` | Add `arg_names` helper to `IRFunction` |
| `ktir_cpu/dialects/arith_ops.py` | Add unbraced `dense<...>` branch for `arith.constant` |
| `ktir_cpu/dialects/ktdp_ops.py` | Fix `intermediate_variables\s*\(` whitespace in regex |
| `ktir_cpu/dialects/ktdp_helpers.py` | Minor refactor for regex parser path |

## Test plan

- [x] All existing tests pass (`uv run pytest tests/ -v`)
- [x] `KTIRInterpreter(parser=None)` falls back to `KTIRParser()` (backwards compatible)
- [x] `ParseTestMixin` assertion API covers: op_type, num_operands, result_type, attribute, operand_names
- [x] All 5 parse test classes inherit `ParseTestMixin`
- [x] All 8 execution test classes inherit `InterpreterTestMixin` and use `arg_names()`

### DIFF
--- a/ktir_cpu/ir_types.py
+++ b/ktir_cpu/ir_types.py
@@ -142,8 +142,10 @@ class IRFunction:
     grid: Tuple[int, int, int]  # Grid shape from function attributes
     return_type: Optional[str] = None
     tensor_sizes: Dict[str, Dict[str, Any]] = field(default_factory=dict, init=False, repr=False)
+    arg_names: List[str] = field(default_factory=list, init=False, repr=False)
 
     def __post_init__(self):
+        self.arg_names = [name.lstrip("%") for name, _ in self.arguments]
         for op in _iter_ops(self.operations):
             if op.op_type == "ktdp.construct_memory_view" and op.operands:
                 arg_name = op.operands[0].lstrip("%")
--- a/ktir_cpu/ir_types.py
+++ b/ktir_cpu/ir_types.py
@@ -142,10 +142,12 @@ class IRFunction:
     grid: Tuple[int, int, int]  # Grid shape from function attributes
     return_type: Optional[str] = None
     tensor_sizes: Dict[str, Dict[str, Any]] = field(default_factory=dict, init=False, repr=False)
-    arg_names: List[str] = field(default_factory=list, init=False, repr=False)
+
+    @property
+    def arg_names(self) -> List[str]:
+        return [name.lstrip("%") for name, _ in self.arguments]
 
     def __post_init__(self):
-        self.arg_names = [name.lstrip("%") for name, _ in self.arguments]
         for op in _iter_ops(self.operations):
             if op.op_type == "ktdp.construct_memory_view" and op.operands:
                 arg_name = op.operands[0].lstrip("%")
--- a/tests/test_dialects_parse.py
+++ b/tests/test_dialects_parse.py
@@ -73,10 +73,14 @@ class ParseTestMixin:
     def _parse(self, op_text, parse_ctx=None, args=None):
         """Parse a single op and return the resulting Operation.
 
-        ``args`` is an optional schema that declares the SSA arguments the op
-        depends on, as a ``{name: mlir_type}`` mapping.  This is a convenience
-        for single-op testing: it avoids writing a full MLIR module while still
-        providing explicit types for operands with complex types.  Example::
+        ``args`` is an optional ``{name: mlir_type}`` mapping that declares
+        **all** external SSA operands the op depends on — values defined
+        outside the op text that the op references.  Names defined *within*
+        the op (results on the LHS of ``=``, region block args like ``%i``
+        in ``scf.for %i = ...``, and iter-args) should **not** be included.
+
+        This is a convenience for single-op testing: it avoids writing a
+        full MLIR module while still providing explicit types.  Example::
 
             self._parse(
                 "%r = linalg.reduce { arith.maxnumf }"
@@ -86,10 +90,13 @@ class ParseTestMixin:
                 args={"%x": "tensor<1x1024xf16>", "%init": "tensor<1xf16>"},
             )
 
-        The regex parser ignores ``args`` and operates on the op text directly.
-        ``BindingsParseTestMixin`` uses it to build a typed function wrapper.
+        The regex parser ignores ``args`` and operates on the op text
+        directly.  The MLIR frontend adapter uses it to build a typed
+        ``func.func`` wrapper, so **missing operands will cause MLIR
+        verification errors** when the adapter tests run.
 
-        If ``args`` is provided, every name must appear in ``op_text``.
+        If ``args`` is provided, every declared name must appear in
+        ``op_text``.
         """
         args = self._resolve_args(op_text, args)
         ctx = parse_ctx or _parse_ctx()
--- a/tests/test_dialects_parse.py
+++ b/tests/test_dialects_parse.py
@@ -23,6 +23,15 @@ Covers:
 - ktdp dialect parsers: ktdp.get_compute_tile_id, ktdp.construct_memory_view,
                          ktdp.construct_access_tile
 - scf dialect parsers: scf.for, scf.yield
+
+Design rule
+-----------
+The goal of the regex parser is to parse valid MLIR.  All op text examples
+in this file must therefore be valid MLIR.  Non-MLIR syntax forces the
+corresponding test_bindings_adapt.py test to be skipped, which defeats the
+purpose of sharing tests between the two parser backends.
+
+All op text examples have been verified to parse with both backends.
 """
 
 import numpy as np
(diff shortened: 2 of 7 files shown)


tests/test_examples.py
```diff
@@ -70,81 +107,86 @@ def test_construct_memory_view_attributes(self, path, func_name, entry):
...
     """End-to-end execution of softmax MLIR."""
 
     @pytest.mark.parametrize("path,func_name,entry", get_test_params("softmax_kernel", filter="triton-ktir"))
     def test_softmax_correct(self, path, func_name, entry):
         """Run softmax on 32 cores, verify against NumPy ground truth."""
-        interp = KTIRInterpreter()
+        interp = self._make_interp()
         interp.load(path)
 
+        output_ptr, input_ptr, n_rows = interp.arg_names(func_name)
```
kiszk: The new code does not seem to use `n_rows`. 

```suggestion
        output_ptr, input_ptr, _ = interp.arg_names(func_name)
```

tests/test_examples.py
```diff
@@ -70,81 +107,86 @@ def test_construct_memory_view_attributes(self, path, func_name, entry):
...
         ).astype(np.float16)
-        out = np.zeros((n_rows, n_padded_cols), dtype=np.float16)
+        out = np.zeros((n_rows_val, n_padded_cols), dtype=np.float16)
 
         kwargs = {k: v for k, v in entry["execute_kwargs"].items() if v is not None}
         kwargs["n_cols"] = n_real_cols  # fill dynamic kwarg
 
         outputs = interp.execute_function(
-            func_name, output_ptr=out, input_ptr=inp, **kwargs,
+            func_name, **{**kwargs, output_ptr: out, input_ptr: inp, n_rows: n_rows_val},
```
lchu6: sanity check:  based on line 170, it seems n_rows already in kwargs?  so duplicated n_rows here?

fabianlim: @lchu6 thanks for the catch, this is taken care off in https://github.com/torch-spyre/ktir-cpu/pull/9/commits/b629a621af4ffea41f1cf82f8617f6c4a30b301f

lchu6 (dismissed):

tests/test_examples.py
fabianlim: I further improved how kwargs are handled in https://github.com/torch-spyre/ktir-cpu/pull/9/commits/d43fbb5cf2c5c18650f0776b407ea7fc024e19c2, does this also address your concern @kiszk

tests/test_dialects_parse.py
```diff
@@ -39,12 +48,86 @@ def _parse_ctx():
...
+    def _resolve_args(self, op_text, args):
+        """Normalise and validate the args schema against op_text.
+
+        Returns ``args`` as a ``dict``, defaulting to ``{}`` if not provided.
+        Asserts that every declared name appears in ``op_text``.
+        """
+        args = args or {}
+        for name in args:
+            assert name in op_text, f"arg {name!r} not found in op_text"
+        return args
```
kiszk, line 116: @fabianlim  `args` is an optional argument. If `args` exist, do we expect `args` to have all of the RHS values used in `op_text`? Or, `args` may have some of the RHS values used in `op_text`.
If the former, it would be good to add the routine to check that condition like:

```
    import re
    ssa_names = set(re.findall(r'%\w+', op_text))
    result_names = set(re.findall(r'(%\w+)\s*=', op_text))
    operand_names = ssa_names - result_names
    
    missing = operand_names - args.keys()
    if missing:
        import warnings
        warnings.warn(
            f"SSA operands in op_text not declared in args: {missing}. "
            f"This may cause issues with MLIR bindings parser."
        )
```

ktir_cpu/ir_types.py
```diff
@@ -142,8 +142,10 @@ class IRFunction:
     grid: Tuple[int, int, int]  # Grid shape from function attributes
     return_type: Optional[str] = None
     tensor_sizes: Dict[str, Dict[str, Any]] = field(default_factory=dict, init=False, repr=False)
+    arg_names: List[str] = field(default_factory=list, init=False, repr=False)
```
kiszk: Is it better to use property?

```suggestion
    @property
    def arg_names(self) -> List[str]:
        """Return argument names without % prefix."""
        return [name.lstrip("%") for name, _ in self.arguments]
```

tests/test_examples.py
kiszk, line 420: The following verification part is a bit long. How about refactoring the following section into a method to improve readability?

tests/test_dialects_parse.py
fabianlim: Good question. The contract is:

  - args must declare all external SSA operands — values defined outside the op that the op references. For the bindings
  parser path this is mandatory since _parse wraps the op in func.func @_test({sig}), so any undeclared operand triggers
  an MLIR verification error. The regex parser ignores args entirely.

  The suggested completeness check (warning when op_text references SSA names not in args) is a good idea in principle,
  but tricky in practice — a naive %\w+ scan would false-positive on SSA names that are defined within the op text, not
  external inputs:
  - Region block args: %i in scf.for %i = ..., %a/%b in (%a : f32, %b : f32) { ... }
  - Results defined inside region bodies: %acc in scf.yield %acc

  Distinguishing external operands from internally-defined names requires parsing awareness that _resolve_args doesn't
  (and shouldn't) have.

  In practice the bindings parser adapter already enforces completeness strictly — if a test author forgets to declare an
  operand in args, the MLIR verifier rejects it immediately when the adapter tests run. So the gap is caught at test time
  rather than with a warning, which seems sufficient.

  I'll add a note to the args docstring to make the "must be complete" contract explicit.

ktir_cpu/ir_types.py
fabianlim: yes this is a good suggestion. Since arg_names is never set in the dataclass, it does not actually make sense to be an input.

tests/test_examples.py
fabianlim: The reference computation (lines 420–463) is specific to the paged attention tiled online-softmax and won't be reused
  elsewhere. Extracting it into a method would just move the same code one level of indirection away without reducing
  complexity — the reader still needs to understand the same math, now split across two locations. The inline comments
  already document the algorithm step by step (gather Q/K/V, apply causal mask, tiled online-softmax loop). I'd prefer to
  keep it inline for now; if we add more paged attention test variants that share this reference logic, we can extract it
  then.

kiszk (approved):
thanks, lgtm

PR #11, opened 2026-04-27
latency: account for index tensor HBM traffic in indirect access loads

WarningRan:
## Summary

The latency tracker underestimates `ktdp.load` cost when operating over an `IndirectAccessTile` — it only counted the result tensor bytes, ignoring the HBM reads for the index tensors that drive the gather. Hardware team confirmed each `ind(...)` dimension in `construct_indirect_access_tile` is a real HBM load at runtime, not free metadata.

**Concrete undercount for `indirect-access-copy.mlir` (64×64 tiles, 2×i32 index tensors):**

| HBM traffic | Bytes | Previously tracked |
|---|---|---|
| Read X (result, f16) | 64×64×2 = 8,192 | Yes |
| Read IDX1 (i32) | 64×64×4 = 16,384 | No |
| Read IDX2 (i32) | 64×64×4 = 16,384 | No |
| **Total** | **40,960** | **8,192 (5× undercount)** |

## Changes

**`ktir_cpu/latency.py`**
- `_memory_space()`: add explicit `IndirectAccessTile` branch — previously fell through to `"HBM"` fallback by accident (correct, but fragile)
- `_data_size()`: restructured to accumulate a running total; for `IndirectAccessTile` operands, sums bytes of all `index_views` in HBM; uses `elif isinstance(v, Tile)` + assert to handle future scatter stores correctly (when scatter store `NotImplementedError` is removed, both index bytes and data tile bytes will be counted without silent undercounting)

**`tests/test_latency.py`**
- Add `TestIndirectAccessLatency`: verifies `total_bytes` (49,152) and `memory_cycles` for `indirect-access-copy.mlir`; uses `tensor_input_output_sizes` for shape/dtype info

## Open question

Does gather access on Spyre require different bandwidth parameters than dense transfer? If yes, a follow-up PR will add per-element gather penalty modeling and a `gather_bandwidth` hardware parameter. If no, this PR is the complete solution for index-tensor latency modeling.

cc @fabianlim

### DIFF
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -30,7 +30,7 @@ import math
 import numpy as np
 
 from .ir_types import AccessTile, IndirectAccessTile, Tile, TileRef
-from .memory import _bytes_per_elem
+from .dtypes import bytes_per_elem
 
 
 from .dialects.registry import get_latency_category
@@ -293,7 +293,7 @@ class LatencyTracker:
             if isinstance(v, IndirectAccessTile):
                 for iv in v.index_views:
                     if iv.memory_space == "HBM":
-                        total += int(np.prod(iv.shape)) * _bytes_per_elem(iv.dtype)
+                        total += int(np.prod(iv.shape)) * bytes_per_elem(iv.dtype)
             elif isinstance(v, Tile):
                 if result is not None:
                     raise ValueError(
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -77,6 +77,8 @@ class HardwareConfig:
             costs ``2·M·N·K / systolic_flops_per_cycle`` cycles.
         transcendental_penalty: Multiplier for transcendental ops vs elementwise (estimated).
     """
+    # TODO: add gather_bandwidth_tb_s if Spyre scatter/gather BW differs from
+    # sequential HBM BW. Until confirmed by hardware team, both share hbm_bandwidth_tb_s.
     num_cores: int = 32
     clock_ghz: float = 1.0
     hbm_bandwidth_tb_s: float = 1.0
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -274,9 +274,9 @@ class LatencyTracker:
             if isinstance(v, AccessTile):
                 return v.parent_ref.memory_space
             if isinstance(v, IndirectAccessTile):
-                if any(iv.memory_space == "HBM" for iv in v.index_views):
-                    return "HBM"
-                return v.parent_ref.memory_space
+                all_lx = (v.parent_ref.memory_space == "LX" and
+                          all(iv.memory_space == "LX" for iv in v.index_views))
+                return "LX" if all_lx else "HBM"
         return "HBM"
 
     @staticmethod
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -272,6 +272,8 @@ class LatencyTracker:
             if isinstance(v, AccessTile):
                 return v.parent_ref.memory_space
             if isinstance(v, IndirectAccessTile):
+                if any(iv.memory_space == "HBM" for iv in v.index_views):
+                    return "HBM"
                 return v.parent_ref.memory_space
         return "HBM"
 
@@ -291,10 +293,11 @@ class LatencyTracker:
                     if iv.memory_space == "HBM":
                         total += int(np.prod(iv.shape)) * _bytes_per_elem(iv.dtype)
             elif isinstance(v, Tile):
-                assert result is None, (
-                    f"_data_size: Tile in operands but result is also a Tile ({type(result)}); "
-                    "no ktdp op should produce both"
-                )
+                if result is not None:
+                    raise ValueError(
+                        f"_data_size: Tile in operands but result is also a Tile ({type(result)}); "
+                        "no ktdp op should produce both"
+                    )
                 total += v.data.nbytes
         return total
 
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -29,7 +29,8 @@ from typing import Any, Dict, List, Optional, Tuple
 import math
 import numpy as np
 
-from .ir_types import AccessTile, Tile, TileRef
+from .ir_types import AccessTile, IndirectAccessTile, Tile, TileRef
+from .memory import _bytes_per_elem
 
 
 from .dialects.registry import get_latency_category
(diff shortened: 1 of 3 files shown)


fabianlim (changes requested):
# PR #11 Review: `latency: account for index tensor HBM traffic in indirect access loads`

**Author:** WarningRan | **Files changed:** `ktir_cpu/latency.py`, `tests/test_latency.py`

---

## Overview

The fix is correct and addresses a real 5× undercount. For a gather load `Y[m,k] = X[IDX1[m,k], IDX2[m,k]]` with 64×64 f16 tiles and two i32 index tensors, the model previously tracked only 8,192 bytes (result tile); the fix adds 16,384 + 16,384 bytes for the two index tensors, totaling 40,960 bytes. This is a legitimate bug: hardware confirmed each `ind(...)` dimension is a live HBM read.

---

## Issues

**1. `_memory_space` / `_data_size` inconsistency for mixed memory spaces**

This is the most important correctness issue. `_estimate()` (line 199-200) gates all memory cost on `_memory_space()`:
```python
if self._memory_space(operands) == "LX":
    return ("memory", 0.0, 0.0, 0)   # early exit — never reaches _data_size
```
`_memory_space()` now correctly returns `parent_ref.memory_space` for `IndirectAccessTile`. But if `parent_ref` is LX while the index views are in HBM, the entire op is treated as free — the HBM index traffic is silently skipped without ever calling `_data_size()`. This is a correctness hole in the mixed-memory-space case. Consider adding index-view HBM accounting directly in `_memory_space()` or restructuring `_estimate()` so the early return only applies when all participating refs are LX.

**2. `assert` inside production code at `latency.py:294`**

```python
assert result is None, (
    f"_data_size: Tile in operands but result is also a Tile ({type(result)}); "
    "no ktdp op should produce both"
)
```
`assert` is disabled with `-O`/`PYTHONOPTIMIZE`, so this silently double-counts bytes in an optimized run rather than raising. The contract violation is real and worth enforcing, but use `raise ValueError(...)` instead.

**3. Hardcoded address map in the test (`test_latency.py:626`)**

```python
_addr_map = {"X_addr": 0, "IDX1_addr": 8192, "IDX2_addr": 16384, "Y_addr": 24576}
```
These are replicated from the `arith.constant` values in `indirect-access-copy.mlir`. If those constants change, the test fails with a silent wrong-data error rather than a clear assertion. The addresses can be parsed from the module using the same `interp` that's already loaded — see the pattern in `test_indirect_access.py`.

**4. `pytest.approx` inconsistency in the new test**

```python
assert abs(counters.memory_cycles - expected_memory_cycles) < 1.0
```
Every other tolerance check in this file uses `pytest.approx(..., rel=1e-3)`. The `< 1.0` absolute tolerance is fine numerically here (40.96 cycles), but it would silently pass with much larger error if tile sizes were reduced. Use `pytest.approx(expected_memory_cycles, rel=1e-3)` to be consistent and scale-agnostic.

**5. Missing test coverage**

- **LX index views:** No test verifying that index views in LX memory space contribute 0 bytes (tests only the all-HBM case).
- **Scatter store:** The PR docstring mentions scatter stores, and `_data_size()` handles `IndirectAccessTile` in operands for that path — but there is no test exercising it.

**6. Open question on gather bandwidth (flagged in PR, needs tracking)**

The PR correctly flags that gather accesses may have lower effective bandwidth than dense transfers due to memory controller inefficiency. This is unaddressed and optimistic in the current model. The PR should either open a follow-up issue or add a `# TODO` comment at the bandwidth application site in `_estimate()` so it doesn't get lost.

---

## Minor

- `int(np.prod(iv.shape))` on a 0-d `TileRef` (`shape = ()`) returns `1`, not `0`. Degenerate but worth noting.
- The new `_data_size` docstring is accurate. No concerns there.
- KTIR spec compliance: no new ops invented, no semantics changed — this is purely a latency model fix. ✓

---

## Summary

The core fix is sound. Before merging, I'd request:

1. Address the `_memory_space` / early-return inconsistency for mixed LX/HBM scenarios (correctness)
2. Change `assert` → `raise ValueError` (robustness)
3. Derive seed addresses from the parsed module instead of hardcoding them (test fragility)
4. Replace `< 1.0` with `pytest.approx` (consistency)
5. Add a follow-up issue or `TODO` comment for gather-specific bandwidth modeling

ktir_cpu/latency.py
```diff
@@ -270,20 +273,35 @@ def _memory_space(operands: List[Any]) -> str:
                 return v.memory_space
             if isinstance(v, AccessTile):
                 return v.parent_ref.memory_space
+            if isinstance(v, IndirectAccessTile):
+                if any(iv.memory_space == "HBM" for iv in v.index_views):
```
WarningRan: Previously the IndirectAccessTile branch always returned `parent_ref.memory_space`, so an LX parent caused `_estimate()` to early-return 0 cycles and skip `_data_size()` entirely - silently dropping all HBM index tensor traffic.

ktir_cpu/latency.py
```diff
@@ -270,20 +273,35 @@ def _memory_space(operands: List[Any]) -> str:
...
-            if isinstance(v, Tile):
-                return v.data.nbytes
-        return 0
+            if isinstance(v, IndirectAccessTile):
+                for iv in v.index_views:
+                    if iv.memory_space == "HBM":
+                        total += int(np.prod(iv.shape)) * _bytes_per_elem(iv.dtype)
+            elif isinstance(v, Tile):
+                if result is not None:
+                    raise ValueError(
```
WarningRan, line 299: assert is silently disabled under `python -O` / `PYTHONOPTIMIZE`. `raise ValueError` ensures the invariant is always enforced.

WarningRan:
@fabianlim Thanks for the detailed review! Addressed all six points in three commits:

**Correctness fixes** (`latency.py`, see inline comments):
- **1 Mixed memory space**: `_memory_space()` now checks `index_views` before falling back to `parent_ref.memory_space`.
- **2 assert → ValueError**: `_data_size()` now uses `raise ValueError` so the invariant holds under `python -O`.

**Test quality** (`tests/test_latency.py`):
- **3 Hardcoded addresses**: `_addr_map` is now derived from `arith.constant` ops in the parsed module, keyed by the same SSA names used by `tensor_input_output_sizes()`.
- **4 Tolerance**: Replaced `abs(...) < 1.0` with `pytest.approx(rel=1e-3)`.
- **5 LX index view coverage**: Added `test_lx_index_views_excluded_from_hbm_bytes` to `TestLatencyEdgeCases`, using `parse_affine_set` to construct a valid `IndirectAccessTile` (same pattern as `test_latency_modeling.py`).
- **6 Gather bandwidth**: Added a `# TODO` in `HardwareConfig` noting that `gather_bandwidth_tb_s` is needed if Spyre scatter/gather BW differs from sequential — pending hardware team confirmation.

ktir_cpu/latency.py
```diff
@@ -77,6 +77,8 @@ class HardwareConfig:
             costs ``2·M·N·K / systolic_flops_per_cycle`` cycles.
         transcendental_penalty: Multiplier for transcendental ops vs elementwise (estimated).
     """
+    # TODO: add gather_bandwidth_tb_s if Spyre scatter/gather BW differs from
```
fabianlim, line 80: we should document this in `docs/latency.md` also. In the doc, we should question if it makes sense to have the same bandwidth for float and i32 tensors.

ktir_cpu/latency.py
fabianlim: This logic is a little fragile, and will be error prone if we decide to add more memory classes. I feel this is better

 ```python
 if isinstance(v, IndirectAccessTile):
     all_lx = (v.parent_ref.memory_space == "LX" and
               all(iv.memory_space == "LX" for iv in v.index_views))
     return "LX" if all_lx else "HBM"
 ```

fabianlim (approved):
All five issues from the previous review are addressed in the current diff:

1. Mixed LX/HBM early-return: `_memory_space()` now returns `"HBM"` if any participating ref is HBM, so `_estimate()` no longer silently drops index tensor traffic.
2. `assert` → `raise ValueError`: enforced unconditionally.
3. Hardcoded address map: replaced with constants derived from the parsed module.
4. `pytest.approx` consistency: `rel=1e-3` used throughout.
5. LX index view test: `test_lx_index_views_excluded_from_hbm_bytes` covers the mixed-memory case.

The `gather_bandwidth_tb_s` TODO and `docs/latency.md` note are good enough to track the open bandwidth question.

Approved.

PR #10, opened 2026-04-24
 MLIR Frontend PR 3 — MLIRFrontendParser + adapter test suite

fabianlim:
> [!IMPORTANT]
  >  To be merged after #9. Also maybe consider incorporating the ktir-mlir-frontend package first before merging this so we can activate the extra tests.

Closes #6 

## Roadmap

> MLIR frontend parser series: add `MLIRFrontendParser` as a swappable
> alternative to the regex-based `KTIRParser`.

**PR 1 — Decouple parser**: Make the tokenizer and region detector fully
structural so they require zero dialect-specific metadata. Pure
refactoring — no new API, no new features.

**PR 2 — Test mixins + KTIRParserBase ABC**: Introduce the
`KTIRParserBase` abstract base class and refactor both test suites to use
mixins that decouple test assertions from parser implementation. This is
the infrastructure that enables any future parser backend to reuse all
existing tests.

**PR 3 — MLIRFrontendParser**: Implement `MLIRFrontendParser` in
`ktir_cpu/mlir_frontend/` using `mlir_ktdp.ir` structural walk with
`MLIRTypeAdapter` (30+ per-op handlers). Add adapter test suites that
reuse all existing tests via the mixin infrastructure from PR 2.

---

- PR 1 — Decouple parser from dialect-specific logic
- PR 2 — Test mixins + KTIRParserBase ABC
- **→ PR 3 — MLIRFrontendParser + adapter test suite**

## Goal

Provide a second parser backend that uses the MLIR Python bindings
(`mlir_ktdp.ir`) instead of regex-based text parsing. This validates the
`KTIRParserBase` contract, catches divergences between the two parsers
via shared tests, and lays groundwork for consuming the upstream
https://github.com/torch-spyre/ktir-mlir-frontend as a Python package.

## Background

The regex-based `KTIRParser` works well but is inherently fragile: every
new MLIR syntax variation requires a regex update. The MLIR Python
bindings provide a structural parse/verify/walk pipeline that is
authoritative — if MLIR accepts it, the frontend parser can handle it.

The architecture uses the **adapter pattern**: `MLIRTypeAdapter` converts
`mlir.ir` typed attribute objects (e.g. `IntegerAttr`, `AffineMapAttr`)
to the same ktir_cpu abstraction types (`int`, `AffineMap`, `AffineSet`)
that the regex parser produces. Per-op handlers rename, synthesise, or
drop attributes so the resulting `Operation` objects are execution-ready.

[...]

PR #12, opened 2026-04-27
Wire CI to build and test mlir_frontend against ktir-mlir-frontend

fabianlim:
> [!IMPORTANT]
>  To be merged either before or after #10. So that we can trigger the mlir frontend tests 

## Summary

`tests/mlir_frontend/` was always skipped in CI because `mlir_ktdp` was never
installed. This PR pins `ktir-mlir-frontend` as an optional dependency, adds a
CI step that builds it from source using `setup_mlir.py`, and documents the
manual install path for contributors.

## This PR — add `mlir-frontend` optional dep and CI build step

`tests/mlir_frontend/conftest.py` skips the entire suite at module level when
`mlir_ktdp` is not importable. Because `ktir-mlir-frontend` has no PyPI release
and was never installed in CI, these tests have never run in CI.

- **`pyproject.toml`**: add `mlir-frontend` optional dependency pinned to
  `fabianlim/ktir-mlir-frontend@55a8a7ea` (tracks LLVM `e9846648`,
  llvmorg-22.1.3). Will move to `torch-spyre/ktir-mlir-frontend` once
  [torch-spyre/ktir-mlir-frontend#14](https://github.com/torch-spyre/ktir-mlir-frontend/pull/14)
  merges.

- **`.github/workflows/ci.yml`**: replace the single `uv sync --extra dev` step
  with three steps:
  - *Install build tools* — `apt-get install ninja-build` (required by
    scikit-build-core).
  - *Set up MLIR* — `curl` down `scripts/setup_mlir.py` from the pinned commit
    and run it with `--wheel --hash $LLVM_HASH`. `--wheel` activates the
    `mlir_wheel` fallback (no GitHub token needed in public CI); `--hash` makes
    the targeted LLVM build explicit so removing `--wheel` is the only change
    needed once artifacts are public. `CMAKE_ARGS` is written to `$GITHUB_ENV`
    so `uv sync` inherits it automatically in the next step.
  - *Install project and dependencies* — `uv sync --extra dev --extra
    mlir-frontend --no-build-isolation`.

- **`README.md`**: add *MLIR frontend bindings* subsection under Setup with the
  full manual install recipe and a note that the preamble goes away once
  [torch-spyre/ktir-mlir-frontend#12](https://github.com/torch-spyre/ktir-mlir-frontend/issues/12)
  is resolved.

- **`CONTRIBUTING.md`**: add a short pointer to the README section; kept
  general so it stays accurate after the wheel transition.

## Build strategy: why we build `ktir-mlir-frontend` from source

The `mlir-frontend` dep is pinned to a commit on `ktir-mlir-frontend@main`, not
a release tag. This is intentional: `ktir-mlir-frontend` is under active
development and has no PyPI releases yet, so CI must build from source in order
to test against unreleased upstream changes.

A future option is to download a pre-packaged wheel of `ktir-mlir-frontend`
once it publishes releases. However, switching to released wheels would mean CI
can only test against stable snapshots — it would no longer catch breaking
changes introduced on `ktir-mlir-frontend@main` before they reach a release.
Nightly wheel infrastructure for `ktir-mlir-frontend` would be needed to close
that gap, which is overhead that lives in that repo, not here.

**Decision for now:** keep the build-from-source approach. The cost is more CI
bandwidth and build time (downloading LLVM + compiling the extension). The
benefit is that updating `KTIR_MLIR_FRONTEND_COMMIT` in `ci.yml` is sufficient
to test any commit of `ktir-mlir-frontend`, including unreleased work. Revisit
once `ktir-mlir-frontend` adopts a release cadence with wheels.

> **Note:** `--wheel` / `--hash` in `setup_mlir.py` refer to how the
> **LLVM/MLIR** dependency is resolved (stable `mlir_wheel` vs. a pinned
> nightly artifact), not to packaging of `ktir-mlir-frontend` itself.

## What is NOT changing

- All existing tests and their pass/fail status are unaffected.
- The `uv sync` (no extras) path for end users is unchanged.
- `tests/mlir_frontend/` still skips gracefully if `mlir_ktdp` is absent
  (local dev without the optional dep).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,5 +55,8 @@ jobs:
       - name: Install project and dependencies
         run: uv sync --extra dev --extra mlir-frontend --no-build-isolation
 
-      - name: Run tests
-        run: uv run pytest -v
+      - name: Run basic tests (with regex parser)
+        run: uv run pytest -v --ignore=tests/mlir_frontend
+
+      - name: Run MLIR frontend parser tests
+        run: uv run pytest tests/mlir_frontend/ -v
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,34 @@ jobs:
       - name: Set up Python
         run: uv python install ${{ matrix.python-version }}
 
-      - name: Install dependencies
-        run: uv sync --extra dev
+      - name: Install build tools
+        run: sudo apt-get install -y ninja-build
+
+      # Temporary: download setup_mlir.py from the pinned ktir-mlir-frontend commit
+      # and run it to obtain MLIR_DIR.  We pass --wheel to explicitly activate the
+      # mlir_wheel fallback — the artifact-based path (keyed by LLVM_HASH) requires a
+      # GitHub token not available in public CI.  --hash is still passed so the intent
+      # is explicit; remove --wheel once the artifact is publicly accessible and this
+      # will automatically use the pinned LLVM build instead.
+      # CMAKE_ARGS is written to $GITHUB_ENV so uv sync inherits it automatically
+      # in the next step without needing an explicit prefix on the command line.
+      # Drop-in removal: once ktir-mlir-frontend publishes a wheel to PyPI, delete
+      # this step entirely — CMAKE_ARGS will be unset and `uv sync --extra mlir-frontend`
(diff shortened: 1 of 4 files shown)


lchu6 (approved):
LG. temporary solution to unblock dev, we should have a following PR once wheels are ready.

PR #13, opened 2026-04-27
[Bug] Remove `{ dim = N }` single-result form of `ktdp.get_compute_tile_id`

fabianlim:
> [!IMPORTANT]
  >  This PR is stacked on top of #12, and should be merged after, so that we can test this change after the mlir-frontend tests are triggered

cc: @lchu6 @tnakaike @kiszk 

## Summary

Three MLIR examples used the single-result-with-explicit-dim form
`%a = ktdp.get_compute_tile_id { dim = N } : index` to extract one
dimension from a multi-dim grid.  This form is not declared in the
`GetComputeTileIdOp` ODS (`let arguments = (ins);`), so its preservation
depends on undocumented attribute-forwarding behaviour of the dialect
implementation.  This PR replaces all such uses with the well-defined
multi-result form and removes the `dim` attribute handling from the
entire stack.

## This PR — replace single-result-with-dim form with multi-result form

The `{ dim = N }` attribute on `ktdp.get_compute_tile_id` is not an
inherent attribute in the ODS definition; its preservation in parsed IR
relies on unspecified dialect behaviour.

- **`examples/triton-ktir/indexed_add.mlir` (line 9–10)**:
  two separate single-result ops `{ dim = 0 }` / `{ dim = 1 }` →
  one multi-result op `%grid0, %grid1 = ktdp.get_compute_tile_id : index, index`.

- **`examples/triton-ktir/paged_attention.mlir` (line 78–79)**:
  same pattern → `%pid0, %pid1 = ktdp.get_compute_tile_id : index, index`.

- **`examples/triton-ktir/sdpa_2d.mlir` (line 38)**:
  `{ dim = 0 }` on a 1-D grid (`grid = [1]`) — attribute is a no-op →
  removed, leaving `%pid_m = ktdp.get_compute_tile_id : index`.

- **`ktir_cpu/dialects/ktdp_ops.py` — executor (`ktdp__get_compute_tile_id`, line 34–36)**:
  removed `dim = op.attributes.get("dim", 0)`; single-result path now
  unconditionally returns `GridOps.gridid(context, 0)`.

- **`ktir_cpu/dialects/ktdp_ops.py` — regex parser (`parse_get_compute_tile_id`, lines 126–152)**:
  removed the optional `(?:\s*\{([^}]*)\})?` group and the `dim_match`
  extraction; regex now only matches the canonical `results = op : types` form.

[...]

PR #15, opened 2026-04-28
Fix examples for static shapes and block sizes

tnakaike:
This PR fixes the inconsistency between shapes and indexes (offsets).

This PR will cover [a ktir-mlir-frontend issue](https://github.com/torch-spyre/ktir-mlir-frontend/issues/15).

### DIFF
--- a/examples/triton-ktir/layernorm_fwd_ktir.mlir
+++ b/examples/triton-ktir/layernorm_fwd_ktir.mlir
@@ -7,7 +7,7 @@ module {
       %Mean: index, // mean 1152xf16
       %Rstd: index, // rstd 1152xf16
       %eps: f16
-  ) attributes {grid = [32, 1]} {
+  ) attributes {grid = [32]} {
     %core_id = ktdp.get_compute_tile_id : index
 
     %c0 = arith.constant 0 : index
--- a/examples/triton-ktir/layernorm_fwd_ktir.mlir
+++ b/examples/triton-ktir/layernorm_fwd_ktir.mlir
@@ -1,57 +1,62 @@
 module {
   func.func @_layer_norm_fwd_fused(
-      %X: index, // input 1151x8192xf16
-      %Y: index, // output 1151x8192xf16
-      %W: index, // weight 1151x8192xf16
-      %B: index, // bias 1151x8192xf16
-      %Mean: index, // mean 1151xf16
-      %Rstd: index, // rstd 1151xf16
-      %N: index, // number of columns in X: 8192
-      %eps: f16,
-      %BLOCK_SIZE: index // 1024
+      %X: index, // input 1152x8192xf16
+      %Y: index, // output 1152x8192xf16
+      %W: index, // weight 1152x8192xf16
+      %B: index, // bias 1152x8192xf16
+      %Mean: index, // mean 1152xf16
+      %Rstd: index, // rstd 1152xf16
+      %eps: f16
   ) attributes {grid = [32, 1]} {
     %core_id = ktdp.get_compute_tile_id : index
 
-    %c0_i32 = arith.constant 0 : index
-    %c1151_i32 = arith.constant 1151 : index
-    %c32_i32 = arith.constant 32 : index
+    %c0 = arith.constant 0 : index
+    %c1 = arith.constant 1 : index
+    %c32 = arith.constant 32 : index
+    %c36 = arith.constant 36 : index
+    %c1152 = arith.constant 1152 : index
+    %c1024 = arith.constant 1024 : index
+    %c8192 = arith.constant 8192 : index
     %f1_f16 = arith.constant 1.0 : f16
 
-    %X_view = ktdp.construct_memory_view %X, sizes: [1151, 8192], strides: [8192, 1] {
-        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1150 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
-    } : memref<1151x8192xf16>
+    %start_row = arith.muli %core_id, %c36 : index
+    %end_row = arith.addi %start_row, %c36 : index
 
-    %Y_view = ktdp.construct_memory_view %Y, sizes: [1151, 8192], strides: [8192, 1] {
-        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1150 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
-    } : memref<1151x8192xf16>
+    %X_view = ktdp.construct_memory_view %X, sizes: [1152, 8192], strides: [8192, 1] {
+        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1151 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
+    } : memref<1152x8192xf16>
 
-    %W_view = ktdp.construct_memory_view %W, sizes: [1151, 8192], strides: [8192, 1] {
-        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1150 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
-    } : memref<1151x8192xf16>
+    %Y_view = ktdp.construct_memory_view %Y, sizes: [1152, 8192], strides: [8192, 1] {
+        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1151 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
+    } : memref<1152x8192xf16>
 
-    %B_view = ktdp.construct_memory_view %B, sizes: [1151, 8192], strides: [8192, 1] {
-        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1150 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
-    } : memref<1151x8192xf16>
+    %W_view = ktdp.construct_memory_view %W, sizes: [1152, 8192], strides: [8192, 1] {
+        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1151 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
+    } : memref<1152x8192xf16>
 
-    %Mean_view = ktdp.construct_memory_view %Mean, sizes: [1151], strides: [1] {
-        coordinate_set = affine_set<(d0) : (d0 >= 0, -d0 + 1150 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
-    } : memref<1151xf16>
+    %B_view = ktdp.construct_memory_view %B, sizes: [1152, 8192], strides: [8192, 1] {
+        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1151 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
+    } : memref<1152x8192xf16>
 
-    %Rstd_view = ktdp.construct_memory_view %Rstd, sizes: [1151], strides: [1] {
-        coordinate_set = affine_set<(d0) : (d0 >= 0, -d0 + 1150 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
-    } : memref<1151xf16>
+    %Mean_view = ktdp.construct_memory_view %Mean, sizes: [1152], strides: [1] {
+        coordinate_set = affine_set<(d0) : (d0 >= 0, -d0 + 1151 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
+    } : memref<1152xf16>
 
-    scf.for %row = %core_id to %c1151_i32 step %c32_i32  : index {
+    %Rstd_view = ktdp.construct_memory_view %Rstd, sizes: [1152], strides: [1] {
+        coordinate_set = affine_set<(d0) : (d0 >= 0, -d0 + 1151 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
+    } : memref<1152xf16>
+
+    scf.for %row = %start_row to %end_row step %c1 : index {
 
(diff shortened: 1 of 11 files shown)


tnakaike:
@fabianlim I also updated tests to use the updated examples.

examples/triton-ktir/softmax_fwd_ktir.mlir
```diff
@@ -1,13 +1,15 @@
...
+      %input_ptr: index
   ) attributes {grid = [32, 1]} {
     %core_id = ktdp.get_compute_tile_id : index
 
-    %c32_i32 = arith.constant 32 : index
-    %c0_i32 = arith.constant 0 : index
+    %c0 = arith.constant 0 : index
+    %c1 = arith.constant 1 : index
+    %c128 = arith.constant 128 : index
+    %start_row = arith.muli %core_id, %c128 : index
```
kiszk, line 11: According to [RFC](https://github.com/torch-spyre/RFCs/blob/main/0682-KtirSpec/0682-KtirSpecRFC.md#1-ktdpget_compute_tile_id-ktdpgetcomputetileid), `ktdp.get_compute_tile_id` returns `index, index` when `grid` is 2D. Can this arith operation accept `index, index`?

cc @fabianlim

fabianlim: @kiszk yes it was already accepting `index, index`, and another `{ dim = [] }` form, but in this PR #13 I was trying to remove the latter, because that is non-standard. so yes for 2D (and even 3D I believe) we accept the tuple form

tnakaike:
@kiszk Current implementation allows us to get multiple index-type IDs from ktdp.get_compute_tile_id. Upper bound is the number of grid dimension. Also, we can get only one ID even when the number of grid dimension is 2. In this case, a user can assume that the ID corresponds to the first dimension.

fabianlim:
@tnakaike regarding your comment:

> Also, we can get only one ID even when the number of grid dimension is 2. In this case, a user can assume that the ID corresponds to the first dimension.

I don't think this is supported by RFC 0682. The spec defines the result count as determined by `type(results)`, and only gives two forms:

- 1D grid: `%tile_id = ktdp.get_compute_tile_id : index`
- 2D grid: `%tile_id:2 = ktdp.get_compute_tile_id : index, index`

There's no provision for requesting fewer results than grid dimensions, and no rule assigning "first dimension" semantics to a single-result call on a 2D grid. Accepting `ktdp.get_compute_tile_id : index` inside a `{grid = [32, 1]}` function silently discards the second dimension ID — the spec does not sanction this, and which dimension is returned would be implementation-defined rather than specified.

The correct pattern when only one dimension is needed would be to receive both and simply not use one:

```mlir
%pid_0, %pid_1 = ktdp.get_compute_tile_id : index, index
// use %pid_0, ignore %pid_1
```

If single-dimension extraction is intended as a valid shorthand, it should be explicitly specified in the RFC before the implementation relies on it. I've opened [ktir-mlir-frontend#16](https://github.com/torch-spyre/ktir-mlir-frontend/issues/16) for related underspecified dynamic-shape semantics — this could be tracked there or in a separate issue.

cc @kiszk

examples/triton-ktir/softmax_fwd_ktir.mlir
fabianlim: To clarify the confusion here: kiszk's question and tnakaike's answer are talking past each other slightly.

kiszk is asking whether `arith.muli` can accept a tuple `index, index` — the answer is no, `arith.muli` only takes scalar/tensor operands. You'd need to destructure the tuple first:
```mlir
%pid_0, %pid_1 = ktdp.get_compute_tile_id : index, index
%offset = arith.muli %pid_0, %c128 : index
```

tnakaike's answer addresses the parser supporting the tuple form, which is correct, but doesn't address the spec conformance issue.

The deeper problem: this file has `attributes {grid = [32, 1]}` (a 2D grid) but uses `%core_id = ktdp.get_compute_tile_id : index` (single result). RFC 0682 is unambiguous — a 2D grid must yield two results. The `[32, 1]` grid is effectively 1D in the second dimension, but the spec has no provision for dropping dimensions from the result.

The correct form here would be:
```mlir
%core_id, %_ = ktdp.get_compute_tile_id : index, index
```
(or keep `%pid_1` unused if MLIR has no discard syntax)

I've raised this more broadly in the top-level PR comment. It may warrant an RFC clarification if single-result extraction on a 2D grid is intentionally supported.

cc @fabianlim @tnakaike

examples/triton-ktir/layernorm_fwd_ktir.mlir
```diff
@@ -1,57 +1,62 @@
...
-        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1150 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
-    } : memref<1151x8192xf16>
+    %start_row = arith.muli %core_id, %c36 : index
+    %end_row = arith.addi %start_row, %c36 : index
 
-    %Y_view = ktdp.construct_memory_view %Y, sizes: [1151, 8192], strides: [8192, 1] {
-        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1150 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
-    } : memref<1151x8192xf16>
+    %X_view = ktdp.construct_memory_view %X, sizes: [1152, 8192], strides: [8192, 1] {
+        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 1151 >= 0, d1 >= 0, -d1 + 8191 >= 0)>, memory_space = #ktdp.spyre_memory_space<HBM>
```
fabianlim: The 1151→1152 row count change concerns me. The original strided loop:
```mlir
scf.for %row = %core_id to %c1151_i32 step %c32_i32
```
naturally handles non-divisible sizes — core `k` processes rows `k, k+32, k+64, ...` and the last few cores simply do one fewer iteration when `n_rows % n_cores != 0`.

The PR replaces this with a contiguous block partition:
```mlir
%start_row = arith.muli %core_id, %c36 : index
%end_row   = arith.addi %start_row, %c36 : index
scf.for %row = %start_row to %end_row step %c1
```
This silently requires `n_rows % n_cores == 0`. Bumping 1151→1152 (= 32×36) papers over that requirement rather than fixing it. If we ever need to handle non-divisible sizes in the future — which is the common case for real Triton kernels — this change hides that the requirement exists at all.

I'd suggest either keeping the strided loop (preserving the original semantics), or if the contiguous partition is intentional, explicitly documenting the divisibility precondition and keeping 1151 as a separate test case that exercises the remainder path.

cc @tnakaike

fabianlim (changes requested):
Several issues raised in inline comments that should be addressed before merging. Summary:

**1. `get_compute_tile_id` spec non-conformance (all changed kernels with `grid = [X, 1]`)**
RFC 0682 requires result count to match grid dimensionality. `grid = [32, 1]` is a 2D grid and must use `%pid_0, %pid_1 = ktdp.get_compute_tile_id : index, index`. Using a single result is not specified by the RFC and should not be relied upon. See inline comments on `softmax_fwd_ktir.mlir`.

**2. Strided loop → contiguous partition changes the semantics for non-divisible sizes (layernorm, softmax_small)**
The original `scf.for %row = %core_id to %n_rows step %c32` naturally handles `n_rows % n_cores != 0` — the last few cores simply do one fewer iteration. The replacement contiguous partition (`start = core_id * tile, end = start + tile`) silently requires even divisibility. Bumping 1151 → 1152 (= 32×36) papers over this rather than fixing it. If non-divisible sizes need to be handled in the future, this change hides that requirement entirely. See inline comment on `layernorm_fwd_ktir.mlir`.

**3. `conftest.py` is stale and inconsistent with the changed MLIRs**
- The `_layer_norm_fwd_fused` entry still has comment `# stride and N match construct_memory_view sizes: [1151, 8192]` (should be 1152) and still includes `stride`, `N`, `BLOCK_SIZE` in `execute_kwargs` — these args were removed from the MLIR function signature.
- The `softmax_kernel_small` entry still lists `n_rows`, `BLOCK_SIZE`, `input_row_stride`, `output_row_stride` in `execute_kwargs`, but the MLIR no longer has those function arguments. `_run_softmax` in `test_latency.py` passes them via `**kwargs` directly to `execute_function` — this will either silently be ignored or cause an error.

**4. `matmul_small.mlir` not updated consistently**
`matmul_fwd_ktir.mlir` had `K`, `BLOCK_SIZE_M/N/K` removed from the function signature and baked as constants. `matmul_small.mlir` was not changed — it still takes these as arguments. If the intent is to eliminate scalar function args across all kernels, `matmul_small.mlir` should be updated consistently. If not, the rationale for the asymmetry should be documented.

**Merge ordering**
This PR should be merged after #12 (Wire CI to build and test mlir_frontend against ktir-mlir-frontend) so that the updated examples can also be validated by the mlir-frontend parser, not just the interpreter.

cc @tnakaike

lchu6:
@fabianlim @tnakaike 

what was the reason that many of the args are gone in the ktir examples?   if I recall correctly, original ktir examples had same set of args as original kernel.  e.g. vector_add has n_elements, matmul has M/N/etc.

many of those seem gone in the current ktir examples in this repo?

fabianlim:
@lchu6 

> what was the reason that many of the args are gone in the ktir examples? if I recall correctly, original ktir examples had same set of args as original kernel. e.g. vector_add has n_elements, matmul has M/N/etc.

Can I clarify what are you referring to? In this PR some of the examples are being adjusted because of inconsistencies as documented in https://github.com/torch-spyre/ktir-mlir-frontend/issues/15. Since these are standalone examples, regardless if they were transpiled from triton or not, we decided not to leave arguments in the `func` if they already have been backed into the ktir, as in the issue above describes inconsitencies can arise.

lchu6:
@fabianlim basically I am repeating this

>  M and N
currently appear nowhere in the function signature.

this is not addressed yet in this pr

lchu6:
ok maybe I misunderstood, this pr does not fix https://github.com/torch-spyre/ktir-mlir-frontend/issues/15, right?  only address a small portion of it.

lchu6:
and

> M and N
currently appear nowhere in the function signature.

still needs to be fixed, right?  maybe in a different PR

fabianlim:
@lchu6 — sorry for the confusion in my earlier response, I misread your intent. You're right that this PR does not fully resolve torch-spyre/ktir-mlir-frontend#15 as originally written. The issue was filed expecting that dimensions like M and N would become dynamic function arguments threaded through `sizes:`, `strides:`, and `coordinate_set`. @tnakaike instead took the complementary approach: remove the dangling args and make the examples self-consistent with static shapes. That leaves the dynamic-arg direction (M/N as function args propagated into `construct_memory_view`) unresolved.

We opened torch-spyre/ktir-mlir-frontend#16 to track that, partly because it's not yet clear how dynamic sizes should propagate through `construct_memory_view` → `construct_access_tile` → `load`/`store` — the RFC leaves several of those semantics underspecified. So the fix for M/N in matmul (and the equivalent in the other kernels) is deferred until that design question is settled.

lchu6:
it should be just as-is?

if you look at the signature of memory_view:
```
...
Variadic<Index>:$sizes,
DenseI64ArrayAttr:$static_sizes,
...
```

it supports static size and dynamic size (and a mix if necessary)

so 
```
# cast n_elements to index first, then should be just:
shape=[%n_elements]
```
should be fine.

fabianlim:
@lchu6 — agreed, it may be simpler than I made it sound. Since we're not tackling it in this PR, I'll add a note to torch-spyre/ktir-mlir-frontend#16 pointing to your observation so whoever picks it up has a clear starting point.

kiszk:
@fabianlim Thank you for summarizing a list of issues.
> 1. get_compute_tile_id spec non-conformance (all changed kernels with grid = [X, 1])

We can simply replace `grid = [32, 1]` with `grid = [32]` since 2D core layout is not used in these examples.
Then, we can discuss at https://github.com/torch-spyre/ktir-mlir-frontend/issues/17

fabianlim:
@tnakaike @kiszk Agreed — pushed a commit (fd93e78) that replaces `grid = [N, 1]` with `grid = [N]` in all affected examples and updates the parser to handle 1D grid attributes. All tests pass.

fabianlim:
@tnakaike Updating on the status of the [requested changes](https://github.com/torch-spyre/ktir-cpu/pull/15#pullrequestreview-4189471400):

**1. `get_compute_tile_id` spec non-conformance** — ✅ Resolved in fd93e78. All examples now use 1D `grid = [N]` which is consistent with the single-result `get_compute_tile_id : index` form. Parser updated to handle 1D grids. The broader question about sub-dimensional grid access is tracked in torch-spyre/ktir-mlir-frontend#17.

**2. Strided loop → contiguous partition semantics** — Still open. The contiguous partition (`start = core_id * tile, end = start + tile`) requires even divisibility. The bump from 1151 → 1152 hides this. Should be addressed or documented.

**3. `conftest.py` stale entries** — Still open. `execute_kwargs` for `_layer_norm_fwd_fused` and `softmax_kernel_small` reference args that no longer exist in the MLIR signatures.

**4. `matmul_small.mlir` inconsistency** — Still open. `matmul_fwd_ktir.mlir` has scalar args baked as constants but `matmul_small.mlir` still takes them as function arguments.

**Merge ordering** — #12 is now merged, so this is no longer a blocker.

fabianlim (approved):
Approving — the three issues from my earlier review are now addressed:

**1. `get_compute_tile_id` spec non-conformance** — All examples now use 1D `grid = [N]`, consistent with the single-result `get_compute_tile_id : index` form. Parser updated to handle 1D grids. Broader question tracked in torch-spyre/ktir-mlir-frontend#17.

**2. Strided loop → contiguous partition** — Layernorm reverted to 1151 rows (non-divisible by 32 cores) with a proper `min(end, n_rows)` clamp, demonstrating that contiguous partition does not require even divisibility.

**3. Stale `conftest.py` entries** — Removed dead `execute_kwargs` (`stride_*`, `n_elements`, `BLOCK_SIZE`, `n_rows`, `n_cols`, etc.) that no longer match MLIR function signatures. Fixed `_run_softmax` and `_run_vector_add` to stop passing phantom kwargs.

Point 4 (`matmul_small.mlir` keeping scalar function args) is intentional — `test_work_splitting_matmul` needs to override `BLOCK_SIZE_M` at runtime.

cc @tnakaike @kiszk

fabianlim:
@tnakaike We can't merge until your two commits (`dfa6135` and `68e98d6`) are cryptographically signed (GPG or SSH `-S` signature). Could you rebase and sign them? For example:

```
git rebase --onto main main nakaike/examples --exec 'git commit --amend --no-edit -S'
git push --force-with-lease
```

Alternatively if you have `commit.gpgsign = true` in your git config, a simple interactive rebase to re-commit them should pick up the signature automatically.

lasch:
It's been dormant for ~2 months. Any updates on this PR?
Any help needed? Any reevaluation required because of other code changes that happened in the meantime?

PR #18, opened 2026-04-29
[BUG] MLIRFrontendParser compatibility fixes

fabianlim:
## Summary

- Add positional argument remapping in `execute_function` so callers can use
  source-level argument names regardless of which parser backend is active.
- Register missing no-attribute op handlers in `MLIRTypeAdapter`: `arith.fptosi`,
  `math.log`.
- Update the pinned `ktir-mlir-frontend` commit to `bad31fb` (removes
  `mlir-wheel` build dependency, enabling `uv sync --extra mlir-frontend`
  without a GitHub token or custom find-links).

## Motivation

Running real-world KTIR kernels through `MLIRFrontendParser` exposed two
classes of failure:

1. **Argument name mismatch** — the MLIR bindings walk normalises function
   argument names to positional `%arg0`, `%arg1`, … regardless of the source
   names (`%X`, `%Y`, `%d`, …).  Callers who construct kwargs with source names
   hit a `KeyError` at execution time because the interpreter maps kwargs to SSA
   values via `"%" + arg_name`.

2. **Missing adapter handlers** — `arith.fptosi` and `math.log` were not
   registered in `MLIRTypeAdapter._adapt_handlers`, causing
   `NotImplementedError` during the structural walk even though the regex parser
   already handles both ops. (A third op, `arith.maxf`, was also missing but
   this is a deprecated alias for `arith.maximumf` which is already registered —
   the fix is to rename `maxf` → `maximumf` in the kernel MLIR files rather than
   adding a handler for the deprecated name.)

### `ktir_cpu/interpreter.py`

`execute_function` now performs a one-time positional remap of kwargs before
allocating HBM: if *none* of the caller's keys appear in the function's
declared arg list but the counts match, kwargs are rebound to the declared
names in declaration order.  This is transparent to callers using the regex
parser (source names are preserved) and fixes the mlir-frontend silently.

```python
declared = func.arg_names
if kwargs and declared and not any(k in declared for k in kwargs):
    if len(kwargs) == len(declared):
        kwargs = dict(zip(declared, kwargs.values()))
```

### `ktir_cpu/mlir_frontend/parser.py`

Added to the `_no_attrs` install list:
- `arith.fptosi` — float-to-signed-integer cast
- `math.log` — natural log

### `pyproject.toml` / `uv.lock`

Updated pinned `ktir-mlir-frontend` commit from `55a8a7e` to `bad31fb`.  The
new commit removes `mlir_wheel` from `build-system.requires`, so `uv sync
--extra mlir-frontend` resolves from PyPI without needing a custom find-links
or GitHub token.


lchu6:
> Missing adapter handlers — arith.maxf, arith.fptosi, and math.log
were not registered

> Register missing no-attribute op handlers in MLIRTypeAdapter: arith.fptosi,
math.log.

you mentioned 3 were missing but only added 2?

lchu6 (approved):

fabianlim:
Good catch — `arith.maxf` is a deprecated LLVM op (removed in newer builds). Rather than adding a handler for it, the kernel MLIR files were updated to use `arith.maximumf`, which is already registered in the adapter. So only `arith.fptosi` and `math.log` needed new handlers.

I'll update the PR description to clarify this.

PR #20, opened 2026-04-30
Fix hex float constants and arith.truncf no-op

### LINKED ISSUE #19: [Bug] Hex float constants (0xFC00 : f16, 0xFF800000 : f32) misinterpreted as integers
## Bug

The regex parser's `parse_numeric` treats hex literals like `0xFC00` and
`0xFF800000` as raw unsigned integers instead of IEEE 754 bit patterns.
This produces wrong values when the constant is declared as a float type.

## Reproducer

```python
from ktir_cpu import KTIRInterpreter
import numpy as np

mlir_f16 = """
module {
  func.func @test_hex_f16(%out: index) attributes {grid = [1, 1]} {
    %neg_inf = arith.constant 0xFC00 : f16
    %tile = tensor.splat %neg_inf : tensor<1xf16>
    %view = ktdp.construct_memory_view %out, sizes: [1], strides: [1] {
      coordinate_set = affine_set<(d0) : (d0 >= 0, -d0 + 0 >= 0)>,
      memory_space = #ktdp.spyre_memory_space<HBM>
    } : memref<1xf16>
    %acc = ktdp.construct_access_tile %view[%c0] {
      access_tile_set = affine_set<(d0) : (d0 >= 0, -d0 + 0 >= 0)>,
      access_tile_order = affine_map<(d0) -> (d0)>
    } : memref<1xf16> -> !ktdp.access_tile<1xindex>
    ktdp.store %tile, %acc : tensor<1xf16>, !ktdp.access_tile<1xindex>
    return
  }
}
"""

interp = KTIRInterpreter()
interp.load(mlir_f16)
out = np.zeros(1, dtype=np.float16)
result = interp.execute_function("test_hex_f16", out=out)
print(result["out"])  # actual: [6.45e+04]  expected: [-inf]
```

The f32 case (`arith.constant 0xFF800000 : f32`) raises an overflow:
```
Python integer 4286578688 out of bounds for int32
```

## Expected behavior

`0xFC00 : f16` → f16 -inf (bit pattern reinterpretation)
`0xFF800000 : f32` → f32 -inf (bit pattern reinterpretation)

## Suggested fix

In `parse_numeric`, when the declared type is `fN`, reinterpret the hex
value as an IEEE 754 bit pattern via `np.array([val], dtype=np.uintN).view(np.floatN)`.
This is the same approach already used in `arith.bitcast`.

## Impact

Blocks any kernel using hex float constants for -inf initialization
(common in softmax and reduction patterns).

fabianlim:
Fixes #19, cc: @ohadeytan

## Summary

- **Hex float constant parsing** (`parse_numeric`): hex literals like `0xFF800000 : f32` and `0xFC00 : f16` are IEEE 754 bit patterns. Previously parsed as raw unsigned ints, causing OverflowError (f32) or wrong values (f16). Now reinterprets via `np.view` when the declared type is float. Integer types keep the plain int behavior.
- **`arith.truncf` was a no-op**: tiles stayed f32 after "truncation". When stored to f16 memrefs, the 4-byte itemsize caused wrong offset calculations in `_write_flat`, corrupting memory for all subsequent rows.

## Test plan

- [x] All 481 ktir-cpu tests pass


ohadeytan:
Thanks @fabianlim ! I can confirm that with this PR all the tests in Tritokti are passing :)

ktir_cpu/parser_utils.py
```diff
@@ -41,16 +41,28 @@ def parse_tensor_type(type_str: str) -> Optional[Dict]:
...
-        return int(s, 16)
+        bits = int(s, 16)
+        # Float types: reinterpret hex as IEEE 754 bit pattern (e.g. 0xFC00 : f16 → -inf)
+        # Integer/index types: keep as plain int (e.g. 0xFF800000 : i32 → 4286578688)
+        if dtype == 'f32':
+            return float(np.array([bits & 0xFFFFFFFF], dtype=np.uint32).view(np.float32)[0])
+        elif dtype == 'f16':
+            return float(np.array([bits & 0xFFFF], dtype=np.uint16).view(np.float16)[0])
+        elif dtype == 'bf16':
+            return float(np.array([bits & 0xFFFF], dtype=np.uint16).view(np.float16)[0])
```
kiszk: The original code is for f16
```suggestion
            bits_32 = (bits & 0xFFFF) << 16
            return float(np.array([bits_32], dtype=np.uint32).view(np.float32)[0])
```

ktir_cpu/ops/arith_ops.py
```diff
@@ -247,16 +247,19 @@ def extf(value):
...
+            Value converted to f16, or passthrough if already f16
         """
-        return value
+        if isinstance(value, Tile):
+            if value.data.dtype == np.float16:
+                return value
+            return Tile(value.data.astype(np.float16), "f16", value.shape)
+        if isinstance(value, np.float16):
+            return value
+        return np.float16(value)
```
kiszk: It would be good to handle types a bit strictly. IBM Bob suggested the following code.

```suggestion
        if isinstance(value, Tile):
            if value.data.dtype == np.float16:
                return value
            # Validate input dtype is a float type
            if not np.issubdtype(value.data.dtype, np.floating):
                raise TypeError(
                    f"truncf expects float Tile, got dtype={value.data.dtype}"
                )
            return Tile(value.data.astype(np.float16), "f16", value.shape)
    
        # Handle scalar values
        if isinstance(value, np.float16):
            return value
    
        # Accept numeric types: int, float, np.number
        if isinstance(value, (int, float, np.number)):
            try:
                result = np.float16(value)
                # Check for overflow (inf) or invalid conversion (nan from non-numeric)
                if np.isinf(result) and not np.isinf(float(value)):
                    raise ValueError(
                        f"truncf overflow: {value} exceeds f16 range "
                        f"[{np.finfo(np.float16).min}, {np.finfo(np.float16).max}]"
                    )
                return result
            except (ValueError, TypeError) as e:
                raise TypeError(
                    f"truncf cannot convert {type(value).__name__} to f16: {e}"
                ) from e
    
        raise TypeError(
            f"truncf expects Tile or numeric scalar, got {type(value).__name__}"
        )
```

kiszk:
A comment on bf16 is mandatory, but a comment on truncf is nice-to-have.

fabianlim:
@kiszk Thanks for the review!

- **bf16 fix** (parser_utils.py:64): Applied your suggestion — now shifts into upper 16 bits of a u32 and views as f32, since bf16 shares f32's exponent layout. Added a comment explaining the bit-layout difference vs f16. (decb2c3)

- **truncf type strictness** (arith_ops.py:262): Went a step further and extracted a shared `arith_cast()` helper that all type-conversion ops (`truncf`, `extf`, `sitofp`, `index_cast`) now use. It validates input dtype category, detects overflow on narrowing, and gives clear error messages. This way future conversion ops (fptosi, trunci, exti) get the same checks for free. (3800dd9)

ktir_cpu/parser_utils.py
fabianlim: Applied — also added a comment explaining the bf16 vs f16 bit-layout difference. (decb2c3)

ktir_cpu/ops/arith_ops.py
fabianlim: Applied and generalized into a shared `arith_cast()` helper — covers truncf, extf, sitofp, and index_cast. (3800dd9)

kiszk (approved):

ktir_cpu/parser_utils.py
```diff
@@ -41,16 +41,36 @@ def parse_tensor_type(type_str: str) -> Optional[Dict]:
...
-    Handles: integers, floats, scientific notation, hex floats (0xFC00).
+    Handles: integers, floats, scientific notation, hex constants.
+    When *dtype* is a float type (f16, f32, bf16) and the literal is hex,
+    the value is reinterpreted as an IEEE 754 bit pattern.
+
+    All float types are widened to Python ``float`` (64-bit double) and
+    all integers to Python ``int`` (arbitrary precision) — the caller
+    is responsible for narrowing to the target dtype at use-site.
     """
+    import numpy as np
```
lchu6, line 55: weird location for the import.

lchu6:
good fix, it would be better to have some unit tests along with it.

I do see you have end to end softmax test that could capture this, but some smaller tests paired with a fix PR would be better.  e.g. simply add tests for the two examples you mentioned in the original issue.

lchu6:
another one on test:

the newly added `arith_cast` is a good function here, and it could use couple tiny tests due to its nature:
1. it is consumed by multiple downstream function/consumers.
2. it explicitly has multiple `raise Error` paths - _if you explicitly write the raise, you believed that case was reachable and worth distinguishing. A test is what proves you were right._

Issue #1, opened 2026-04-22
[RFC conformance] Implement ktdp.construct_distributed_memory_view


fabianlim:
## 🚀 Feature

Add parser and runtime support for `ktdp.construct_distributed_memory_view`,
which composes multiple per-partition memory views into a single distributed
logical view.

## Motivation

`construct_distributed_memory_view` is the second large hole in RFC `ktdp`
surface coverage. It is required for modeling distributed scratchpads — e.g.
a tensor whose rows are striped across HBM partitions or across LX
scratchpads on different cores. Without it, multi-partition memory layouts
cannot be expressed in KTIR at all.

## Pitch

- [ ] Add a parser for `ktdp.construct_distributed_memory_view` (syntax to
  be confirmed against [RFC](https://github.com/torch-spyre/RFCs/tree/main/0682-KtirSpec/examples); expected: a list of constituent
  memory view operands plus a distribution affine map)
- [ ] Add `DistributedTileRef` in `ktir_cpu/ir_types.py` wrapping multiple
  `TileRef` objects and a partition map
- [ ] Update `construct_access_tile` to handle `DistributedTileRef` as
  parent, routing each coordinate tuple to the correct partition
- [ ] Promote `examples/rfc/distributed_view.mlir` spec-gap test
   from xfail to passing
- [ ] Add a unit test with a small 2-partition distributed view verifying
  load/store correctly routes to each partition

## Additional context

- Full gap analysis: [`docs/gap_analysis.md`](https://github.ibm.com/msrivats/ktir_cpu/blob/main/docs/gap_analysis.md)
- RFC [reference](https://github.com/torch-spyre/RFCs/tree/main/0682-KtirSpec/examples): (`examples/rfc/distributed_view.mlir`)

cc @mudhakar @raghukiran1224

PR #21, opened 2026-05-01
Add missing interpreter ops: arith.minimumf, arith.cmpf, tensor.generate

fabianlim:
cc: @ohadeytan

> **Important:** This PR depends on [#20](https://github.com/torch-spyre/ktir-cpu/pull/20) — merge that first, then rebase this onto main.

## Summary

- Add `arith.minimumf` / `arith.minnumf` handlers (mirrors existing `maximumf` / `maxnumf`)
- Fix `arith.maxnumf` / `arith.minnumf` to use `np.fmax` / `np.fmin` (NaN non-propagating), not `np.maximum` / `np.minimum` which propagate NaN
- Add `arith.cmpf` handler and parser with ordered (`oeq`, `oge`, ...) and unordered (`ueq`, `uge`, ...) float comparison predicates
- Implement `tensor.generate` and `tensor.yield` — evaluates a region body at each index to produce a tensor
- Fix parser to preserve trailing text after `}` of a region body (e.g. `: tensor<16x16xf16>`)
- Refactor `^bb0` block-arg parsing: rename from `linalg.bb0_args` to `region.bb0_args`, move parser to `scf_ops.py` (shared across `linalg.generic`, `tensor.generate`, and future region-based ops)
- Add TODO to consider deprecating `arith.maxf` / `arith.minf` aliases

## Motivation

These are standard MLIR ops that kernels need but the interpreter was missing:

| Op | Use case |
|----|----------|
| `arith.minimumf` | Element-wise minimum (clamping, bounds) |
| `arith.cmpf` | Float comparison with IEEE 754 NaN semantics |
| `tensor.generate` | Build tensors from index arithmetic (masks, index tables) |

## Not addressed

- `construct_distributed_memory_view` — skipped, larger scoped change (tracked in [torch-spyre/ktir-cpu#1](https://github.com/torch-spyre/ktir-cpu/issues/1)).


lchu6:
somehow I cannot start a review and inline commenting also stopped working.  so I will write things here

lchu6:
1.

> Fix arith.maxnumf / arith.minnumf to use np.fmax / np.fmin (NaN non-propagating), not np.maximum / np.minimum which propagate NaN

you claim is right but your code change (using `np.max` and `np.min`) conflicts with what you claimed here.  maybe typo?  should use a function that does not propagate nan.

lchu6:
2. [this list of supported op](https://github.com/fabianlim/ktir-cpu/blob/5862ee4fb58c5829daecf2e92fe655799dd61244/ktir_cpu/dialects/arith_ops.py#L284-L295) does not agree with [this list](https://github.com/fabianlim/ktir-cpu/blob/5862ee4fb58c5829daecf2e92fe655799dd61244/ktir_cpu/mlir_frontend/parser.py#L492-L495) for arith.
should we make them consistent?  e.g. remove true and false in the later list if we never meant to use true or false for cmpf?

lchu6:
3.

> if 'return' in text or '.yield' in text:

can we think of something more robust here?  custom_named_ssa might appear in:
1. human-written ktir (could happen)
2. we have some sort of name-inference from passes.

in either case, most likely no but still theoretically you could hit some SSA like  `%sum_returned_val`

lchu6:
4. 

> for idx in itertools.product(*(range(s) for s in shape)):

could this be slow?   if shape is decently large, say, 4k * 4k, can this overhead be big?

lchu6 (changes requested):
inline commenting somehow not working at the moment. I left some comments above.

fabianlim:
@lchu6 Thanks for the thorough review! All four points addressed:

1. **linalg.reduce NaN semantics** — Good catch, fixed. `arith.maxnumf` / `arith.minnumf` now map to `np.fmax.reduce` / `np.fmin.reduce` (NaN non-propagating), consistent with the element-wise `ArithOps` methods. (752c80c)

2. **cmpf predicate map** — Added `"false"` and `"true"` predicates to the executor so it's consistent with the MLIR frontend adapter's full predicate map. (752c80c)

3. **`_is_op_complete` false match** — Replaced the substring check with a regex that matches `return` and `*.yield` as op names only. Added a negative test for SSA names like `%sum_returned_val`. (752c80c)

4. **`tensor.generate` loop performance** — Great catch, this was unnecessarily O(prod(shape)) in Python. Vectorized it using meshgrid — block args are always index-typed, so we pass full index grids as Tiles and execute the body once. The arith/compute ops already handle Tile inputs element-wise, so the whole tensor is computed in a single numpy pass regardless of shape. (c326c30)

lchu6 (dismissed):
LG.

ktir_cpu/dialects/arith_ops.py
```diff
@@ -252,6 +268,40 @@ def arith__cmpi(op, context, env):
...
+        "ugt": lambda: (lhs > rhs)  | (np.isnan(lhs) | np.isnan(rhs)),
+        "uge": lambda: (lhs >= rhs) | (np.isnan(lhs) | np.isnan(rhs)),
+        "ord": lambda: ~(np.isnan(lhs) | np.isnan(rhs)),
+        "uno": lambda: np.isnan(lhs) | np.isnan(rhs),
+        "true": lambda: np.ones_like(lhs, dtype=bool) if is_tile else True,
+    }
+    if predicate not in cmp_ops:
+        raise NotImplementedError(f"arith.cmpf: unsupported predicate '{predicate}'")
+    result = cmp_ops[predicate]()
+    return Tile(result, (a if isinstance(a, Tile) else b).dtype, result.shape) if is_tile else result
```
kiszk: Question: if a return is a tile, dtype looks float type (e.g. `f16`) since `a.dtype` or `b.dtype` will be used. I think the type is `bool` in Python or `i1` in MLIR. Am I wrong?

ktir_cpu/dialects/arith_ops.py
```diff
@@ -252,6 +268,40 @@ def arith__cmpi(op, context, env):
...
+        "oeq": lambda: lhs == rhs,  "one": lambda: (lhs != rhs) & ~(np.isnan(lhs) | np.isnan(rhs)),
+        "olt": lambda: lhs < rhs,   "ole": lambda: lhs <= rhs,
+        "ogt": lambda: lhs > rhs,   "oge": lambda: lhs >= rhs,
+        "ueq": lambda: (lhs == rhs) | (np.isnan(lhs) | np.isnan(rhs)),
+        "une": lambda: lhs != rhs,
+        "ult": lambda: (lhs < rhs)  | (np.isnan(lhs) | np.isnan(rhs)),
+        "ule": lambda: (lhs <= rhs) | (np.isnan(lhs) | np.isnan(rhs)),
+        "ugt": lambda: (lhs > rhs)  | (np.isnan(lhs) | np.isnan(rhs)),
+        "uge": lambda: (lhs >= rhs) | (np.isnan(lhs) | np.isnan(rhs)),
+        "ord": lambda: ~(np.isnan(lhs) | np.isnan(rhs)),
```
kiszk:
```suggestion
        "ord": lambda: not (np.isnan(lhs) | np.isnan(rhs)),
```

`~` in Python is bitwise NOT. So, `~ True` will return `-2`.

ktir_cpu/dialects/arith_ops.py
fabianlim: Good catch! Fixed in 73d2802 — comparison ops (cmpi and cmpf) now return Tile with dtype "i1" (added to SUPPORTED_DTYPES as np.bool_) instead of copying the input operand's float dtype.

ktir_cpu/dialects/arith_ops.py
fabianlim: Nice catch! Fixed in 73d2802 — introduced a `_bool_not` helper that dispatches to `~` for numpy arrays and `not` for Python scalars. Used it for both `ord` and `one` predicates.

ktir_cpu/dialects/arith_ops.py
```diff
@@ -411,6 +465,37 @@ def parse_arith_cmpi(op_text, parse_ctx):
...
+    result_match = re.match(r'(%\w+)\s*=\s*arith\.cmpf\s+', op_text)
+    if not result_match:
+        return None
+
+    result_name = result_match.group(1)
+
+    predicate = "olt"
+    pred_match = re.search(
+        r'arith\.cmpf\s+(oeq|one|olt|ole|ogt|oge|ueq|une|ult|ule|ugt|uge|ord|uno)', op_text)
+    if pred_match:
```
kiszk: Is it ok to use `olt` if no predication is found. How about throwing an exception?

fabianlim: Fixed in 2916542 — both `arith.cmpf` and `arith.cmpi` parsers now raise a `ValueError` if no valid predicate is found, instead of silently defaulting. Also removed the fallback defaults from the executors for consistency.

kiszk (approved):

Issue #14, opened 2026-04-28
[Feature, Bug] HBM Memory Model: Stick Addressing and Latency Correctness


fabianlim:
cc: @mudhakar @Prasanth-Chatarasi @moriohara  

## Background

Spyre HBM is organised into sticks. The hardware addressing breaks down as:

- **3-bit segment index** — 8 segments
- **27-bit intra-segment stick index** — 2^27 sticks per segment
- **Total stick address: 30 bits → 2^30 sticks**
- **Each stick: 128 B = 2^7 bytes**
- **Total addressable bytes: 2^37 = 128 GB** ✓ (matches `HBMSimulator` default)

The 30-bit stick address fits in a signed `int32` (max ~2.1 B; stick addresses go up to ~1 B), so `int32` is sufficient for stick-indexed base pointers.

---

## Issue 1 — Gather data traffic penalty is unmodelled for scattered accesses

**File:** `ktir_cpu/latency.py`, `_data_size()` (line ~286)

**Context:** For an indirect (gather) load, `_data_size()` currently counts:

```python
total += result.data.nbytes   # e.g. 64×64×2 = 8,192 B for f16
```

This is correct when the gathered elements are **contiguous** in HBM — 8,192 B / 128 B per stick = 64 sticks, all fully utilised. The stick boundary introduces no overhead.

**Problem:** When gather indices are **scattered** (each output element lands in a different stick), the hardware must load a full 128 B stick per element even though only `element_size` bytes are used. In the worst case the actual data traffic is:

```
n_elements × 128 B   (one stick per element, none shared)
```

vs. the best case:

```
result.data.nbytes   (all elements packed into the minimum number of sticks)
```

For a 64×64 f16 gather with fully scattered indices the ratio is 128 / 2 = **64×**. The latency model currently always assumes the best case and has no way to express anything in between, since it does not inspect the runtime index values.

**Proposed fix:** Since this is a simulator that actually executes the kernel, the index values are available at runtime. `indirect_load()` in `memory_ops.py` already builds the full `coords` list — every N-dimensional coordinate tuple into `parent_ref` that the gather will read. From those coordinates, the exact number of distinct sticks touched can be computed:

```python
# inside indirect_load(), after coords is built:
ndim = len(parent_ref.strides)
bytes_per_elem = _bytes_per_elem(parent_ref.dtype)
sticks = {
[...]

PR #24, opened 2026-05-02
Switch ktir-mlir-frontend dependency from fork to upstream

fabianlim:
## Summary

- Point `mlir-frontend` extra at `torch-spyre/ktir-mlir-frontend` instead of the `fabianlim` fork pin, now that [torch-spyre/ktir-mlir-frontend#14](https://github.com/torch-spyre/ktir-mlir-frontend/pull/14) is merged.
- CI and README now fetch `setup_mlir.py` and `llvm-hash.txt` directly from upstream main (no hardcoded commits).
- `uv.lock` pins the resolved commit (`c2412cdd`).

## Test plan

- [x] `uv lock` resolves successfully
- [x] `uv sync --extra mlir-frontend` builds ktir-mlir-frontend from upstream
- [x] Full test suite passes (520 passed, 6 xfailed)
- [x] CI passes on this PR


fabianlim:
cc @kiszk @lchu6

.github/workflows/ci.yml
```diff
@@ -24,31 +24,16 @@ jobs:
...
-          curl -fsSL \
-            "https://raw.githubusercontent.com/$KTIR_MLIR_REPO/ktir-mlir-frontend/$KTIR_MLIR_FRONTEND_COMMIT/scripts/setup_mlir.py" \
-            -o /tmp/setup_mlir.py
-          MLIR_DIR=$(uv run python /tmp/setup_mlir.py --wheel --hash "$LLVM_HASH")
-          # Pre-install build-system deps for ktir-mlir-frontend so the source build
-          # can run with --no-build-isolation (mlir_wheel is not on PyPI so uv cannot
-          # resolve build-system.requires in an isolated env; scikit-build-core and
-          # nanobind are on PyPI and safe to install here).
+          SETUP_MLIR="https://raw.githubusercontent.com/torch-spyre/ktir-mlir-frontend/main/scripts/setup_mlir.py"
+          LLVM_HASH=$(curl -fsSL "https://raw.githubusercontent.com/torch-spyre/ktir-mlir-frontend/main/cmake/llvm-hash.txt")
```
lchu6: should we use a pinned version (which leads to a pinned llvm) here rather than a floating version?

README.md
```diff
@@ -28,20 +28,22 @@ Until a PyPI release is available, build from source:
...
+Resolve `MLIR_DIR` using one of:
 
-# Fetch setup_mlir.py from the pinned commit
-curl -fsSL \
-  "https://raw.githubusercontent.com/fabianlim/ktir-mlir-frontend/$KTIR_MLIR_FRONTEND_COMMIT/scripts/setup_mlir.py" \
-  -o /tmp/setup_mlir.py
+```bash
+# Option 1: pin to the LLVM hash tested by ktir-mlir-frontend
+SETUP_MLIR="https://raw.githubusercontent.com/torch-spyre/ktir-mlir-frontend/main/scripts/setup_mlir.py"
+LLVM_HASH=$(curl -fsSL "https://raw.githubusercontent.com/torch-spyre/ktir-mlir-frontend/main/cmake/llvm-hash.txt")
```
lchu6: same here.

README.md
```diff
@@ -28,20 +28,22 @@ Until a PyPI release is available, build from source:
...
+MLIR_DIR=$(curl -fsSL "$SETUP_MLIR" | uv run python - --wheel --hash "$LLVM_HASH")
+
+# Option 2: use the latest mlir_wheel (simplest, no pinned hash)
+uv pip install mlir_wheel --find-links https://llvm.github.io/eudsl
```
lchu6, line 42: same here, and it can be more floating here.

lchu6 (commented):
It seems in this change, we are using floating llvm hash from upstream latest main, rather than any pinned hash (say, from a pinned upstream).

I think it is better to use pinned version.  But since you purposely made this change I assume you had that in mind but decided to change the design. one possible reason I could think of:   if llvm version is validated by frontend, then we can safely use any floating version.

But even this is the case, your uv currently pulls a pinned upstream, which can also cause drift.

.github/workflows/ci.yml
fabianlim: The LLVM version is pinned upstream in `ktir-mlir-frontend` — it's the repo that builds against LLVM, so that's where the pinning belongs. Upstream [PR #14](https://github.com/torch-spyre/ktir-mlir-frontend/pull/14) established this: `cmake/llvm-hash.txt` is pinned to `llvmorg-22.1.3` and only moves when explicitly bumped. Since we control upstream `main`, reading from it here is equivalent to reading a pinned value — we're just avoiding duplicating that pin in this repo.

README.md
fabianlim: Same as above — the pinning is done upstream (which is the repo that actually builds against LLVM). We just read the pinned value from there.

README.md
fabianlim: Yes, Option 2 is intentionally floating — it's for convenience (quick local dev) or testing the frontend against nightly LLVM. It is not meant for reproducibility; that's what Option 1 is for.

fabianlim:
Re: the general review comment on pinning — yes, the LLVM version is tied to the upstream repo because `ktir-mlir-frontend` is the repo that builds against LLVM, so the pinning naturally lives there. Upstream [PR #14](https://github.com/torch-spyre/ktir-mlir-frontend/pull/14) explicitly pins to `llvmorg-22.1.3` in `cmake/llvm-hash.txt` and documents three flows (pinned artifact, LLVM rebuild on hash bump, bleeding-edge fallback).

This repo just reads that pinned value rather than duplicating it. The `uv.lock` separately pins the frontend Python package commit (`c2412cdd`) — these are two orthogonal concerns (which LLVM toolchain build vs. which frontend Python code) and both are pinned in their respective upstream locations.

fabianlim:
To preempt a possible concern about drift between `uv.lock` (pinned frontend commit) and the LLVM hash (read from upstream `main`): this cannot cause a real incompatibility. The upstream repo's own CI validates that its `main` branch builds and tests correctly against the LLVM hash in `cmake/llvm-hash.txt`. So any commit on upstream `main` — including the one `uv.lock` pins to — is guaranteed to be compatible with the LLVM hash that `main` advertises. The two can only "drift" in the sense that `main` might have newer commits than what `uv.lock` resolves, but the LLVM hash will still be valid for building the locked commit (it only ever moves forward to a newer compatible release, never backward).

lchu6:
> yes, the LLVM version is tied to the upstream repo because ktir-mlir-frontend is the repo that builds against LLVM, so the pinning naturally lives there. 

there is no question regarding this.  the question is more about:
we tie to a dynmaic (always main) upstream's pinned version or a pinned upstream's pinned version.

I understand what you are saying and most liklely in practice this won't be a problem, but I still don't think thats a good idea to do versioning like this. a drifting is a drifting, you cannot say a drifting isn't a problem because "this software will always guarantee backward compatibility", which I don't even know if it is true.

what happens if llvm introduce a breaking change, and our upstream frontend adopted that and also made some breaking changes, won't this break your local locked commit?    less likely happen but not impossible.

in any case, I think locking in frontend should naturally use the locked frontend's pinned version.  mixing a commit and laltest main isn't a sound solution even in practice the chance of failure is low to zero.

lchu6:
unless the claim is that this repo will always use latest main of upstream, and a locked commit is only for dev or debugging.   in such case, it is valid.   but is it a good practice to "always use latest main"?

for the dependency tree, it is still a good practice to pin versions on all items on the chain.

lchu6:
but just to get some clarification first:

are we positioning this repo to **always target latest main from upstream frontend?**

fabianlim:
To clarify the positioning: this repo has a partial dependency on the frontend (the MLIR-based parser), alongside a regex parser we also maintain. To keep both parsers coherent, we track upstream frontend as closely as possible for parsing behavior changes.

That said, the upstream dependency here is really only two files: `setup_mlir.py` and `llvm-hash.txt` — and these should rarely change since the LLVM version is locked upstream. Even if parsing behavior evolves in the frontend, these two files stay stable.

The risk scenario would be: LLVM introduces a breaking change, upstream frontend adopts it (bumping `llvm-hash.txt`), and that also changes parsing behavior in a breaking way. In that case yes, floating on `main` would be a problem.

To be defensive, I'll freeze the hash in this repo so we have an explicit control point. Will push a follow-up.

lchu6:
> The risk scenario would be: LLVM introduces a breaking change, upstream frontend adopts it (bumping llvm-hash.txt), and that also changes parsing behavior in a breaking way. In that case yes, floating on main would be a problem.

this is precisely what I meant.  again, low-to-zero chance of failure due to the nature of the repos as you mentioned, but still good to have it under control.


> I'll freeze the hash in this repo so we have an explicit control point

why not simply add a commit when pulling upstream hash?  we can also avoid duplicated copies of hash.

lchu6:
upstream does not yet have a releasing control yet I believe.  so for now we can use a commit?

maybe have a commit defined somewhere (just like a version number), then all three places pull the same commit?

And in the future when upstream frontend does have a commit, we can swap that in.

fabianlim:
Addressed in 6e4bf5c: pinned the frontend commit in `pyproject.toml` (single source of truth) and updated CI + README to parse the commit hash from there rather than fetching from floating `main`. This way `uv.lock`, the LLVM hash, and `setup_mlir.py` all derive from the same pinned commit — no drift possible.

lchu6:
I pushed a PR into your branch.

It is not meant to ask you to merge my chances, but I find a PR diff easier to illustrate what I have in mind for potential code change on this.

Feel free to close that one, it was just a better-than-text way of describing what kind of change in my mind.

lchu6:
great. same solution, didn't see you already pushed the solution.

lchu6 (approved):

fabianlim:
@lchu6 oh im sorry I didnt see that you pushed a PR. thank you for approving.

PR #27, opened 2026-05-05
Add all missing math dialect ops

fabianlim:
## Summary

- Add 12 missing math ops: `rsqrt`, `log2`, `log1p`, `tanh`, `sin`, `cos`, `absf`, `ceil`, `floor`, `erf`, `powf`, `fma`
- `math.erf` uses Abramowitz & Stegun polynomial approximation to avoid a scipy dependency
- Register all ops in the mlir_frontend `_no_attrs` adapter list
- Update gap_analysis.md: items #20–24 now ✅

## Test plan

- [x] 28 execution tests (`TestMath` in `test_dialects_exec.py`) — tile + scalar for each unary op, plus powf/fma
- [x] 15 regex parser tests (`TestMathParsers` in `test_dialects_parse.py`) — parametrized unary + binary + ternary
- [x] 15 mlir_frontend adapter tests (`TestMathAdapt` in `mlir_frontend/test_parse_adapt.py`)
- [x] Full suite: 515 passed, 1 skipped, 6 xfailed


ktir_cpu/dialects/math_ops.py
```diff
@@ -36,9 +36,104 @@ def math__sqrt(op, context, env):
...
+def math__erf(op, context, env):
+    operand = context.get_value(op.operands[0])
+    if isinstance(operand, Tile):
+        return MathOps.erf(operand)
+    return MathOps.erf_scalar(operand)
+
+
+@register("math.powf", latency_category=LC.COMPUTE_TRANSCENDENTAL)
+def math__powf(op, context, env):
+    base = context.get_value(op.operands[0])
```
lchu6, line 137: any reason all other ops have `operand = context.get_value(op.operands[0])` except this one?

ktir_cpu/dialects/math_ops.py
```diff
@@ -36,9 +36,104 @@ def math__sqrt(op, context, env):
...
+    exponent = context.get_value(op.operands[1])
+    return MathOps.powf(base, exponent)
+
+
+@register("math.fma", latency_category=LC.COMPUTE_FLOAT)
+def math__fma(op, context, env):
+    a = context.get_value(op.operands[0])
```
lchu6, line 146: this one as well.

ktir_cpu/dialects/math_ops.py
```diff
@@ -36,9 +36,104 @@ def math__sqrt(op, context, env):
...
+    if isinstance(operand, Tile):
+        return MathOps.cos(operand)
+    return MathOps.cos_scalar(operand)
+
+
+@register("math.absf", latency_category=LC.COMPUTE_FLOAT)
+def math__absf(op, context, env):
+    operand = context.get_value(op.operands[0])
+    if isinstance(operand, Tile):
+        return MathOps.abs(operand)
```
lchu6: should this be abs or absf?   I don't know, but you name it math__absf but call a abs?

it might be correct but I am pointing out some inconsistency I observe.

lchu6: there could be a reason for this but leaving a note so you can double check if you meant to do both abs and absf as two functions.

lchu6 (commented):
LG. some minors left for you to double check.

ktir_cpu/dialects/math_ops.py
fabianlim: `math.powf` is a binary op (`base`, `exponent`), so it needs two `context.get_value` calls — one per operand. Unlike the unary ops above, there's no single `operand` to extract.

ktir_cpu/dialects/math_ops.py
fabianlim: Same as above — `math.fma` is ternary (`a`, `b`, `c`), requiring three `context.get_value` calls.

lchu6: got it.

ktir_cpu/dialects/math_ops.py
lchu6: got it.

ktir_cpu/dialects/math_ops.py
fabianlim: Good catch — `math.absf` and `math.absi` are distinct ops in the MLIR math dialect (float vs integer absolute value), so the naming matters. Fixed in e1168b9: renamed `MathOps.abs` → `MathOps.absf` and added the missing `MathOps.absi` / `math.absi` handler.

ktir_cpu/dialects/math_ops.py
lchu6: actually， I copied the wrong line.

I meant to ask: 

any reason all other ops have `isinstance(operand, Tile)` except this one and the one below?

fabianlim: You're right — `powf` and `fma` were missing the `isinstance(operand, Tile)` branch and had no scalar variants, inconsistent with all other ops. Fixed in 6958c80: added `powf_scalar` / `fma_scalar` to `MathOps` and updated both handlers to dispatch on tile vs scalar.

lchu6 (dismissed):

lchu6 (approved):

PR #29, opened 2026-05-05
latency: charge per-stick HBM traffic for gather loads

WarningRan:
Addresses #14 - this PR implements **Issue 1** (gather latency) only.

## Summary

- Latency now charges gather (`ktdp.load` with `IndirectAccessTile`) HBM traffic as `unique_sticks × 128` instead of `result.data.nbytes`. The old path silently assumed best-case packing and under-counted fully scattered accesses by up to 64× (128 / bpe for f16).
- `Tile.unique_sticks: Optional[int]` sideband set by `indirect_load()`, consumed by `_data_size()`. Contiguous loads leave it `None` and keep the existing `data.nbytes` path - no regression in non-gather latency.
- `MemoryOps._count_unique_sticks(parent_ref, coords)` extracted as a private static helper, reusing the existing `_gather_indices` so the element-offset formula is not duplicated.
- `Tile.coalescing_efficiency` property exposes `data.nbytes / (unique_sticks × 128)` for analysis (`None` for non-gather tiles).
- `Tile.copy()` deliberately does **not** propagate `unique_sticks`: copies are produced by comm ops for message buffers / reduce results, and re-charging them as gathers would double-count.

## Motivation

Before this change, `_data_size()` counted packed bytes regardless of access pattern:

```python
# latency.py _data_size() - before
total += result.data.nbytes     # 64×64 f16 gather = 8,192 B regardless of indices
```

Real Spyre HBM reads at 128-byte stick granularity. Scattered gather indices pull a full stick per element, giving worst-case traffic of `n_elements × 128`. Because the interpreter already executes the kernel, the resolved `coords` list is available — the fix counts the distinct sticks those coords land in:

```python
# memory_ops.py — new helper
@staticmethod
def _count_unique_sticks(parent_ref, coords):
    offsets = MemoryOps._gather_indices(parent_ref.shape, parent_ref.strides, coords)
    bpe = _bytes_per_elem(parent_ref.dtype)
    return len({
        (parent_ref.base_ptr + offset * bpe) // 128
        for offset in offsets
    })
```

## Changes

- `ktir_cpu/ir_types.py` — `Tile.unique_sticks` field, `coalescing_efficiency` property, `copy()` docstring explaining why `unique_sticks` is reset.
- `ktir_cpu/ops/memory_ops.py` — `MemoryOps._count_unique_sticks` helper; `indirect_load()` sets `tile.unique_sticks` before returning.
- `ktir_cpu/latency.py` — `_data_size()` branches on `unique_sticks` when set; index-tensor byte accounting unchanged.

## Test changes

New tests in `tests/test_latency.py::TestIndirectAccessLatency`:

- [x] `test_count_unique_sticks_returns_n_when_every_element_lands_on_its_own_stick` — worst-case boundary, `unique_sticks = n_elements`
- [x] `test_count_unique_sticks_dedups_sticks_shared_by_multiple_gather_reads` — set-vs-list dedup regression guard, `unique_sticks = 2` strictly between 1 and 6
- [x] `test_data_size_uses_unique_sticks_for_gather_result` — `_data_size()` returns `unique_sticks × 128`, not `data.nbytes`
- [x] `test_coalescing_efficiency_returns_bpe_over_stick_for_worst_case` — `2/128` for f16 fully scattered (matches the table in #14)
- [x] `test_coalescing_efficiency_is_none_for_non_gather_tile` — `None` short-circuit branch
- [x] `test_copy_does_not_propagate_unique_sticks` — guards the invariant that copies aren't re-charged as gathers

Updated test:

- [x] `test_indirect_load_includes_index_tensor_bytes` — expected `X` gather traffic updated 8192 → 128 to match the new model (zero-seeded indices → `unique_sticks = 1`); docstring rewritten.

Full suite: **487 passed, 1 skipped, 6 xfailed** (baseline 481; +6 new).

cc: @fabianlim

### DIFF
--- a/ktir_cpu/ir_types.py
+++ b/ktir_cpu/ir_types.py
@@ -39,14 +39,41 @@ class Tile:
     data: np.ndarray
     dtype: str  # see dtypes.SUPPORTED_DTYPES
     shape: Tuple[int, ...]
+    # Number of distinct 128-byte HBM sticks touched when this tile was
+    # produced by an indirect (gather) load.  None for contiguous loads
+    # and for tiles produced by compute ops.  Consumed by the latency
+    # tracker to model scatter penalty: actual HBM traffic is
+    # ``unique_sticks * 128`` rather than ``data.nbytes`` when set.
+    unique_sticks: Optional[int] = None
 
     def copy(self) -> 'Tile':
-        """Create a deep copy of this tile."""
+        """Create a deep copy of this tile.
+
+        ``unique_sticks`` is intentionally NOT propagated: the field marks
+        a freshly-gathered tile so latency can charge stick-granular HBM
+        traffic.  A copy lives in a different context (e.g. comm buffers)
+        and should not be re-charged as a gather result.
+        """
         return Tile(self.data.copy(), self.dtype, self.shape)
 
     def size_bytes(self) -> int:
         """Return size in bytes."""
         return self.data.nbytes
+
+    @property
+    def coalescing_efficiency(self) -> Optional[float]:
+        """Ratio of packed traffic to actual stick traffic for gather results.
+
+        Defined as ``data.nbytes / (unique_sticks * 128)``.  Ranges from
+        ``bytes_per_elem / 128`` (worst: every element on its own stick, no
+        sharing) to ``1.0`` (best: sticks fully packed).
+
+        Returns ``None`` when this tile is not a gather result
+        (``unique_sticks is None``) or has zero sticks.
+        """
+        if self.unique_sticks is None or self.unique_sticks == 0:
+            return None
+        return self.data.nbytes / (self.unique_sticks * 128)
 
 
 @dataclass
--- a/ktir_cpu/ir_types.py
+++ b/ktir_cpu/ir_types.py
@@ -39,22 +39,21 @@ class Tile:
     data: np.ndarray
     dtype: str  # see dtypes.SUPPORTED_DTYPES
     shape: Tuple[int, ...]
-    # Number of distinct 128-byte HBM sticks touched when this tile was
-    # produced by an indirect (gather) load.  None for contiguous loads
-    # and for tiles produced by compute ops.  Consumed by the latency
-    # tracker to model scatter penalty: actual HBM traffic is
-    # ``unique_sticks * 128`` rather than ``data.nbytes`` when set.
+    # Number of distinct HBM sticks touched by the load that produced this
+    # tile.  Set by ``MemoryOps.load`` for all access patterns.  None only
+    # for tiles produced by compute ops (not loaded from memory).
     unique_sticks: Optional[int] = None
 
     def copy(self) -> 'Tile':
         """Create a deep copy of this tile.
 
-        ``unique_sticks`` is intentionally NOT propagated: the field marks
-        a freshly-gathered tile so latency can charge stick-granular HBM
-        traffic.  A copy lives in a different context (e.g. comm buffers)
-        and should not be re-charged as a gather result.
+        Used by comm ops for message buffers / reduce accumulators.
+        When cross-core communication is fixed (gap analysis K1/K2),
+        reconsider whether unique_sticks should be recomputed for the
+        target device's base_ptr — currently we propagate the source
+        value since the memory layout is abstracted away from Tile.
         """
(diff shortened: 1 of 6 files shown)


fabianlim (commented):
### `128` → named constant

The literal `128` (HBM stick size in bytes) appears in `_count_unique_sticks`, `_data_size`, `coalescing_efficiency`, and across test assertions. Please extract it as a named constant (e.g. `HBM_STICK_BYTES = 128`) at the top of `memory_ops.py` and import it where needed.

WarningRan:
@fabianlim  Thanks for the review! Extracted `HBM_STICK_BYTES = 128` as suggested, with one small placement deviation I wanted to flag:

I placed the constant in `ktir_cpu/dtypes.py` rather than `ktir_cpu/ops/memory_ops.py`. The reason is a circular-import risk:

- `Tile.coalescing_efficiency` (in `ir_types.py`) uses the constant.
- `memory_ops.py` already imports `ir_types.py` (`from ..ir_types import Tile, TileRef`).
- Putting the constant in `memory_ops.py` would require `ir_types.py` to import back from `memory_ops.py`, forming a cycle.

`dtypes.py` is a leaf module (numpy only), already imported by `memory_ops.py` and `latency.py`, and already hosts analogous hardware-granularity constants (`SUPPORTED_DTYPES`, `bytes_per_elem`), so semantically it felt like the natural home.

All four usage sites (`_count_unique_sticks`, `_data_size`, `coalescing_efficiency`, test assertions) now reference the constant; full suite still passes (487 passed, 1 skipped, 6 xfailed). Happy to move it elsewhere if you'd prefer a different layout.

fabianlim:
## Refactored changes (commit `14392aa`)

### Core refactor: `_gather_indices` → `_flat_memory_offsets`

Renamed `_gather_indices` → `_flat_memory_offsets` — it's just coordinate linearization, not specific to indirect/gather access. Extended to compute unique stick counts in a single pass (same loop builds both the offset list and the stick set).

### `unique_sticks` set on all loads, not just indirect

Previously only `indirect_load` post-hoc attached `unique_sticks`. Now `MemoryOps.load()` sets it for all paths:
- **Contiguous fast path**: alignment-aware formula `(end + STICK_BYTES - 1) // STICK_BYTES - base_ptr // STICK_BYTES`
- **Strided / coord-set path**: computed inside `_flat_memory_offsets`

This removes the separate `_count_unique_sticks` helper and the redundant offset recomputation it caused.

### `HBMSimulator.STICK_BYTES = 128` constant

Stick size constant lives on `HBMSimulator` in `memory.py` (where HBM is modeled and the value is already used for allocation alignment). Imported by `memory_ops.py`, `latency.py`, and `ir_types.py`.

### `Tile.copy()` propagates `unique_sticks`

`unique_sticks` is a property of the data's memory footprint, not the creation context. Added a note that this may need revisiting when comm ops are fixed (gap K1/K2) — the target device's base_ptr could differ.

### `Tile` constructor used directly

`unique_sticks` passed as a positional arg in the constructor instead of being set as a separate statement.

### Test updates

- `_count_unique_sticks` tests → `_flat_memory_offsets` tests
- `test_copy_does_not_propagate` → `test_copy_propagates_unique_sticks`
- `test_work_splitting_elementwise` / `test_work_splitting_transcendental`: expected formulas updated to account for stick-granular load traffic
- `test_work_splitting_matmul`: relaxed to verify cross-core consistency (sub-tile alignment within allocations makes exact analytical formulas brittle)

cc @WarningRan

WarningRan:
Took a pass through `90c2ab7` — LGTM.

The unification of `unique_sticks` across all load paths (not just `indirect_load`) is cleaner than the separate code path I had, and the alignment-aware formula in the contiguous fast path catches a boundary case my version missed — a contiguous load whose `base_ptr` isn't stick-aligned would straddle an extra stick, which my `data.nbytes` fallback ignored.

Merging `_count_unique_sticks` into `_flat_memory_offsets` also removes the redundant second pass over coords, and the rename reflects the helper's real job (coordinate linearization, not gather-specific).

On `Tile.copy()` propagating `unique_sticks`: agreed on keeping the information and revisiting when comm_ops lands (gap K1/K2). The docstring TODO is a good marker.

fabianlim (approved):
we have addressed the issues

PR #32, opened 2026-05-06
hbm: migrate HBMSimulator to stick-addressed pointers

WarningRan:
Closes the Issue2 item from #14: `HBMSimulator.allocate()` now returns a stick address instead of a byte address, and all `TileRef.base_ptr` values are in stick units throughout the simulator.

## What changed

**`HBMSimulator.allocate()` returns a stick address.**
`allocate()` returns `ptr // STICK_BYTES` instead of the raw byte pointer. The existing stick-alignment assertion (Part A, already on this branch) guarantees the division is always exact.

**`TileRef`: two-field address representation.**
`base_ptr` is now always a stick number. A new `intra_stick_byte_offset` field (default 0) carries the remainder byte offset within that stick. A `byte_ptr` property reconstructs the raw byte address: `byte_ptr = base_ptr * STICK_BYTES + intra_stick_byte_offset`. Top-level tensors from `allocate()` always have `intra_stick_byte_offset == 0` (guaranteed by the alignment assert). Sub-tiles from `tile_access()` may have a non-zero offset when the element offset doesn't land on a stick boundary.

**`memory_ops.py`:** `tile_access()` splits the byte position into `(stick, intra)`. All `load()`, `store()`, and `indirect_load()` call sites use `tile_ref.byte_ptr` when passing addresses to the underlying HBM/LX dict — the dict itself is unchanged and remains keyed by byte addresses.

**`interpreter.py`:** `input_ptrs` now stores stick addresses as SSA values. `hbm.write()` and `hbm.read()` still receive byte addresses via explicit `stick * STICK_BYTES` conversion.

**MLIR example constants updated:** `indirect-access-copy.mlir` and the inline MLIR in `test_indirect_access.py` had hardcoded byte addresses as `arith.constant` values; these are now stick addresses.

## Points that may look surprising

**Why UNIFORM (no HBM/LX branch in `byte_ptr`).**
The initial plan had `byte_ptr` return `base_ptr` directly for LX TileRefs (since LX has no hardware stick concept). This broke at runtime: both HBM and LX tensor arguments flow through the same `input_ptrs` dict in `interpreter.py`, which stores stick addresses regardless of memory space. If `byte_ptr` did not multiply by `STICK_BYTES` for LX, LX reads would use the stick number as a byte address and hit unmapped memory. The UNIFORM formula (`base_ptr * STICK_BYTES + intra`) is correct for both spaces. This is also consistent with the DMA architecture: L3LU/L3SU transfers for both HBM and LX operate at 128-byte (one stick) granularity.

**`intra_stick_byte_offset` is always 0 for allocator results.**
`allocate()` aligns `next_ptr` to stick boundaries on every call, so top-level `TileRef`s always have `intra_stick_byte_offset == 0`. The field exists for sub-tiles produced by `tile_access()` where element offsets can land mid-stick.

**MLIR constant addresses.**
MLIR files that use `arith.constant N : index` as a base address must now use the stick equivalent (`N // 128`). Two test fixtures were updated; MLIR programs that derive addresses from function arguments (the common case) need no changes.

/cc @fabianlim


ktir_cpu/ir_types.py
```diff
@@ -81,13 +81,19 @@ class TileRef:
...
+      HBM and LX.  ``intra_stick_byte_offset`` carries the remainder byte offset
+      within that stick.  Use ``byte_ptr`` to recover the raw byte address.
     """
     base_ptr: int
     shape: Tuple[int, ...]
     strides: List[int]
     memory_space: str  # "HBM" or "LX"
     dtype: str = "f16"
     coordinate_set: Optional[AffineSet] = None  # parsed; None if omitted in MLIR
+    intra_stick_byte_offset: int = 0  # byte offset within base_ptr stick (0..STICK_BYTES-1)
```
kiszk: It would be good to add the following assert in `__post_init__()`.
```
assert 0 <= intra_stick_byte_offset < STICK_BYTES
```

examples/rfc/indirect-access-copy.mlir
```diff
@@ -18,9 +18,9 @@ module {
   func.func @indirect_access_copy() {
         // In this example, all tensors X, IDX1, IDX2, Y are in a single memory space (HBM)
         %X_addr    = arith.constant 0 : index
-        %IDX1_addr = arith.constant 8192 : index
-        %IDX2_addr = arith.constant 16384 : index
-        %Y_addr    = arith.constant 24576 : index
+        %IDX1_addr = arith.constant 64 : index
```
kiszk: It would be good to add a comment like
```
// stick 64 = byte 8192
```

fabianlim:
Pushed a refactor on top of this PR that introduces a **`MemRef`** type to cleanly separate concerns.

## Motivation

The original PR had a tension: `TileRef.base_ptr` was stick-addressed for HBM but byte-addressed for LX, and callers had to carry `HBMSimulator.STICK_BYTES` everywhere to convert between the two. `intra_stick_byte_offset` was added to handle sub-stick precision, but this made the type more complex without fully resolving the duality.

## Key changes

- **`MemRef`** (hardware-aware): result of `construct_memory_view`. Holds stick index (HBM) or byte address (LX), knows `memory_space`. Provides:
  - `byte_address` property — converts to absolute bytes regardless of space
  - `mem_addr(byte_ptr)` helper — returns `(stick, intra_offset)` tuple for HBM or raw byte address for LX, so callers never do manual divmod
- **`TileRef`** (hardware-agnostic): result of `tile_access()`. Always byte-addressed, used by load/store internals. Holds a back-reference to its parent `MemRef`.
- **`HBMSimulator.read/write`** now accept either `int` (stick index) or `(stick, intra_byte_offset)` tuple, with `_resolve_addr` handling translation internally. Added range assertion on `intra_byte_offset` per @kiszk's suggestion.
- Removed `intra_stick_byte_offset` field from `TileRef` — no longer needed.

This should address @kiszk's comments:
- The assert on `intra_stick_byte_offset` is now in `_resolve_addr` as a range check on the tuple input.
- Added `// stick N = byte M` comments to the MLIR constants.

CC @kiszk @WarningRan

fabianlim:
@kiszk Sorry, I misread your second comment — the MLIR address comment suggestion is still relevant even with the MemRef refactor. Just pushed a follow-up commit adding `// stick N = byte M` annotations to the constants in `indirect-access-copy.mlir`.

WarningRan:
@fabianlim Following up on your MemRef/TileRef refactor — pushed a cleanup for the `_resolve_addr` / `mem_addr()` confusion you mentioned. Ended up being three commits; here's the full picture.

**What was confusing (root causes):**

1. `HBMSimulator.read/write` silently accepted two address formats — `interpreter.py` passed a plain stick `int`, `memory_ops.py` passed a `(stick, intra)` tuple via `MemRef.mem_addr()`. The dispatch was hidden inside `_resolve_addr` behind `isinstance(addr, tuple)`.

2. `MemRef.mem_addr(byte_ptr)` lived on `MemRef` but only used `self.memory_space` — the byte address it converted came from `TileRef`. Wrong object.

3. `TileRef.memref` was `Optional` but every construction site always supplies it. The Optional forced defensive `if v.memref else "HBM"` guards in `latency.py`, and an `isinstance(ref, TileRef)` branch on `AccessTile.parent_ref` that was dead code.

**What changed:**

*Commit 1* — unified HBMSimulator API:
- `HBMSimulator.read/write`: explicit `stick: int, *, intra_byte: int = 0`. `_resolve_addr` deleted.
- `MemRef.mem_addr()` deleted; `TileRef` gains `hbm_addr` property returning `(stick, intra)` — HBM only, asserts if called on LX.
- `memory_ops.py` load/store/indirect_load updated accordingly.

*Commit 2* — TileRef.memref non-optional:
- `TileRef.memref` is now a required field (confirmed all 3 construction sites always pass it).
- `latency.py _memory_space()`: None-guards and dead MemRef branch removed.

*Commit 3* — address unit clarity in load/store:
- `load()` and `store()` now open with an if/else that builds `_read`/`_write` closures once, capturing the correct simulator and address format. No mixed-unit variable used after that point.
- `LXScratchpad` restored to pure byte interface (the `intra_byte=0` param from commit 1 was a bad trade-off).
- `indirect_load()` index lookup: inline if/else per iteration; temp TileRef construction removed.

fabianlim:
Thanks for the review @kiszk!

On the `intra_stick_byte_offset` comment — that field no longer exists in the current state of the PR; the design evolved since those commits. The `intra_byte` concern is now fully encapsulated in a single private abstraction `_MemAccessor` in `memory_ops.py`. It is the only place in the codebase that handles the HBM stick/intra-byte split: `MemRef.split_addr(byte_addr)` produces the `(main, intra)` pair, `_MemAccessor` captures it and forwards the right kwargs to the underlying simulator — callers (`load`, `store`, `indirect_load`) never see stick addresses or `intra_byte` at all.

The MLIR comment suggestion is already addressed — the latest commits add `// stick 64 = byte 8192` style annotations to the constants in `indirect-access-copy.mlir` and the inline MLIR in `test_indirect_access.py`.

/cc @lchu6

fabianlim:
@WarningRan Good analysis of the root causes. The latest commit (`fea6fa0`) takes this one step further on commit 3's concerns:

- `hbm_addr` on `TileRef` is removed — it was HBM-specific naming on a type that's meant to be memory-space-agnostic.
- Address splitting moves to `MemRef.split_addr(byte_addr)` — correct object since `MemRef` owns `memory_space`.
- The `_read`/`_write` closures in `load`/`store` are replaced by `_MemAccessor`, a small private class that captures `(context, memref, byte_addr)` and is the **single place** `intra_byte` is managed. `load`, `store`, and `indirect_load` all go through `_MemAccessor.read/write` — no caller handles `intra_byte` or stick arithmetic directly.

WarningRan:
Agreed - `MemRef.split_addr` is the right home and `_MemAccessor` is a cleaner terminus than commit 3 left. `TileRef.hbm_addr` was leaking an HBM concern onto a space-agnostic type; the HBM-only assert was the tell.

fabianlim:
@WarningRan so if you are ok we can merge this? Shall we go ahead?

fabianlim (approved):
we completed the final refactor - approving

lchu6:
I guess we will need a new PR after https://github.com/torch-spyre/ktir-cpu/pull/110 merged

fabianlim:
> I guess we will need a new PR after https://github.com/torch-spyre/ktir-cpu/pull/110 merged

@lchu6 I think no need, #110 if merged should fill up the remaining gap, which is to make the MemRefs point to elements instead of sticks

examples/rfc/indirect-access-copy.mlir
```diff
@@ -17,10 +17,10 @@
 module {
   func.func @indirect_access_copy() {
         // In this example, all tensors X, IDX1, IDX2, Y are in a single memory space (HBM)
-        %X_addr    = arith.constant 0 : index
-        %IDX1_addr = arith.constant 8192 : index
-        %IDX2_addr = arith.constant 16384 : index
-        %Y_addr    = arith.constant 24576 : index
+        %X_addr    = arith.constant 0 : index    // stick 0 = byte 0
+        %IDX1_addr = arith.constant 64 : index  // stick 64 = byte 8192
```
lchu6, line 21: @fabianlim at least the comments?

`// stick 64 = byte 8192`

PR #33, opened 2026-05-06
Rewind lx.next_ptr on pop_scope (fixes #26)

### LINKED ISSUE #26: [Bug] LX next_ptr never rewound on scope pop — diverges from lx.used capacity tracking
## Summary

The LX scratchpad has two allocation tracking mechanisms that have diverged:

- **`lx.used`** (capacity gate): correctly decremented by `untrack_lx()` when region scopes pop — so `track_lx()` sees free capacity and allows new allocations.
- **`lx.next_ptr`** (address space): never rewound — only reset by `clear()`. Each `_write_to_lx()` call advances `next_ptr` monotonically.

In a loop body that loads tiles, each iteration's `pop_scope()` frees `lx.used`, but `next_ptr` keeps advancing. After enough iterations, `next_ptr` exceeds the 2 MB LX capacity while `lx.used` reports free space. The `Dict[int, np.ndarray]` storage masks this (no out-of-bounds crash), but the simulation no longer faithfully models the hardware 2 MB constraint.

## Reproducible Test

```python
from ktir_cpu.memory import LXScratchpad, HBMSimulator
from ktir_cpu.grid import CoreContext
from ktir_cpu.ir_types import Tile
import numpy as np

def _make_context(lx_size_mb=2):
    lx = LXScratchpad(size_mb=lx_size_mb, core_id=0)
    hbm = HBMSimulator()
    return CoreContext(core_id=0, grid_pos=(0, 0, 0), lx=lx, hbm=hbm)

# 128 KB LX so the bug shows quickly
ctx = _make_context(lx_size_mb=0.125)

for i in range(100):
    ctx.push_scope()
    tile = Tile(data=np.zeros((4, 256), dtype=np.float16), dtype='f16', shape=(4, 256))

    # Simulate _write_to_lx + track_lx
    size = tile.data.nbytes  # 2048
    lx_ptr = ctx.lx.next_ptr
    ctx.lx.next_ptr += size
    ctx.lx.next_ptr = (ctx.lx.next_ptr + 127) & ~127  # 128B align
    ctx.lx.write(lx_ptr, tile.data)
    ctx.track_lx(f'%tile_iter{i}', size)

    ctx.pop_scope()  # frees lx.used but NOT next_ptr

    if ctx.lx.next_ptr > ctx.lx.capacity:
        print(f'BUG at iter {i}: next_ptr={ctx.lx.next_ptr} > capacity={ctx.lx.capacity} but lx.used={ctx.lx.used}')
        break
```

Output:
```
BUG at iter 64: next_ptr=133120 > capacity=131072.0 but lx.used=0
```

## Key Files

- `ktir_cpu/ops/memory_ops.py:118-130` — `_write_to_lx()` advances `next_ptr` without rewind
- `ktir_cpu/grid.py:157-176` — `track_lx()`/`untrack_lx()` manages `lx.used` but not `next_ptr`
- `ktir_cpu/memory.py:291-338` — `LXScratchpad` class; `next_ptr` only reset by `clear()`

## Acceptance Criteria

- [ ] `next_ptr` is rewound or reclaimed when region scopes pop (matching `lx.used` behavior)
- [ ] LX capacity check accounts for both `lx.used` and `next_ptr`
- [ ] Existing LX scoping tests continue to pass
- [ ] New test added that would have caught this divergence

cc @WarningRan @nwang-ibm

nwang-ibm:
## Summary

Fixes #26.

`LXScratchpad` tracked byte usage (`lx.used`) and the bump-pointer
allocation address (`lx.next_ptr`) independently. `pop_scope`
decremented `lx.used` via `untrack_lx` but never touched `next_ptr`,
so any loop that loaded tiles drifted the bump-pointer past capacity
while `lx.used` stayed small — the hardware capacity constraint that
LX is supposed to model was silently bypassed. The issue's reproducer
overflows a 128 KB LX at iter 64 with `lx.used=0`.

## Fix 

`CoreContext` now holds a watermark stack. `push_scope` snapshots
`lx.next_ptr`; `pop_scope` restores it after freeing the scope's
tracked values. `clear_values` also clears the stack so fresh function
invocations start clean.

Four small edits in `ktir_cpu/grid.py`, ~5 lines of production code.
`LXScratchpad`, `_write_to_lx`, the scf/linalg region handlers, and
the interpreter are all untouched — the fix is transparent to them.

## Tests

New `TestNextPtrRewind` in `tests/test_lx_scoping.py`:

- `test_issue_26_reproducer_next_ptr_bounded_in_loop` — the exact
  failure mode from #26. Asserts the strong invariant
  (`lx.used == 0 and lx.next_ptr == 0` after every pop), so
  divergence is caught at iter 0 rather than overflow at iter 64.
- `test_next_ptr_restored_on_pop_single_level`
- `test_next_ptr_restored_on_pop_with_outer_allocation` — outer
  allocations preserved, only inner ones reclaimed.
- `test_nested_scopes_rewind_lifo` — three-level nesting.
- `test_legitimate_overflow_still_raises` — fix must not hide genuine
  LX exhaustion.
- `test_clear_values_resets_watermark_stack`.

### DIFF
--- a/ktir_cpu/grid.py
+++ b/ktir_cpu/grid.py
@@ -76,7 +76,7 @@ class CoreContext:
         self.hbm = hbm  # Shared HBM
         self._scope_stack: List[Dict[str, Any]] = [{}]  # bottom = function body
         self._lx_bytes: Dict[str, int] = {}  # SSA name -> LX bytes (single source of truth)
-        self._lx_next_ptr_stack: List[int] = []  # watermarks saved on push_scope, restored on pop_scope (#26)
+        self._lx_next_ptr_stack: List[int] = []  # watermarks for lx.next_ptr rewind on pop_scope
 
     def get_grid_id(self, dim: int) -> int:
         """Get grid ID for dimension.
@@ -91,16 +91,38 @@ class CoreContext:
 
     # ------------------------------------------------------------------
     # Region scoping
+    # LX allocation model: scope-watermark bump allocator
+    # LX address space is managed as a bump allocator with scope-level watermarks.
+    # push_scope snapshots lx.next_ptr; pop_scope rewinds it. This is the
+    # most aggressive deallocation strategy the IR permits without liveness analysis:
+
+    # 1). SSA values are immutable bindings with scope-lifetime semantics. A value
+    # that is "in scope" is reachable — there is no point in the IR where a value
+    # is dead but still in scope. So scope-exit is the earliest safe deallocation
+    # point.
+
+    # 2). Intra-scope reuse is not possible because all tiles produced within a scope
+    # coexist simultaneously in LX until the scope pops. Peak LX usage within a
+    # scope equals the sum of all tiles produced in that scope, even if some are
+    # logically "consumed" by a downstream op and never read again.
+
+    # 3). A free-list or explicit dealloc op would require liveness analysis or
+    # last-use annotations, neither of which exist in KTIR's current model
+    # (compute uses standard MLIR dialects; SCF regions define lifetime).
+
+    # The watermark stack stays in lockstep with _scope_stack — the invariant
+    # len(_lx_next_ptr_stack) == len(_scope_stack) - 1 holds by construction.
     # ------------------------------------------------------------------
 
     def push_scope(self):
         """Enter a new region scope (scf.for body, scf.if branch).
 
         Snapshots ``lx.next_ptr`` so ``pop_scope`` can rewind it and
-        keep the bump-pointer in lockstep with ``lx.used`` (see #26).
+        keep the bump-pointer in lockstep with ``lx.used``.
         """
         self._lx_next_ptr_stack.append(self.lx.next_ptr)
         self._scope_stack.append({})
+        assert len(self._lx_next_ptr_stack) == len(self._scope_stack) - 1
 
     def pop_scope(self):
         """Exit the current region scope.
@@ -116,6 +138,7 @@ class CoreContext:
         for name in scope:
             self.untrack_lx(name)
         self.lx.next_ptr = self._lx_next_ptr_stack.pop()
+        assert len(self._lx_next_ptr_stack) == len(self._scope_stack) - 1
 
     # ------------------------------------------------------------------
     # SSA value access
--- a/ktir_cpu/grid.py
+++ b/ktir_cpu/grid.py
@@ -76,6 +76,7 @@ class CoreContext:
         self.hbm = hbm  # Shared HBM
         self._scope_stack: List[Dict[str, Any]] = [{}]  # bottom = function body
         self._lx_bytes: Dict[str, int] = {}  # SSA name -> LX bytes (single source of truth)
+        self._lx_next_ptr_stack: List[int] = []  # watermarks saved on push_scope, restored on pop_scope (#26)
 
     def get_grid_id(self, dim: int) -> int:
         """Get grid ID for dimension.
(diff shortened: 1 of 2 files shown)


nwang-ibm:
cc @fabianlim @WarningRan

ktir_cpu/grid.py
```diff
@@ -76,6 +76,7 @@ def __init__(self, core_id: int, grid_pos: Tuple[int, int, int], lx: LXScratchpa
         self.hbm = hbm  # Shared HBM
         self._scope_stack: List[Dict[str, Any]] = [{}]  # bottom = function body
         self._lx_bytes: Dict[str, int] = {}  # SSA name -> LX bytes (single source of truth)
+        self._lx_next_ptr_stack: List[int] = []  # watermarks saved on push_scope, restored on pop_scope (#26)
```
fabianlim: nit: consider dropping the `(#26)` suffix here — the docstrings on `push_scope`/`pop_scope` already explain the *why*, and issue numbers in inline comments tend to rot as the code evolves. Something like `# watermarks for lx.next_ptr rewind on pop_scope` is self-contained.

fabianlim:
Nice fix — clean and minimal. One suggestion: it'd be worth documenting the LX allocation model and why the watermark approach is the right fit. Something like this in the `CoreContext` class docstring (or as a section in the region-scoping comment block):

---

**LX allocation model: scope-watermark bump allocator**

LX address space is managed as a bump allocator with scope-level watermarks.
`push_scope` snapshots `lx.next_ptr`; `pop_scope` rewinds it. This is the
most aggressive deallocation strategy the IR permits without liveness analysis:

- SSA values are immutable bindings with scope-lifetime semantics. A value
  that is "in scope" is reachable — there is no point in the IR where a value
  is dead but still in scope. So scope-exit is the earliest safe deallocation
  point.

- Intra-scope reuse is not possible because all tiles produced within a scope
  coexist simultaneously in LX until the scope pops. Peak LX usage within a
  scope equals the sum of *all* tiles produced in that scope, even if some are
  logically "consumed" by a downstream op and never read again.

- A free-list or explicit `dealloc` op would require liveness analysis or
  last-use annotations, neither of which exist in KTIR's current model
  (compute uses standard MLIR dialects; SCF regions define lifetime).

The watermark stack stays in lockstep with `_scope_stack` — the invariant
`len(_lx_next_ptr_stack) == len(_scope_stack) - 1` holds by construction.

---

This helps future readers understand *why* the watermark model (vs. e.g. a free-list) and what the peak-usage implication is for the compiler/user. Could live in the class docstring or as a comment block above `push_scope`.

cc @WarningRan

ktir_cpu/grid.py
```diff
@@ -93,20 +94,28 @@ def get_grid_id(self, dim: int) -> int:
     # ------------------------------------------------------------------
 
     def push_scope(self):
-        """Enter a new region scope (scf.for body, scf.if branch)."""
+        """Enter a new region scope (scf.for body, scf.if branch).
+
+        Snapshots ``lx.next_ptr`` so ``pop_scope`` can rewind it and
+        keep the bump-pointer in lockstep with ``lx.used`` (see #26).
+        """
+        self._lx_next_ptr_stack.append(self.lx.next_ptr)
```
fabianlim, line 123: suggestion: consider adding an invariant assert here to catch stack misuse early:

```python
self._lx_next_ptr_stack.append(self.lx.next_ptr)
assert len(self._lx_next_ptr_stack) == len(self._scope_stack) - 1
```

Same in `pop_scope` after the pop. The invariant `len(_lx_next_ptr_stack) == len(_scope_stack) - 1` is maintained by construction, but a defensive assert would surface programming errors (e.g. an op handler that pushes without popping) immediately rather than letting them silently corrupt the watermark state.

fabianlim:
One more follow-up suggestion (separate from this PR's scope): `_write_to_lx` currently advances `next_ptr` and writes without checking capacity — it relies on `track_lx` being called afterward. If those two calls ever get separated or reordered, you'd silently write past the LX boundary. Consider adding a guard directly in `_write_to_lx`:

```python
def _write_to_lx(context: CoreContext, data: np.ndarray):
    size = data.nbytes
    if context.lx.next_ptr + size > context.lx.capacity:
        raise MemoryError(...)
    lx_ptr = context.lx.next_ptr
    context.lx.next_ptr += size
    context.lx.next_ptr = (context.lx.next_ptr + 127) & ~127
    context.lx.write(lx_ptr, data)
```

That way both the address-space path and the byte-accounting path independently enforce the capacity limit.

(Side note: the 128-byte alignment in `_write_to_lx` appears to be copy-pasted from the HBM allocator's stick-boundary alignment in the initial commit — LX is core-local SRAM with no documented stick granularity. Harmless but probably worth a separate cleanup.)

nwang-ibm:
Thanks, @fabianlim, I made the changes accordingly. Please check.

fabianlim:
Thanks for picking up both suggestions. The doc comment block landed in the right place and the test suite is solid — especially  and  covering the edge cases.

The  capacity guard is a good follow-up but agreed it's out of scope here.

LGTM from my side.

fabianlim (dismissed):
approved - the capacity check should be documented and applied in the future

fabianlim:
@nwang-ibm need to have your commits signed [like this](https://github.com/torch-spyre/ktir-cpu/pull/15#issuecomment-4356774947) before we can merge

nwang-ibm:
Fix my gpg signature problem. Hope it works now.

fabianlim (approved):
re-approving after previous approval dismissed by rebase

PR #34, opened 2026-05-07
[Fix] Fix dynamic shapes: `memref<?>`, SSA sizes as operands, symbolic affine dims

### LINKED ISSUE #30: [Bug] Dynamic shapes not supported: symbolic affine dims, memref<?>, SSA sizes
## Summary

The interpreter does not support kernels with runtime-determined input sizes. Three concrete failures prevent execution of kernels like:

\`\`\`mlir
#set = affine_set<(d0)[s0] : (d0 >= 0, -d0 + s0 - 1 >= 0)>
#map = affine_map<(d0) -> (d0)>
#set1 = affine_set<(d0) : (d0 >= 0, -d0 + 1023 >= 0)>

%x_mem = ktdp.construct_memory_view %arg0, sizes: [%n_idx], strides: [1]
            {coordinate_set = #set, memory_space =
            #ktdp.spyre_memory_space<HBM>} : memref<?xf32>

%x_tile = ktdp.construct_access_tile %x_mem[%off_idx]
            {access_tile_order = #map, access_tile_set = #set1} : memref<?xf32>
            -> !ktdp.access_tile<1024xindex>
\`\`\`

This is the interpreter-side counterpart to torch-spyre/ktir-mlir-frontend#16.

---

## Bug 1 — Symbolic affine set dimensions (`[s0]`) crash the parser

`affine_set<(d0)[s0] : (d0 >= 0, -d0 + s0 - 1 >= 0)>` uses `s0` as a symbolic dimension (runtime-known size). The parser in `ktir_cpu/parser_ast.py` only handles dimension variables (`d0`, `d1`, …). When `_atom()` encounters `s0` it raises:

```
ValueError: Unexpected token: 's0'
```

The `[s0]` symbol list between dim list and `:` is also not parsed at all — `parse_affine_set` splits on `:` without accounting for it.

**Affected code:** `parser_ast.py:233`, `parser_ast.py:331-360`, `affine.py:AffineSet`

---

## Bug 2 — `memref<?xf32>` dynamic dimension crashes type parsing

In `parse_construct_memory_view` (`ktdp_ops.py:197`):

```python
memref_dims = [int(p) for p in parts[:-1]]
```

`int("?")` raises `ValueError`. The `?` token is never handled.

**Affected code:** `ktdp_ops.py:197`

---

## Bug 3 — SSA sizes (`sizes: [%n_idx]`) never registered as operands

When a size token is not an integer literal, the parser stores `None` (`ktdp_ops.py:160`) but does **not** add the SSA name to the operand list (unlike SSA strides, which are correctly appended to `ssa_stride_operands`). At execution time, `context.get_value("%n_idx")` is never called and the runtime shape is unavailable.

**Affected code:** `ktdp_ops.py:150-160`, `ktdp_ops.py:221-227`

---

## Note — `construct_access_tile` is already fine

`parse_construct_access_tile` reads shape exclusively from the `!ktdp.access_tile<...>` result type and never inspects the memref type token, so `memref<?xf32>` passes through without issue.

---

## Failing tests

Added in `tests/test_dialects_parse.py` (`TestKtdpParsers`):

```python
@pytest.mark.xfail(reason="symbolic affine dims [s0] not yet supported (#30)", strict=True)
def test_affine_set_with_symbolic_dim(self):
    # affine_set<(d0)[s0] : (d0 >= 0, -d0 + s0 - 1 >= 0)> uses a symbolic
    # dimension s0 whose value is only known at runtime (the tensor size).
    # The current parser raises ValueError on the unknown token 's0'.
    op = self._parse(
        "%view = ktdp.construct_memory_view %ptr, sizes: [%n_idx], strides: [1]"
        " { coordinate_set = affine_set<(d0)[s0] : (d0 >= 0, -d0 + s0 - 1 >= 0)>,"
        " memory_space = #ktdp.spyre_memory_space<HBM> } : memref<?xf32>",
        args={"%ptr": "index", "%n_idx": "index"},
    )
    self.assert_op_type(op, "ktdp.construct_memory_view")
    self.assert_attribute(op, "dtype", "f32")
    self.assert_attribute(op, "memory_space", "HBM")
    # shape should be dynamic (None or -1 convention TBD)
    # operands must include both %ptr and %n_idx
    self.assert_num_operands(op, 2)

@pytest.mark.xfail(reason="memref<?xf32> dynamic dimension not yet supported (#30)", strict=True)
def test_construct_memory_view_dynamic_memref_type(self):
    # memref<?xf32> uses '?' for a dimension whose size is an SSA value.
    # The current parser calls int(p) on each dimension token and raises
    # ValueError when it encounters '?'.
    op = self._parse(
        "%view = ktdp.construct_memory_view %ptr, sizes: [%n_idx], strides: [1]"
        " { memory_space = #ktdp.spyre_memory_space<HBM> } : memref<?xf32>",
        args={"%ptr": "index", "%n_idx": "index"},
    )
    self.assert_op_type(op, "ktdp.construct_memory_view")
    self.assert_attribute(op, "dtype", "f32")

@pytest.mark.xfail(reason="SSA sizes not registered as operands (#30)", strict=True)
def test_construct_memory_view_ssa_size_as_operand(self):
    # When sizes: [%n_idx] appears, %n_idx must be added to the op's
    # operand list (like SSA strides already are) so the executor can
    # call context.get_value("%n_idx") to resolve the runtime size.
    # Currently sizes stored as None; %n_idx is never in op.operands.
    op = self._parse(
        "%view = ktdp.construct_memory_view %ptr, sizes: [%n_idx], strides: [1]"
        " { memory_space = #ktdp.spyre_memory_space<HBM> } : memref<?xf32>",
        args={"%ptr": "index", "%n_idx": "index"},
    )
    # %ptr + %n_idx = 2 operands
    self.assert_num_operands(op, 2)
    self.assert_operand_names(op, "%ptr", "%n_idx")

def test_construct_access_tile_dynamic_memref(self):
    # construct_access_tile already handles memref<?xf32> correctly:
    # it reads shape only from the !ktdp.access_tile<...> result type,
    # never from the memref type, so '?' passes through without issue.
    op = self._parse(
        "%x_tile = ktdp.construct_access_tile %x_mem[%off_idx]"
        " { access_tile_order = affine_map<(d0) -> (d0)>,"
        "   access_tile_set = affine_set<(d0) : (d0 >= 0, -d0 + 1023 >= 0)> }"
        " : memref<?xf32> -> !ktdp.access_tile<1024xindex>",
        args={"%x_mem": "memref<?xf32>", "%off_idx": "index"},
    )
    self.assert_op_type(op, "ktdp.construct_access_tile")
    self.assert_attribute(op, "shape", (1024,))
```

---

## Fix sketch

1. **`parse_affine_set`**: after `parse_dim_list()`, check for a `[s0, s1, ...]` block and parse it into a symbol list. Add `n_syms` and `symbols` to `AffineSet`. Add `("sym", i)` node type to `_eval_node`. Thread symbol values through `contains` / `enumerate`.

2. **`memref<?>`**: treat `?` as a sentinel (e.g. `-1` or `None`) when parsing memref dimensions. Store dynamic dims in `TileRef.shape` using the same sentinel and resolve them from the SSA operand at execution time.

3. **SSA sizes**: mirror the existing SSA-stride pattern — append SSA size names to an `ssa_size_operands` list and include them in `Operation.operands`. In the executor (`ktdp__construct_memory_view`), resolve them via `context.get_value()`.

---

## Related

- torch-spyre/ktir-mlir-frontend#16 — same gap in the MLIR frontend parser/verifier

fabianlim:
Fixes #30.

cc: @nwang-ibm @kcirred @lchu6 @kiszk 

Three parser/executor bugs prevented kernels with runtime-determined input sizes from running. All three are fixed here; the previously-`xfail` tests now pass on both the regex parser and the MLIR frontend adapter.

### Bug 1 — `memref<?xf32>` crashes on `?` (`ktdp_ops.py`)

`int(p)` raised `ValueError` on the `?` token. Changed to `None if p == "?" else int(p)`. `None` and SSA-string entries round-trip as `?` in the result type.

### Bug 2 — SSA sizes never registered as operands (`ktdp_ops.py`)

`sizes: [%n_idx]` stored the SSA name as `None` and never added it to `op.operands`, so `context.get_value("%n_idx")` would fail at execution time. Now stored as a string and collected in `ssa_size_operands` (mirroring the existing `ssa_stride_operands` pattern). The executor resolves dynamic shape entries via `context.get_value()` at runtime.

### Bug 3 — Symbolic affine dims `[s0]` crash the parser (`parser_ast.py`, `affine.py`)

`_TOKEN_RE` did not include `[` or `]`, so `(d0)[s0]` tokenised to `['(', 'd0', ')', 's0']` — the `[s0]` part was silently dropped and `s0` was an unknown identifier. Changes:

- Add `[]` to `_TOKEN_RE`.
- `_Parser` gains `sym_index` dict and `parse_sym_list()`.
- `_atom` recognises `("sym", i)` AST nodes; fallback accepts canonical `sN` names.
- `_eval_node` / `affine_set_contains` / `enumerate_affine_set` accept an optional `symbols` sequence.
- `AffineSet` gains `n_syms: int = 0` and `symbols=` on `contains()` / `enumerate()`.
- `parse_affine_set` parses the optional `[s0, ...]` list and sets `n_syms`.

### Test changes

- The three `xfail` markers removed; op text updated to include `coordinate_set` (required by the MLIR C++ verifier when the memref is dynamic — without it the verifier rejects the op).
- Adapter tests in `test_parse_adapt.py` now inherit the three tests unchanged (no skips).
- New unit tests in `test_ast.py` and `test_affine.py` cover symbol list parsing, `n_syms`, and symbol-parameterised `contains` / `enumerate`.

### Test plan

- [ ] `uv run pytest tests/ -v` — 538 passed, 6 xfailed, 0 failures

### DIFF
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -205,26 +205,37 @@ def parse_construct_memory_view(op_text, parse_ctx: ParseContext):
     # '?' in the memref type means the dimension is dynamic (value only known at
     # runtime).  We keep it as None until we can substitute the SSA size below.
     memref_dims = [None if p == "?" else int(p) for p in parts[:-1]]
-    # Build shape: prefer SSA size tokens for dynamic dims, fall back to memref_dims.
-    # An SSA name (e.g. "%n_idx") is kept as a string in the shape tuple; the
-    # executor resolves it via context.get_value() at runtime.  The result_type
-    # round-trips the string back to '?' so downstream MLIR consumers see a valid
-    # dynamic memref type.
     if sizes is not None:
+        if len(sizes) != len(memref_dims):
+            raise ValueError(
+                f"construct_memory_view: sizes count {len(sizes)} does not match "
+                f"memref dimension count {len(memref_dims)}"
+            )
         resolved = []
-        for i, s in enumerate(sizes):
-            mem_d = memref_dims[i] if i < len(memref_dims) else None
-            if isinstance(s, str):
-                resolved.append(s)  # SSA name — resolved at runtime
-            elif s is not None and mem_d is not None and s != mem_d:
-                raise ValueError(
-                    f"construct_memory_view: sizes[{i}]={s} does not match "
-                    f"memref dimension {mem_d}"
-                )
+        for i, (s, mem_d) in enumerate(zip(sizes, memref_dims)):
+            if mem_d is not None:
+                # concrete dim — s must equal it
+                if not isinstance(s, str) and s != mem_d:
+                    raise ValueError(
+                        f"construct_memory_view: sizes[{i}]={s} does not match "
+                        f"memref dimension {mem_d}"
+                    )
+                resolved.append(s)
             else:
-                resolved.append(s if s is not None else mem_d)
+                # dynamic dim ('?') — s must be an SSA name
+                if not isinstance(s, str):
+                    raise ValueError(
+                        f"construct_memory_view: sizes[{i}]={s} given for a '?' dim; "
+                        f"dynamic dim requires an SSA name, not a literal"
+                    )
+                resolved.append(s)
         shape = tuple(resolved)
     else:
+        if any(d is None for d in memref_dims):
+            raise ValueError(
+                "construct_memory_view: memref has dynamic '?' dim(s) but no sizes: "
+                "attribute was provided; dynamic dims require SSA sizes"
+            )
         shape = tuple(memref_dims)
 
     attrs = parse_attr_block(op_text, parse_ctx.aliases)
--- a/ktir_cpu/parser_ast.py
+++ b/ktir_cpu/parser_ast.py
@@ -22,31 +22,34 @@ This module owns all AST-level concerns for affine maps and sets:
 
 Public API
 ----------
-parse_affine_map(s)          -> AffineMap
-parse_affine_set(s)          -> AffineSet
-eval_affine_map(amap, dims)  -> tuple[int, ...]
-affine_set_contains(aset, point) -> bool
-enumerate_affine_set(aset, shape) -> list[tuple[int, ...]]
+parse_affine_map(s)                        -> AffineMap
+parse_affine_set(s)                        -> AffineSet
+eval_affine_map(amap, dims)                -> tuple[int, ...]
+affine_set_contains(aset, point, symbols)  -> bool
+enumerate_affine_set(aset, shape, symbols) -> list[tuple[int, ...]]
 
-The first two are called by ``parser.py`` (MLIR structure parser) when it
-encounters an ``affine_map<...>`` or ``affine_set<...>`` string.
-The remaining three are called by the convenience methods on ``AffineMap``
-and ``AffineSet`` in ``affine.py``, and directly by ``memory_ops.py``.
+``parse_affine_map`` / ``parse_affine_set`` are called by ``parser.py``
+when it encounters an ``affine_map<...>`` or ``affine_set<...>`` string.
+The evaluation functions are called by the convenience methods on
+``AffineMap`` and ``AffineSet`` in ``affine.py``, and directly by
+``memory_ops.py``.
 
 Supported scope
 ---------------
-Linear affine expressions only:
-  - Dimension variables:  d0, d1, ...
+Linear affine expressions:
+  - Dimension variables: d0, d1, ... (or arbitrary names via the dim list)
+  - Symbol variables:    s0, s1, ... (or arbitrary names via the symbol list)
   - Integer constants
   - Addition (+), subtraction (-), negation (unary -)
-  - Multiplication by a constant coefficient (N * dI)
+  - Multiplication by a constant coefficient (N * dI or dI * N)
 
-No symbolic identifiers (s0, s1, ...), floordiv, ceildiv, or mod.
+Not supported: floordiv, ceildiv, mod.
 
-Assumption for affine_set enumeration
--------------------------------------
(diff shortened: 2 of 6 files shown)


ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -48,7 +48,11 @@ def ktdp__construct_memory_view(op, context, env):
     for attr in ("shape", "strides", "memory_space", "dtype"):
         if attr not in op.attributes:
             raise ValueError(f"construct_memory_view: missing required attribute '{attr}'")
-    shape = tuple(op.attributes["shape"])
+    # SSA size names are stored as strings by the parser; resolve them at runtime.
```
lchu6, line 51: since values are lazy evaluated at runtime, majority of the lifecycle will be `tuple(str | int)` type. 
I am thinking if there is any way to make this type hint better so consumers of this value will remember this type change.  right now I don't see any issue, but new consumers of this "shape" need to keep this in mind.

this comment does not require code change, but pointing out a small potential risk if any existing/new consumer of "shape" wrongly thought thats an `tuple(int)` type.

lchu6 (dismissed):
LG.

only one tiny concern which does not require action now.

ktir_cpu/dialects/ktdp_ops.py
fabianlim: Opened #40 to track adding an optional attributes schema to `@register` for runtime validation and documentation.

fabianlim:
@lchu6 I added more comments and cleaned it up a little bit more just some minor changes in the last 2 commits. If you dont mind give me another approval, then go ahead to merge it

PR #35, opened 2026-05-07
[Fix] setup_mlir: separate --hash and --wheel into two distinct options

fabianlim:
## Summary

Follow-up to #24, which switched to downloading the pinned LLVM artifact from `torch-spyre/ktir-mlir-frontend`. That PR inadvertently kept `--wheel --hash` in both the README and CI — but after [ktir-mlir-frontend#14](https://github.com/torch-spyre/ktir-mlir-frontend/pull/14), `--wheel` takes precedence and silently ignores `--hash`, so we were installing the bleeding-edge `mlir_wheel` instead of the pinned artifact.

This PR fixes that:
- CI now calls `--hash "$LLVM_HASH" --repo torch-spyre/ktir-mlir-frontend` (no `--wheel`), using `GITHUB_TOKEN` which has cross-repo `actions:read` within the org — no PAT required.
- README is updated to present two explicit options: Option 1 (recommended, pinned artifact via `--hash`) and Option 2 (dev/testing only, `--wheel` fallback).

## Test plan

- [x] CI passes with `GITHUB_TOKEN` downloading the pinned artifact from `torch-spyre/ktir-mlir-frontend`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,8 @@ jobs:
         run: sudo apt-get install -y ninja-build
 
       # Download setup_mlir.py from upstream ktir-mlir-frontend and resolve MLIR_DIR.
-      # We pass --wheel to use the mlir_wheel fallback (artifact path requires a GitHub
-      # token not available in public CI). --hash pins to the LLVM build tested upstream.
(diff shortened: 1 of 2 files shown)


fabianlim:
cc @lchu6 @kiszk

lchu6 (approved):

PR #42, opened 2026-05-12
[Fix] Fix dynamic shapes: `memref<?>`, SSA sizes as operands, symbolic affine dims

fabianlim:
Redoing #30 because the original PR had some problems. @lchu6 if you can approve this I will go ahead to merge. Its basically the same thing as #30

### DIFF
--- a/examples/triton-ktir/vector_add_dynamic_ktir.mlir
+++ b/examples/triton-ktir/vector_add_dynamic_ktir.mlir
@@ -1,0 +1,42 @@
+#map = affine_map<(d0) -> (d0)>
+#set = affine_set<(d0)[s0] : (d0 >= 0, -d0 + s0 - 1 >= 0)>
+#set1 = affine_set<(d0) : (d0 >= 0, -d0 + 1023 >= 0)>
+module {
+  func.func @add_kernel_dynamic(
+      %x_ptr: index,
+      %y_ptr: index,
+      %output_ptr: index,
+      %n_elements: i32
+  ) attributes {grid = [1]} {
+    %c0 = arith.constant 0 : index
+    %n = arith.index_cast %n_elements : i32 to index
+
+    %x_view = ktdp.construct_memory_view %x_ptr, sizes: [%n], strides: [1] {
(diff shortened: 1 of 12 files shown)


lchu6:
Symbolic affine sets are parsed, but we didn't test downstream consumption parsing.  e.g. if you add the downstream `.load`/`.store`, does parsing go well all the way there?

fabianlim:
Good catch! I have added a simple dynamic-shape vector add example (`vector_add_dynamic_ktir.mlir`) that exercises symbolic affine coordinate sets all the way through `ktdp.load` / `ktdp.store`, with `n_elements` varied across [256, 512, 1024] to cover the masking/padding path. I have raised #47 to track adding richer examples (2-D, multi-core, more dtypes) as they become available.

lchu6:
I honestly wasn't sure if this is an issue, but currently:

the access tile reading does not respect %n (i.e. s0). 

e.g. if n_elements=512, your access tile would still read 1024 elements and 512 of them are out of bounds.  

this is definitely an issue, but in the meantime, I can also argue:  with a valid KTIR, this should be handled already and such case won't happen.

e.g. with n_elements=1025 and tile_size=1024,  the correct ktir should have second tile size as [1] rather than [1024].
similarly, with n_elements=512, tile_size won't be 1024 as inductor won't emit this kind of value combinations.

I will let you decide if this is an issue.  but from parser perspective values of %n (i.e. s0) is not being used or considered in your impl.

lchu6:
actually, I take back what I said above.   I think we discussed this at certain point. for out of bound, it still reads full 1024 but with masked value.    this will be identical behavior as tensor_descriptors.  so your reading 1024 elements should still be valid, but with a mask.

fabianlim:
Yes, that's correct — and you got to the right conclusion yourself! This was an intentional design assumption: the access tile is fixed-size and out-of-bounds positions are masked rather than being an error, matching `tensor_descriptor` semantics. When `n_elements < tile_size`, the load returns masked values for the out-of-range positions and the store ignores them.

I've added a comment to `vector_add_dynamic_ktir.mlir` to make this intention explicit for future readers.

lchu6 (approved):

PR #43, opened 2026-05-12
Implement ktdp.construct_distributed_memory_view (fixes #1)

### LINKED ISSUE #1: [RFC conformance] Implement ktdp.construct_distributed_memory_view

nwang-ibm:
# Distributed memory views + per-core LX routing

Closes #1.

## Summary

Lands `ktdp.construct_distributed_memory_view` end-to-end:

- **New types** — `DistributedMemRef` and `DistributedTileRef`, mirroring
  upstream's `MemRef` / `TileRef` split for distributed views.
- **`BoxSet`** — axis-aligned specialisation of `AffineSet`; parse-time
  lowering puts partition resolution and load/store on an O(ndim) fast
  path. ~58,000× speedup of `distributed_tile_access` on a 256×512
  fixture; ~315× end-to-end including `distributed_load`.
- **Per-core LX routing** — `#ktdp.spyre_memory_space<LX, core = N>`
  reads and writes route to core N's scratchpad via a new `RingBackend`
  transport seam (Phase 1: `DirectLXBackend`, no ring latency).

5 commits; **714 passed, 1 skipped, 7 xfailed**. The 2 new xfails are
BSP multi-round tests that un-xfail once the generator-based scheduler
lands ([#50](https://github.com/torch-spyre/ktir-cpu/issues/50)).

## Type model: distributed analogue of upstream's MemRef/TileRef split

Single-allocation memory views are already separated:

```
construct_memory_view  →  MemRef  (owns memory_space, base_ptr, coordinate_set)
                              ↓
construct_access_tile  →  TileRef (byte-addressed sub-view; .memref points back)
```

Distributed views mirror that:

```
single allocation:                          distributed:

  MemRef                                      DistributedMemRef
   ├ memory_space                              └ partitions: List[MemRef]
   ├ base_ptr                                    (each carries its own
   ├ coordinate_set      ←──────────────────      coordinate_set = B_i)
   └ lx_core_id (NEW)
       ↓ tile_access                          ↓ distributed_tile_access
  TileRef                                     DistributedTileRef
   ├ memref → MemRef                           ├ partitions: List[TileRef]
   ├ base_ptr (byte addr)                      │   (each survivor TileRef
   ├ strides                                   │    carries C_i and p_i,
   └ ...                                       │    .memref → P_i)
                                               ├ shape (global)
                                               └ global_base = x
```

**Key invariant.** Each surviving `TileRef`'s `.memref` points to its
**partition's** `MemRef` (`P_i`), never to the wrapper that holds it. So
at load/store time `_MemAccessor` reads `tile_ref.memref.memory_space`
and `tile_ref.memref.lx_core_id` from a single concrete MemRef — no
distributed-aware code at the call site.

## Notation

```
  x      = global_base = base_map.eval(indices)         (access tile origin in global coords)
  A      = access_tile_set, in local coords [0, access_shape)
  x + A  = global footprint of the access tile
  B_i    = partition i's coordinate_set (in global coords; = C_i source)
  C_i    = (x + A) ∩ B_i                                (per-survivor coord set)
  p_i    = min(B_i)                                     (partition origin in global coords)

  load coords (partition-local) = C_i - p_i
  output coords (access-local)  = C_i - x
```

`distributed_tile_access` resolves all of this once and stores `C_i` and
`p_i` on each survivor; load/store consume them directly.

### Partition resolution at access-tile construction, not at load

Resolving "which partition owns this global coord, and how does the
local coord map back" used to happen on every load. Now
`MemoryOps.distributed_tile_access` does it once: enumerates `B_i`,
filters to `C_i = B_i ∩ (x + A)`, computes `p_i`, and stores both on
each surviving `TileRef`. Load/store walk `C_i` and `p_i` directly.

### `BoxSet` for axis-aligned sets

Every coordinate set we actually construct in distributed views — `B_i`,
`A`, and `C_i` — is an axis-aligned integer box. `AffineSet.enumerate`
on those is brute-force O(∏shape · n_constraints) per call.

`BoxSet` is the **axis-aligned specialisation of `AffineSet`** — peer
dataclasses under a `Union`, NOT a class hierarchy (a prior subclass
attempt failed because polymorphism over a slow parent hid the slow
path). `parse_affine_set` returns `BoxSet` if the set is axis-aligned,
fully-pinned, unit-coefficient, and non-symbolic; otherwise `AffineSet`.

```
distributed_tile_access:
  if B_i is BoxSet and (x + A) is BoxSet:
      C_i = B_i.intersect(BoxSet.translate(x + A))   # O(ndim), no enumeration
  else:
      enumerate B_i, filter by membership in x + A   # AffineSet slow path preserved

distributed_load / distributed_store:
  if survivor.coordinate_set is BoxSet:
      sub_tile_ref = _subtile_ref(survivor, box=C_i)  # inherits parent strides
      out[lo - x : hi - x] = MemoryOps.load(ctx, sub_tile_ref).data  # one slice paste
  else:
      per-coord scatter (slow path preserved)
```

| Stage (256×512 view, 4 partitions, 32×128 access tile) | Fast (BoxSet) | Slow (AffineSet) | Speedup |
|---|---|---|---|
| `distributed_tile_access` | ~8 µs | ~455 ms | ~58,000× |
| Full (access + `distributed_load`) | ~1.5 ms | ~470 ms | ~315× |

Sub-TileRef construction inherits the partition's strides verbatim, so
row-major and column-packed partitions both work uniformly with no
caller-side transposes — verified by
`test_distributed_store_col_packed_does_not_trample_outside_C_i`.

### Per-core LX routing via `RingBackend`

`#ktdp.spyre_memory_space<LX, core = N>` now actually routes to core
N's scratchpad. `RingBackend` is the transport seam between the memory
op (which says "this byte address lives in core N's LX") and how the
data physically arrives:

```
  MLIR:  #ktdp.spyre_memory_space<LX, core = N>
                      │
                      ▼
  MemRef.lx_core_id = N
                      │
                      ▼
  _MemAccessor(ctx, "LX", byte_addr, lx_core_id=N)
                      │
                      ▼
  ctx.get_lx(N) ──────┬─── N is None or N == self.core_id ──► ctx.lx (local)
                      │
                      └─── N != self.core_id ──► ring_backend.get_lx(N)
                                                          │
                                                          ▼
                                  ┌──────────────────────────────────────┐
                                  │  RingBackend (ABC)                   │
                                  │    get_lx(N), send, recv             │
                                  └──────────────┬───────────────────────┘
                                                 │
                                  ┌──────────────┴──────────────┐
                                  ▼                             ▼
                       ┌──────────────────────┐     ┌──────────────────────┐
                       │ DirectLXBackend      │     │ RingTransferBackend  │
                       │  (Phase 1, this PR)  │     │  (future)            │
                       │                      │     │                      │
                       │ get_lx(N):           │     │ get_lx(N):           │
                       │  scratchpad lookup,  │     │  hop-counted send+   │
                       │  no latency          │     │  recv via RingNetwork│
                       └──────────────────────┘     └──────────────────────┘
```

Phase 1 ships `DirectLXBackend` only — direct lookup, valid for
distributed memory views where partitions are pre-seeded by the host.
RFC §C.3's `distributed-view-copy.mlir` now executes correctly with two
LX partitions tagged `core = 0` / `core = 1` (was xfailed before). No
transport latency yet — see [#50](https://github.com/torch-spyre/ktir-cpu/issues/50).

## Tests

| Test | Status |
|---|---|
| `test_distributed_view_copy_rfc` | ✅ passes (was xfail pre-C4) |
| `test_distributed_copy[*]` (15 cases) | ✅ row/col-band × full/partial/sub-tile × HBM/LX × row-major/col-packed |
| `test_distributed_tile_access_fast_path[*]` | ✅ structural BoxSet emission |
| `test_distributed_tile_access_slow_path[*]` | ✅ same C_i via AffineSet brute-force |
| `test_distributed_store_does_not_trample_outside_C_i` | ✅ row-major no-trample |
| `test_distributed_store_col_packed_does_not_trample_outside_C_i` | ✅ col-packed no-trample |
| `test_affine.py` | ✅ +42 BoxSet / try_from_affine_set / parse-time-lowering cases |
| `test_execute_with_communication_*` | ⚠️ xfail (BSP multi-round; pending [#50](https://github.com/torch-spyre/ktir-cpu/issues/50)) |

Full suite: **714 passed, 1 skipped, 7 xfailed**.

## Follow-ups

- **[#50](https://github.com/torch-spyre/ktir-cpu/issues/50)** — real
  `RingBackend` with working cross-core reduce. Generator-based
  scheduler so `ktdp.reduce` / `ktdp.transfer` actually exchange tiles
  (today `CommOps.reduce` reaches past `CoreContext` into `RingNetwork`
  inside one core's handler — sibling cores haven't run, so receives
  find empty buffers; K1/K2 in `docs/gap_analysis.md`). Then
  `RingTransferBackend` adds hop-count latency. Closes the two BSP
  multi-round xfails introduced here.

- **[#51](https://github.com/torch-spyre/ktir-cpu/issues/51)** — BoxSet
  × dynamic shapes. PR #42 added symbolic affine sets
  (`AffineSet.n_syms`, `symbols` argument); `BoxSet.try_from_affine_set`
  rejects `n_syms > 0` and lets symbolic sets fall through to the
  AffineSet slow path. Fix: symbolic `BoxSet.lo`/`hi` and a `symbols`
  argument on box ops.

- **[#52](https://github.com/torch-spyre/ktir-cpu/issues/52)** —
  rectangular `BoxSet` fast paths in direct `ktdp.load`/`store` and
  indirect-access. The distributed paths are box-aware end-to-end; the
  single-allocation paths still enumerate to a coord list even when the
  access is a box. Depends on PR #49 (`indirect_store`) so the
  indirect-load and indirect-store fast paths can land symmetrically.

- **[#53](https://github.com/torch-spyre/ktir-cpu/issues/53)** — open
  question on whether `ktdp.construct_indirect_access_tile` should
  compose with `DistributedMemRef`. Today `IndirectAccessTile`'s type
  signatures only accept `MemRef` and the composition isn't checked —
  needs an explicit reject or a design.

### DIFF
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -27,6 +27,8 @@ from ..ir_types import (
     Tile,
 )
 from ..latency import LatencyCategory as LC
+from ..ops.arith_ops import ArithOps
+from ..ops.comm_ops import CommOps
 from ..ops.grid_ops import GridOps
 from ..ops.memory_ops import MemoryOps
 from ..parser_ast import parse_affine_map, parse_affine_set
@@ -66,7 +68,8 @@ def ktdp__construct_memory_view(op, context, env):
     memory_space = op.attributes["memory_space"]
     dtype = op.attributes["dtype"]
     coordinate_set = op.attributes.get("coordinate_set")
-    return MemoryOps.tile_view(context, ptr, shape, strides, memory_space, dtype, coordinate_set)
+    lx_core_id = op.attributes.get("lx_core_id") if memory_space == "LX" else None
+    return MemoryOps.tile_view(context, ptr, shape, strides, memory_space, dtype, coordinate_set, lx_core_id)
 
 
 @register("ktdp.construct_distributed_memory_view")
@@ -245,9 +248,19 @@ def parse_construct_memory_view(op_text, parse_ctx: ParseContext):
         strides = parsed
 
     memory_space = "HBM"
-    mem_match = re.search(r'#ktdp\.spyre_memory_space<(\w+)>', op_text)
+    lx_core_id = None
+    # Accept both `<HBM>`/`<LX>` and the RFC's per-core LX form
+    # `<LX, core = N>`.  On real hardware each compute core has its own
+    # private LX SRAM, so a partition tagged `core = N` lives in core N's
+    # scratchpad — captured into lx_core_id and used at load/store time.
+    mem_match = re.search(
+        r'#ktdp\.spyre_memory_space<\s*(\w+)(?:\s*,\s*core\s*=\s*(\d+))?\s*>',
+        op_text,
+    )
     if mem_match:
         memory_space = mem_match.group(1)
+        if mem_match.group(2) is not None:
+            lx_core_id = int(mem_match.group(2))
 
     # dtype and shape are parsed from the memref result type.
     # Validate sizes against memref dimensions when both are concrete.
@@ -306,6 +319,8 @@ def parse_construct_memory_view(op_text, parse_ctx: ParseContext):
         "memory_space": memory_space,
         "dtype": dtype,
     }
+    if lx_core_id is not None:
+        attributes["lx_core_id"] = lx_core_id
     if coordinate_set is not None:
         attributes["coordinate_set"] = coordinate_set
 
@@ -736,3 +751,23 @@ def parse_construct_indirect_access_tile(op_text, parse_ctx: ParseContext):
     )
 
 
+
+
+# ---------------------------------------------------------------------------
+# Communication ops (moved from scf_ops.py)
+# ---------------------------------------------------------------------------
+
+@register("ktdp.transfer", latency_category=LC.COMM)
+def ktdp__transfer(op, context, env):
+    tile = context.get_value(op.operands[0])
+    dst_cores = context.get_value(op.operands[1])
+    CommOps.transfer(context, tile, dst_cores, env.ring_backend)
+    return None
+
+
+@register("ktdp.reduce", latency_category=LC.COMM)
+def ktdp__reduce(op, context, env):
+    tile = context.get_value(op.operands[0])
+    core_group = context.get_value(op.operands[1])
+    reduce_fn = lambda t1, t2: ArithOps.addf(t1, t2)
+    return CommOps.reduce(context, tile, core_group, reduce_fn, env.ring_backend)
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -17,7 +17,15 @@
 import re
 
 from .ktdp_helpers import parse_subscript_expr
-from ..ir_types import AccessTile, IndirectAccessTile, Operation, Tile
+from ..ir_types import (
+    AccessTile,
+    DistributedMemRef,
+    DistributedTileRef,
+    IndirectAccessTile,
+    MemRef,
+    Operation,
+    Tile,
+)
 from ..latency import LatencyCategory as LC
 from ..ops.grid_ops import GridOps
 from ..ops.memory_ops import MemoryOps
@@ -61,6 +69,33 @@ def ktdp__construct_memory_view(op, context, env):
     return MemoryOps.tile_view(context, ptr, shape, strides, memory_space, dtype, coordinate_set)
 
 
+@register("ktdp.construct_distributed_memory_view")
+def ktdp__construct_distributed_memory_view(op, context, env):
+    """Compose N per-partition memory views into one distributed view.
+
+    Each operand must resolve to a :class:`MemRef` carrying its own
+    ``coordinate_set`` (= B_i in global coords).  The op does not allocate
+    or move data — partition resolution at access time is performed by
+    ``MemoryOps.distributed_tile_access``.
+    """
+    partitions = [context.get_value(name) for name in op.operands]
+    for i, p in enumerate(partitions):
+        if not isinstance(p, MemRef):
+            raise ValueError(
+                f"construct_distributed_memory_view: operand {i} "
+                f"is {type(p).__name__}, expected MemRef"
+            )
+    if "shape" not in op.attributes or "dtype" not in op.attributes:
+        raise ValueError(
+            "construct_distributed_memory_view: missing required attributes 'shape'/'dtype'"
+        )
+    return DistributedMemRef(
+        partitions=partitions,
+        shape=tuple(op.attributes["shape"]),
+        dtype=op.attributes["dtype"],
+    )
+
+
 @register("ktdp.construct_access_tile")
 def ktdp__construct_access_tile(op, context, env):
     parent_ref = context.get_value(op.operands[0])
@@ -72,12 +107,27 @@ def ktdp__construct_access_tile(op, context, env):
     # Pass it to tile_access so the offset is computed via affine evaluation rather than
     # the old rectangular sum(idx * stride) shortcut.
     base_map = op.attributes["base_map"]
+    coordinate_set = op.attributes.get("coordinate_set")
+    if isinstance(parent_ref, DistributedMemRef):
+        # Distributed parent: resolve partition routing now and produce a
+        # DistributedTileRef carrying surviving partitions.  Load/store
+        # consume it directly.
+        dist_tile_ref = MemoryOps.distributed_tile_access(
+            parent_ref, access_shape, base_map, indices, access_tile_set=coordinate_set
+        )
+        return AccessTile(
+            parent_ref=dist_tile_ref,
+            shape=access_shape,
+            base_map=base_map,
+            coordinate_set=coordinate_set,
+            coordinate_order=op.attributes.get("coordinate_order"),
+        )
     tile_ref = MemoryOps.tile_access(context, parent_ref, indices, access_shape, base_map)
     return AccessTile(
         parent_ref=tile_ref,
         shape=access_shape,
         base_map=base_map,
-        coordinate_set=op.attributes.get("coordinate_set"),
+        coordinate_set=coordinate_set,
         coordinate_order=op.attributes.get("coordinate_order"),
     )
 
@@ -88,6 +138,11 @@ def ktdp__load(op, context, env):
     if isinstance(access_tile, IndirectAccessTile):
         result_shape = op.attributes.get("_result_shape", access_tile.shape)
         return MemoryOps.indirect_load(context, access_tile, result_shape=result_shape)
+    if isinstance(access_tile.parent_ref, DistributedTileRef):
+        result_shape = op.attributes.get("_result_shape", access_tile.shape)
+        return MemoryOps.distributed_load(
+            context, access_tile.parent_ref, result_shape=result_shape
+        )
     css = access_tile.coordinate_set    # AffineSet | None
     cso = access_tile.coordinate_order  # AffineMap | None
     if css is not None:
@@ -106,6 +161,9 @@ def ktdp__store(op, context, env):
     access_tile = context.get_value(op.operands[1])
     if isinstance(access_tile, IndirectAccessTile):
         raise NotImplementedError("ktdp.store with IndirectAccessTile is not yet supported")
+    if isinstance(access_tile.parent_ref, DistributedTileRef):
+        MemoryOps.distributed_store(context, value, access_tile.parent_ref)
+        return None
     tile_ref = access_tile.parent_ref
     css = access_tile.coordinate_set
     cso = access_tile.coordinate_order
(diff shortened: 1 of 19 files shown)


nwang-ibm:
@fabianlim @WarningRan, please take a look.

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -179,7 +251,15 @@ def parse_construct_memory_view(op_text, parse_ctx: ParseContext):
...
 
     memory_space = "HBM"
-    mem_match = re.search(r'#ktdp\.spyre_memory_space<(\w+)>', op_text)
+    # Accept both `<HBM>`/`<LX>` and the RFC's per-core LX form
+    # `<LX, core = N>`.  The core index is preserved by the MLIR author for
+    # future per-core LX routing; the simulator currently treats all "LX"
+    # as the executing core's scratchpad, so the core index is dropped
+    # here with a TODO.
+    mem_match = re.search(
+        r'#ktdp\.spyre_memory_space<\s*(\w+)(?:\s*,\s*core\s*=\s*\d+)?\s*>',
```
WarningRan: I noticed the regex accepts `core = N` but the integer only participates in matching, not capture - `(?:\s*,\s*core\s*=\s*\d+)?` is a non-capturing group, so the parsed value doesn't survive into TileRef. I think that's what's driving the `A_LX1_addr: 16384 → 20480` workaround in the fixture — without the core value on the IR side, the two LX partitions end up indistinguishable at runtime.

Two small things that feel connected to me, curious what you think:

Would it make sense to change `(?:\s*,\s*core\s*=\s*\d+)?` to `(?:\s*,\s*core\s*=\s*(\d+))?` and add `core_id: Optional[int] = None` on `TileRef`? Routing could still land separately (I've opened #XXX for that), but at least the value would be preserved in the AST.

And if we do that, it'd probably be worth a small parser-level test that parses `<LX, core = 0>` and `<LX, core = 1>` and checks the `TileRefs` differ - mostly so the field doesn't quietly drift out of use later.

These feel a bit like one change to me rather than two, but happy to split or drop either if you see it differently.

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -65,9 +94,24 @@ def ktdp__construct_access_tile(op, context, env):
...
     # base_map is an AffineMap object (always present; synthesized as identity if absent in MLIR).
-    # Pass it to tile_access so the offset is computed via affine evaluation rather than
-    # the old rectangular sum(idx * stride) shortcut.
     base_map = op.attributes["base_map"]
+    if isinstance(parent_ref, DistributedTileRef):
+        # Distributed parent: non-zero base offsets would need an ownership
+        # rule for the origin cell, which we haven't defined.  v1 supports
+        # zero-offset access over the full global domain — matches the RFC
+        # §C.3 reference example (distributed-view-copy.mlir).
+        if any(i != 0 for i in indices):
```
WarningRan: Trying to make sure I'm reading this path right - the `AccessTile` constructed just below (line 108-114) stores `base_map`, but `_enumerate_access_coords` further down (line 125-140) only reads `coordinate_set` and `coordinate_order`; `base_map` isn't consulted. The guard here rejects non-zero indices, but a non-identity `base_map` with zero indices (say `(d0,d1) -> (d0+5, d1)` + `[%c0, %c0]`) would pass this guard and the enumeration would still start at `(0, 0)` - which is a different result than the non-distributed path would give for the same MLIR.

If the intent is "v1 only supports identity `base_map` + zero indices on distributed parents," would it be worth checking `base_map.is_identity()` alongside the indices check? That way the limitation is enforced rather than implied — curious if you'd considered that and had a reason to leave it.

ktir_cpu/ops/memory_ops.py
```diff
@@ -19,47 +19,37 @@
...
-from ..ir_types import MemRef, Tile, TileRef
+from ..dtypes import bytes_per_elem as _bytes_per_elem, to_np_dtype as _to_np_dtype
+from ..ir_types import Tile, TileRef, DistributedTileRef
 from ..grid import CoreContext
 from ..memory import HBMSimulator
 
 
-class _MemAccessor:
-    """Resolves a (context, memref, byte_addr) triple into simulator read/write calls.
+def partition_origin(coord_set: AffineSet, global_shape: Tuple[int, ...]) -> Tuple[int, ...]:
```
WarningRan: Question about the assumption here - `enumerate(global_shape) + per-dim min` gives the bounding-box corner of the coord_set. For axis-aligned rectangular sets (all the examples and tests so far), the BB corner equals the partition's local origin, so this works. RFC §2 allows `coordinate_set` to be a general `IntegerSetAttr` though, and for something like `d0 + d1 <= N` the BB corner isn't necessarily the partition's storage origin - it depends on how the producer laid out the physical memref.

Would it make sense to either (a) check the coord_set is rectangular and raise `NotImplementedError` otherwise, making the restriction explicit, or (b) at least note in the docstring that this function assumes BB-corner-aligned physical layout? Right now someone passing a non-rectangular set would get silently-wrong byte offsets with no indication the function wasn't designed for that case. Totally fine if that's a v2 concern, just want to flag it.

WarningRan:
One architectural observation as food for thought, not for this PR:

The `core = N` handling here hints at the limits of `memory_space: str` on `TileRef`. If we end up adding `core_id` as a separate field (per my inline comment on ktdp_ops.py:260), at some point it'd probably be cleaner to have a `SpyreMemorySpace` dataclass kind, core_id, and future fields like bank_id) — the RFC's `#ktdp.spyre_memory_space<...>` is a structured attribute on the MLIR side, so matching it on the Python side would feel natural. Cross-cutting refactor, clearly a separate PR — happy to open an issue to track if you think it's worth it.

ktir_cpu/dialects/ktdp_ops.py
fabianlim: Good catch — fixed in [`ecca1b9`](https://github.com/torch-spyre/ktir-cpu/commit/ecca1b9). The regex now captures the core index (`(?:\s*,\s*core\s*=\s*(\d+))?`) and the parser stores it as `lx_core_id` on `MemRef` (validated to only be set when `memory_space == "LX"`). It propagates through `tile_access` via `TileRef.memref` so `_MemAccessor` can route reads/writes to the correct LX scratchpad via `context.get_lx(N)`.

The `A_LX1_addr: 16384 → 20480` workaround is gone — the example file is back to `16384` and the two `<LX, core = 0>` / `<LX, core = 1>` partitions land in distinct scratchpads.

Parser test for round-tripping `core = N` is in [`tests/test_distributed_view.py::test_distributed_view_copy_rfc`](https://github.com/torch-spyre/ktir-cpu/blob/dist_mem_view/tests/test_distributed_view.py) (the RFC §C.3 fixture exercises both `core = 0` and `core = 1`). Happy to add a more focused parser-only test if you'd prefer that as a regression guard.

ktir_cpu/dialects/ktdp_ops.py
fabianlim: You're right that the original guard was incomplete — it only rejected non-zero indices and would have produced wrong results for a non-identity `base_map` with zero indices. The whole approach is now obsolete though: in [`e6e8e27`](https://github.com/torch-spyre/ktir-cpu/commit/e6e8e27) (C2 of this PR), `MemoryOps.distributed_tile_access` actually evaluates `base_map.eval(indices)` to compute `x = global_base` and uses that as the access tile's origin in global coords. The intersection `C_i = (x + A) ∩ B_i` is then computed against the real global footprint, and the survivor TileRefs carry `C_i` and `p_i` directly — load/store consume them without ever re-checking indices or base_map.

Concretely: the example you mentioned (`(d0,d1) -> (d0+5, d1)` + `[%c0, %c0]`) now produces `x = (5, 0)`, and `C_i` is computed against `[5, 5+access_shape[0]) × [0, access_shape[1])`. So the same MLIR produces the same result on distributed and non-distributed paths. The `if any(i != 0 for i in indices)` guard is gone, replaced by the actual evaluation.

The `_enumerate_access_coords` helper you spotted as the bug source was also removed — both the access-tile-construction and load/store paths now go through the unified `distributed_tile_access` → `DistributedTileRef` flow.

ktir_cpu/ops/memory_ops.py
fabianlim: Good catch — added a docstring note in [`4665e29`](https://github.com/torch-spyre/ktir-cpu/commit/4665e29) on both `MemRef.coordinate_set` and `MemoryOps.distributed_tile_access`.

The short version: this isn't actually an assumption inside `distributed_tile_access` — it's a property of the (`strides`, `coordinate_set`) pair on `MemRef`. Per-axis `strides` is a flat vector that can only describe a strided rectangle, so any `coordinate_set` that isn't axis-aligned has to be stored BB-padded inside the partition's `shape` (slots outside the set are unused but addressable). There's no other valid producer-side encoding given the current type system.

Under that contract, `p_i = min(B_i)` per axis = the BB lower corner = the global coord of element `[0, 0, ..., 0]` of the partition's MemRef. So `(global - p_i) · strides · bpe` lands on the right byte for every point in `B_i`, regardless of whether the set is rectangular or not.

For your `d0 + d1 ≤ N` example: `min(B_i) = (0, 0)`, the producer pads to the `(N+1) × (N+1)` rectangle, and triangle-points outside the set are valid memory but unread. The byte-offset math works identically to the rectangular case.

fabianlim:
Agreed on direction. Holding off until a third field appears (e.g. `bank_id`) — until then the two flat fields + one validator on `MemRef` are cheap enough not to warrant a cross-cutting refactor.

fabianlim (approved):
Approved.

Spawned tracking issues for the follow-ups discussed across the threads:

- **#50** — real `RingBackend` with working cross-core reduce (generator-based scheduler + `RingTransferBackend`); closes the two BSP multi-round xfails this PR introduces.
- **#51** — `BoxSet` × dynamic shapes (`n_syms` / `symbols` support); `try_from_affine_set` currently rejects symbolic sets, so dynamic-shape distributed views fall through to the `AffineSet` slow path.
- **#52** — rectangular `BoxSet` fast paths in direct `ktdp.load`/`store` and indirect-access (depends on #49 for symmetric `indirect_store`).
- **#53** — open question on whether `ktdp.construct_indirect_access_tile` should compose with `DistributedMemRef` (today the type signatures only accept `MemRef` and the composition isn't checked).

PR #45, opened 2026-05-13
Parse MLIR-standard bundled multi-result form for get_compute_tile_id

kcirred:
Previous regex parser handles typical situation of `%pid_m, %pid_n = ktdp.get_compute_tile_id : index, index` but failed to  handle valid syntax of `%pid_m:2 = ktdp.get_compute_tile_id : index, index`. This PR aims to fix that.

### DIFF
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -209,7 +209,13 @@ def parse_get_compute_tile_id(op_text, parse_ctx: ParseContext):
     if not m:
         return None
     names = parse_multi_result_lhs(m.group(1))
-    type_count = sum(1 for t in m.group(2).split(",") if t.strip())
+    types_text = m.group(2).strip()
+    if not types_text:
+        raise ValueError("no result types specified")
+    type_list = [t.strip() for t in types_text.split(",")]
+    if any(not t for t in type_list):
+        raise ValueError(f"empty type in list: {types_text!r}")
+    type_count = len(type_list)
     if len(names) != type_count:
         raise ValueError(
             f"ktdp.get_compute_tile_id: {len(names)} result name(s) but "
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -118,23 +118,81 @@ def ktdp__store(op, context, env):
 # ---------------------------------------------------------------------------
 # Parsers
 # ---------------------------------------------------------------------------
-
-@register_parser("ktdp.get_compute_tile_id")
-def parse_get_compute_tile_id(op_text, parse_ctx: ParseContext):
-    multi_match = re.match(
-        r'((?:%\w+,\s*)*%\w+)\s*=\s*ktdp\.get_compute_tile_id\s*:\s*(.*)', op_text
+# A multi-result op has two MLIR surface forms:
+#   Comma   : "%x, %y = op : index, index"     — referenced as %x, %y
+#   Bundled : "%g:2  = op : index, index"      — referenced as %g#0, %g#1
+#
+# Comma form keeps result names verbatim ("%x", "%y").
+# Bundled form synthesizes "%g#0", "%g#1" so downstream operand lookup
+# finds distinct keys.
+_GET_COMPUTE_TILE_ID_RE = re.compile(
+    r"""
+    ^\s*
+    (?:
+        (?P<comma> %\w+ (?: \s*,\s* %\w+ )* )                # %x  or  %x, %y, ...
+      |
+        (?P<base>  %\w+ ) : (?P<result_count> [1-9]\d* )     # %g:N  (N >= 1)
     )
-    if not multi_match:
-        return None
+    \s*=\s*
+    ktdp\.get_compute_tile_id
+    \s*:\s*
+    (?P<types> [^{(]* )                                      # trailing type list
+    \s*$
+    """,
+    re.VERBOSE,
+)
 
-    result_names = [r.strip() for r in multi_match.group(1).split(',')]
-    result = result_names[0] if len(result_names) == 1 else result_names
+
+def _make_compute_tile_id_op(result: str | list[str]) -> Operation:
     return Operation(
         result=result,
         op_type="ktdp.get_compute_tile_id",
         operands=[],
         attributes={},
-        result_type="index"
+        result_type="index",
+    )
+
+
+@register_parser("ktdp.get_compute_tile_id")
+def parse_get_compute_tile_id(op_text, parse_ctx: ParseContext):
+    match = _GET_COMPUTE_TILE_ID_RE.match(op_text)
+    if not match:
+        return None
+
+    # Trailing ": index, index, ..." must agree with the result
+    # name count. A mismatch is malformed will fail at parse time rather than
+    # silently store the wrong number of names.
+    type_count = sum(1 for t in match["types"].split(",") if t.strip())
+
+    if match["comma"] is not None:
+        names = [r.strip() for r in match["comma"].split(",")]
+        if len(names) != type_count:
+            raise ValueError(
+                f"ktdp.get_compute_tile_id: {len(names)} result name(s) but "
+                f"{type_count} result type(s) in: {op_text!r}"
+            )
+        result = names[0] if len(names) == 1 else names
+        return _make_compute_tile_id_op(result)
+
+    # Bundled form — synthesize the use-site reference names.
+    declared_result_count = int(match["result_count"])
+    if declared_result_count != type_count:
+        raise ValueError(
+            f"ktdp.get_compute_tile_id: bundled form declares "
+            f"{declared_result_count} result(s) but {type_count} result type(s) "
+            f"in: {op_text!r}"
+        )
+    base = match["base"]
+    # Honor the codebase-wide invariant: single result ⇔ bare string,
+    # multi-result ⇔ list[≥2]. The "%base#0" suffix is preserved so that
+    # downstream operands written as "%base#0" still resolve via env lookup.
+    # The MLIR-frontend backend collapses bundled N=1 to a single SSA value
+    # at parse time; emitting a string here keeps both backends consistent
+    # and avoids a list[1] shape that no other producer/consumer expects.
+    if declared_result_count == 1:
+        return _make_compute_tile_id_op(f"{base}#0")
+    return _make_compute_tile_id_op(
+        [f"{base}#{i}" for i in range(declared_result_count)]
     )
 
 
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -22,7 +22,7 @@ from ..latency import LatencyCategory as LC
 from ..ops.grid_ops import GridOps
 from ..ops.memory_ops import MemoryOps
 from ..parser_ast import parse_affine_map, parse_affine_set
-from ..parser_utils import _extract_bracket_content, parse_attr_block, split_top_level
+from ..parser_utils import _extract_bracket_content, find_ssa_names, parse_attr_block, parse_multi_result_lhs, split_top_level
 from .registry import ParseContext, register, register_parser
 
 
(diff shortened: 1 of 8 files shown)


ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -119,22 +119,79 @@ def ktdp__store(op, context, env):
...
+# A multi-result op has two equivalent surface forms in MLIR:
+#
+#   Comma   : "%x, %y = op : index, index"
+#   Bundled : "%g:2  = op : index, index"
+#
+# *bundled* def are referenced as %g#0, %g#1
+# *comma* def are referenced as %x, %y.
+# bundled results become ["%g#0", "%g#1"]
+# comma results stay ["%x", "%y"]. This invariant is
+# Regex with two named alternatives helps us parse and fail-fast when
```
fabianlim: The sentence is cut off — "This invariant is" leads directly into the next comment line without completing the thought. Suggest merging/rewriting the two lines:

```python
# After parsing, results are stored as:
#   bundled "%g:2 = op"  → ["%g#0", "%g#1"]  (synthesized #N use-site names)
#   comma   "%x, %y = op" → ["%x", "%y"]
```

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -119,22 +119,79 @@ def ktdp__store(op, context, env):
...
+
+@register_parser("ktdp.get_compute_tile_id")
+def parse_get_compute_tile_id(op_text, parse_ctx: ParseContext):
+    match = _GET_COMPUTE_TILE_ID_RE.match(op_text)
+    if not match:
+        return None
+
+    # Trailing ": index, index, ..." must agree with the result
+    # name count. A mismatch is malformed will fail at parse time rather than
+    # silently store the wrong number of names.
```
fabianlim: Typo: "A mismatch is malformed will fail" — missing word, probably "A mismatch (malformed input) will fail" or just "A mismatch is malformed and will fail".

tests/test_dialects_parse.py
```diff
@@ -454,6 +454,41 @@ def test_get_compute_tile_id_multi(self):
...
 
+    def test_get_compute_tile_id_bundled(self):
+        # One base name with a ":N" suffix declaring N results. Must canonicalize
+        # to names that use reference (%pid#0, %pid#1).
+        op = self._parse("%pid:2 = ktdp.get_compute_tile_id : index, index")
+        self.assert_op_type(op, "ktdp.get_compute_tile_id")
+        assert op.result == ["%pid#0", "%pid#1"]
+
+    def test_get_compute_tile_id_bundled_three_results(self):
+        op = self._parse(
```
fabianlim: No docstring here while the sibling tests have one. Since this specifically guards the N>2 case (non-obvious edge), a single line would help:

```python
def test_get_compute_tile_id_bundled_three_results(self):
    # Bundled form works for N > 2, not just N == 2.
```

tests/test_dialects_parse.py
```diff
@@ -454,6 +454,41 @@ def test_get_compute_tile_id_multi(self):
...
+            "%g:3 = ktdp.get_compute_tile_id : index, index, index"
+        )
+        assert op.result == ["%g#0", "%g#1", "%g#2"]
+
```
fabianlim, line 587: Consider adding a `N=1` bundled test after this one. The regex allows it (`[1-9]\d*`) and single-result bundled form is valid MLIR, but it's currently untested:

```python
def test_get_compute_tile_id_bundled_single_result(self):
    # %x:1 is valid MLIR — single-result bundled form.
    op = self._parse("%x:1 = ktdp.get_compute_tile_id : index")
    assert op.result == ["%x#0"]
```

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -119,22 +119,79 @@ def ktdp__store(op, context, env):
...
+#   Bundled : "%g:2  = op : index, index"
+#
+# *bundled* def are referenced as %g#0, %g#1
+# *comma* def are referenced as %x, %y.
+# bundled results become ["%g#0", "%g#1"]
+# comma results stay ["%x", "%y"]. This invariant is
+# Regex with two named alternatives helps us parse and fail-fast when
+# identifiying cases that aren't either form.
+_GET_COMPUTE_TILE_ID_RE = re.compile(
+    r"""
```
fabianlim: The bundled `%base:N` form and the `#N` use-site reference convention are generic MLIR syntax, not specific to this op. Right now the logic is duplicated here — if another op ever returns multiple results, we'd need to repeat it.

Two options:
1. **Extract a shared helper** now (e.g. `_parse_multi_result_lhs(lhs_text) -> list[str]`) that both this parser and future parsers call.
2. **File an issue** to track the generalisation with a design sketch, and keep this PR scoped.

Option 2 is lower-risk here since `get_compute_tile_id` is currently the only multi-result op. Which do you prefer?

fabianlim: Here is a draft of the helper if you go with option 1. It could live in `parser_utils.py` (or `ktdp_helpers.py`) so any future parser can import it.

```python
def _parse_multi_result_lhs(lhs_text: str) -> list[str]:
    """
    Parse the LHS of a multi-result MLIR assignment.

    Accepts:
      comma form   "%x, %y"  -> ["%x", "%y"]
      bundled form "%g:2"    -> ["%g#0", "%g#1"]

    Raises ValueError on malformed input.
    """
    # Bundled form: %base:N
    m = re.fullmatch(r'(%\w+):([1-9]\d*)', lhs_text.strip())
    if m:
        base, n = m.group(1), int(m.group(2))
        return [f"{base}#{i}" for i in range(n)]

    # Comma form: %x or %x, %y, ...
    parts = [p.strip() for p in lhs_text.split(",")]
    if all(re.fullmatch(r'%\w+', p) for p in parts):
        return parts

    raise ValueError(f"cannot parse multi-result LHS: {lhs_text!r}")
```

Then `parse_get_compute_tile_id` slims down to:

```python
@register_parser("ktdp.get_compute_tile_id")
def parse_get_compute_tile_id(op_text, parse_ctx: ParseContext):
    m = re.match(
        r"^(.*?)\s*=\s*ktdp\.get_compute_tile_id\s*:\s*([^{(]*)\s*$", op_text
    )
    if not m:
        return None

    names = _parse_multi_result_lhs(m.group(1))
    type_count = sum(1 for t in m.group(2).split(",") if t.strip())
    if len(names) != type_count:
        raise ValueError(
            f"ktdp.get_compute_tile_id: {len(names)} result name(s) but "
            f"{type_count} result type(s) in: {op_text!r}"
        )
    result = names[0] if len(names) == 1 else names
    return _make_compute_tile_id_op(result)
```

The `_GET_COMPUTE_TILE_ID_RE` verbose regex goes away entirely, and the LHS parsing logic is reusable by any future multi-result parser.

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -183,23 +118,81 @@ def ktdp__store(op, context, env):
...
-@register_parser("ktdp.get_compute_tile_id")
-def parse_get_compute_tile_id(op_text, parse_ctx: ParseContext):
-    multi_match = re.match(
-        r'((?:%\w+,\s*)*%\w+)\s*=\s*ktdp\.get_compute_tile_id\s*:\s*(.*)', op_text
+# A multi-result op has two MLIR surface forms:
+#   Comma   : "%x, %y = op : index, index"     — referenced as %x, %y
+#   Bundled : "%g:2  = op : index, index"      — referenced as %g#0, %g#1
+#
+# Comma form keeps result names verbatim ("%x", "%y").
+# Bundled form synthesizes "%g#0", "%g#1" so downstream operand lookup
```
fabianlim, line 192: understsand that you did not want to do the refactoring now, but will be good to put a link to this PR, and a note that with refactoring we can simplify this and share with other parsers.

tests/test_dialects_parse.py
```diff
@@ -522,6 +454,85 @@ def test_get_compute_tile_id_multi(self):
...
+        # form (regex emits "%base#0"; MLIR-frontend emits the SSA value
+        # the C++ parser produces, e.g. "%0"), honoring the codebase-
+        # wide convention: single result ⇔ str; multi-result ⇔ list[≥2].
+        op = self._parse("%x:1 = ktdp.get_compute_tile_id : index")
+        self.assert_op_type(op, "ktdp.get_compute_tile_id")
+        assert isinstance(op.result, str)
+        assert op.result.startswith("%")
+        assert len(op.result) > 1
+
+    def test_get_compute_tile_id_bundled_three_results(self):
```
fabianlim: this test can be parameterialized with `test_get_compute_tile_id_bundled`

tests/test_dialects_parse.py
```diff
@@ -522,6 +454,85 @@ def test_get_compute_tile_id_multi(self):
...
+                   for r in op.result)
+        assert len(set(op.result)) == 3
+
+    def test_get_compute_tile_id_bundled_count_mismatch_rejected(self):
+        # Regex parser raises ValueError("result type ..."); MLIR frontend
+        # raises MLIRError("operation defines N results but was provided M
+        # to bind"). Both pinpoint the same defect (result-count mismatch),
+        # so accept either by widening to Exception and matching either
+        # message.
+        with pytest.raises(Exception, match="result type|results but was provided"):
```
fabianlim: the exception strings are quite short.. if they are longer the error is more obvious and increases readiblityl

tests/test_dialects_parse.py
```diff
@@ -522,6 +454,85 @@ def test_get_compute_tile_id_multi(self):
...
+        #
+        # MLIR-frontend backend: the adapter wraps op_text in a func.func
+        # and declares each ``args`` entry as a function parameter — but
+        # MLIR's grammar reserves "%name#N" for *references* to multi-
+        # result ops, so it can't appear in an argument list. That parse
+        # rejection is itself proof that "#N" is a structural part of the
+        # SSA name, not a stripped suffix — which is exactly what this
+        # test guards.
+        op_text = "%c = arith.index_cast %pid#0 : index to i32"
+        try:
```
fabianlim: the convention is to do `with pytest.raises(Exception, ..)` and match on the exception string. This improves robustness of the test, since `Exception` is very generic.

fabianlim: and wouldnt you pass a numeric value into `%pid#0` instead of the keyword `index`?

fabianlim: and its abit hard to understand that this is an MLIR failure case. if this is so, then the better way is to overwrite the test in the adaptor tests to clarify that this is an MLIR frontend specific behavior

ktir_cpu/dialects/ktdp_ops.py
fabianlim: ok i already did the refactoring

ktir_cpu/dialects/ktdp_ops.py
fabianlim: refactoring is done

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -184,26 +184,41 @@ def ktdp__store(op, context, env):
...
 
+@register_parser("ktdp.get_compute_tile_id")
+def parse_get_compute_tile_id(op_text, parse_ctx: ParseContext):
+    m = re.match(
+        r"^(.*?)\s*=\s*ktdp\.get_compute_tile_id\s*:\s*([^{(]*)\s*$", op_text
+    )
+    if not m:
+        return None
+    names = parse_multi_result_lhs(m.group(1))
+    type_count = sum(1 for t in m.group(2).split(",") if t.strip())
```
kiszk: It would be good to throw an exception for unacceptable cases.
```
%pid:2 = ktdp.get_compute_tile_id : index, index,
```
and
```
%pid:2 = ktdp.get_compute_tile_id : index,, index
```

```suggestion
    types_text = m.group(2).strip()
    if not types_text:
        raise ValueError("no result types specified")
    type_list = [t.strip() for t in types_text.split(",")]
    if any(not t for t in type_list):
        raise ValueError(f"empty type in list: {types_text!r}")
    type_count = len(type_list)
```

ktir_cpu/parser.py
```diff
@@ -602,7 +602,9 @@ def _extract_operands(self, text: str, result: Optional[str]) -> List[str]:
         # attribute blocks.
         cleaned = re.sub(r'\{[^}]*\}', '', text)
 
-        operands = re.findall(r'%\w+', cleaned)
+        # "%name#N" is MLIR's multi-result of N ops defined as 
+        # "%base:N = ..." — keep the "#N" attached so results are distinct keys.
+        operands = re.findall(r'%\w+(?:#\d+)?', cleaned)
```
kiszk: DRY

```suggestion
#import at the beginning of this file
from .parser_utils import find_ssa_names
        ...

        operands = find_ssa_names(cleaned)
```

ktir_cpu/parser.py
```diff
@@ -604,7 +605,7 @@ def _extract_operands(self, text: str, result: Optional[str]) -> List[str]:
 
         # "%name#N" is MLIR's multi-result of N ops defined as
```
fabianlim: nit: this comment is out of place now that the regex has moved into find_ssa_names

fabianlim:
@kcirred left a nit and still pending on this [comment](https://github.com/torch-spyre/ktir-cpu/pull/45/files/b6456dca0e4fd06236d8a3ccf29e439d695b153b#diff-afa7780be2ca0769fd862e989061d3e5b0f57cb34f3b79189f26ca6417d25f6c)

fabianlim (approved):
Thanks for addressing all the comments. when the CI passes will merge.

PR #38, opened 2026-05-07
[Refactor] Shared dispatch helpers + full arith dialect coverage 

fabianlim:
Closes #23.

## Summary

- Add `dialects/_helpers.py` with three shared dispatch helpers (`_float_binop`, `_int_binop`, `_unary`) that eliminate the repetitive Tile-vs-scalar branching duplicated across every arith and math handler
- Rewrite `dialects/arith_ops.py` using the helpers; each float/int binop handler is now a one-liner
- Refactor `dialects/math_ops.py` to use `_unary` for all unary math ops
- Add executors for previously missing arith ops: `remf`, `negf`, `absf`, `minf`/`minimumf`, `minnumf`, `cmpf`, `divsi`, `ceildivsi`, `floordivsi`, `remsi`, `andi`, `ori`, `xori`, `shli`, `shrsi`, `shrui`, `extui`, `trunci`, `uitofp`, `fptosi`, `fptoui`
- Add `parse_arith_cmpf` regex parser
- Update `mlir_frontend/parser.py`: register all new ops; add `_adapt_arith_cmpf` predicate-integer → string adapter
- Extend `ops/arith_ops.py` with backing compute methods for all new ops
- Add tests in `test_dialects_exec.py` and `test_dialects_parse.py` covering the new executors and parsers

## Design note

The helpers live in `dialects/_helpers.py` (not `ops/` or `dtypes.py`) because they are IR dispatch patterns — they read `op.operands` and construct `Tile` objects. The `ops/` layer and `dtypes.py` remain pure compute/data with no IR knowledge.

`linalg`, `scf`, and `tensor` dialect handlers were intentionally left unreafactored: their dispatch logic is bespoke (reductions, loop state, shape construction) and does not fit the uniform binop/unary pattern.

## Test plan

- [ ] `uv run pytest tests/ -v` — 665 passed, 6 xfailed, 0 failures

Issue #3, opened 2026-04-22
Transformer decoder op coverage: gap analysis and MLIR examples


fabianlim:
## 🚀 Feature

Map every MLIR operation required to execute a standard transformer decoder
onto the interpreter's current coverage, and produce MLIR example fixtures for
sub-graphs that lack them.

## Motivation

`docs/gap_analysis.md` already catalogues the generic dialect gaps
(`linalg`, `math`, `arith`, `scf`, …) relative to the RFC.  What is missing is
a workload-driven view: given a realistic decoder (Granite), which
sub-graphs can the interpreter already execute end-to-end, and which are blocked
by a known gap?  That map lets us prioritise which gaps from the gap analysis to
close first and tells us exactly which MLIR example fixtures to write.

### Decoder Architectural References

| Model | Norm | Positional embedding | Attention | Activation |
|-------|------|----------------------|-----------|------------|
| [Granite 3.3-8B-Instruct](https://huggingface.co/ibm-granite/granite-3.3-8b-instruct) | RMSNorm | RoPE | GQA (32Q / 8KV) | SiLU → SwiGLU |
| [Granite 4.0-Micro](https://huggingface.co/ibm-granite/granite-4.0-micro) | RMSNorm | RoPE | GQA (40Q / 8KV) | SiLU → SwiGLU |

## Op inventory by decoder sub-graph

Status legend: ✅ interpreted · ❓ no example yet · ❌ not implemented (see
[`docs/gap_analysis.md`](../docs/gap_analysis.md) for the full gap entry).

### 1. Embedding lookup

> Both models: standard token embedding table lookup; indirect gather required
> for autoregressive decode (next-token index → embedding row).
>
> Closest example: `examples/rfc/indirect-access-copy.mlir` — double-indirect
> gather (IDX1, IDX2 tensors). Delta: needs `ktdp.construct_indirect_access_tile`
> (GA-2); the direct lookup path (contiguous slice) is already covered by
> `examples/rfc/add-with-control-flow.mlir`.

| Op | Dialect | Status |
|----|---------|--------|
| `ktdp.construct_memory_view` | ktdp | ✅ |
| `ktdp.construct_access_tile` | ktdp | ✅ |
| `ktdp.load` | ktdp | ✅ |
| `ktdp.construct_indirect_access_tile` (token→embedding gather) | ktdp | ✅ |

### 2. RMSNorm

> Both models use RMSNorm (not LayerNorm). Applied before attention and MLP
[...]

Issue #2, opened 2026-04-22
[RFC conformance] Add RFC-explicit non-ktdp ops


fabianlim:
## 🚀 Feature

Implement the ops the RFC explicitly calls out that are currently missing.
Completing this issue means all RFC example files parse and execute
correctly.

## Motivation

With the Phase 1 and Phase 2 memory model work in place, the remaining
blocker for running the RFC examples end-to-end is a set of standard
dialect ops that are absent from the interpreter. These are not obscure
edge cases — they appear directly in the RFC's own example programs.

### Linalg dialect (`ktir_cpu/dialects/linalg_ops.py`)
- [ ] `linalg.add` — elementwise addition into an output tensor; used in
  `examples/rfc/matrix_add.mlir`
- [ ] `linalg.map` — applies a scalar region elementwise; sibling of `linalg.broadcast` / `linalg.transpose`
- [ ] `tensor.empty` — allocate a zero-initialized output tensor; required
  by `linalg.add`

### Tensor dialect (`ktir_cpu/dialects/tensor_ops.py`)
- [ ] `tensor.extract_slice` — extract a static rectangular subregion from
  a tensor; used in `examples/rfc/matrix_add.mlir`
- [ ] `tensor.generate` — construct a tensor from a function of its indices;
  standard MLIR op used in Triton's lowering of `tl.where` on index-derived
  masks (e.g. causal attention mask from `pos_q >= pos_k`)

### SCF dialect (`ktir_cpu/ops/control_ops.py`)
- [ ] `scf.parallel` and `scf.forall` — parallel loop ops; executed
  sequentially in the CPU interpreter (same approach as `scf.for`)
- [ ] `scf.reduce` and `scf.reduce.return` — reduction accumulation for use
  inside `scf.parallel`

### Arith dialect (`ktir_cpu/dialects/arith_ops.py`)
- [ ] `arith.cmpf` — float comparison (gap analysis row 13); needed for
  rank-counting and LSE inf-guarding patterns
- [ ] `arith.negf` — unary float negation; needed for SiLU (`-x` term)
- [ ] `arith.absf` — unary float absolute value; mirrors `math.absf`
- [ ] `arith.minimumf`, `arith.minnumf` — elementwise float min; mirrors existing `arith.maximumf` / `arith.maxnumf`
- [ ] `arith.divsi`, `arith.remsi` — signed integer divide/remainder; mirrors `arith.divui` / `arith.remui`
[...]

PR #49, opened 2026-05-14
[Feature] Add ktdp.store with IndirectAccessTile (scatter)

### LINKED ISSUE #44: [Feature] ktdp.store with IndirectAccessTile (scatter)
## Background

`ktdp.store` raises `NotImplementedError` when the access tile is an `IndirectAccessTile`:

```python
# ktir_cpu/dialects/ktdp_ops.py
if isinstance(access_tile, IndirectAccessTile):
    raise NotImplementedError("ktdp.store with IndirectAccessTile is not yet supported")
```

`ktdp.load` with `IndirectAccessTile` (gather) is fully implemented via `MemoryOps.indirect_load`. The symmetric scatter path is missing.

## What needs to be done

1. Add `MemoryOps.indirect_store` in `ktir_cpu/ops/memory_ops.py`, mirroring `indirect_load` — enumerate the variable space set, resolve per-dim subscripts, and write each element of the source `Tile` to the target coordinate in `parent_ref`.
2. Wire it into `ktdp__store` in `ktir_cpu/dialects/ktdp_ops.py` (replacing the `NotImplementedError`).
3. Add tests in `tests/test_indirect_access.py`.

## Motivating examples

Two new examples under `examples/rfc/`:

---

### 1. `indirect-scatter.mlir` — indexed scatter

The scatter dual of `indirect-access-copy.mlir`. That example does a gather:
```
Y[m, k] = X[IDX1[m, k], IDX2[m, k]]
```
This example does the reverse — a scatter write:
```
Y[IDX1[m, k], IDX2[m, k]] = X[m, k]
```
`X` and `Y` are `[64, 64] f16`; `IDX1` and `IDX2` are `[64, 64] i32` row/column index tensors.

```mlir
#X_coord_set   = affine_set<(d0, d1) : (d0 >= 0, -d0 + 63 >= 0, d1 >= 0, -d1 + 63 >= 0)>
#IDX_coord_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 63 >= 0, d1 >= 0, -d1 + 63 >= 0)>
#Y_coord_set   = affine_set<(d0, d1) : (d0 >= 0, -d0 + 63 >= 0, d1 >= 0, -d1 + 63 >= 0)>
#var_space_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 63 >= 0, d1 >= 0, -d1 + 63 >= 0)>
#var_space_order = affine_map<(d0, d1) -> (d0, d1)>

module {
  func.func @indirect_scatter() {
    %X_addr    = arith.constant 0     : index   // X  [64, 64] f16
    %IDX1_addr = arith.constant 8192  : index   // IDX1 [64, 64] i32
    %IDX2_addr = arith.constant 16384 : index   // IDX2 [64, 64] i32
    %Y_addr    = arith.constant 24576 : index   // Y  [64, 64] f16

    // (1) Memory views
    %X_view = ktdp.construct_memory_view %X_addr, sizes: [64, 64], strides: [64, 1] {
        coordinate_set = #X_coord_set,
        memory_space   = #ktdp.spyre_memory_space<HBM>
    } : memref<64x64xf16>

    %IDX1 = ktdp.construct_memory_view %IDX1_addr, sizes: [64, 64], strides: [64, 1] {
        coordinate_set = #IDX_coord_set,
        memory_space   = #ktdp.spyre_memory_space<HBM>
    } : memref<64x64xi32>

    %IDX2 = ktdp.construct_memory_view %IDX2_addr, sizes: [64, 64], strides: [64, 1] {
        coordinate_set = #IDX_coord_set,
        memory_space   = #ktdp.spyre_memory_space<HBM>
    } : memref<64x64xi32>

    %Y_view = ktdp.construct_memory_view %Y_addr, sizes: [64, 64], strides: [64, 1] {
        coordinate_set = #Y_coord_set,
        memory_space   = #ktdp.spyre_memory_space<HBM>
    } : memref<64x64xf16>

    // (2) Direct access tile for X[m, k] — load source contiguously
    %X_access_tile = ktdp.construct_access_tile %X_view[%c0, %c0] {
        access_tile_set   = #X_coord_set,
        access_tile_order = #var_space_order
    } : memref<64x64xf16> -> !ktdp.access_tile<64x64xindex>

    // (3) Indirect access tile for Y[IDX1[m,k], IDX2[m,k]] — scatter destination
    %Y_access_tile = ktdp.construct_indirect_access_tile
        intermediate_variables(%m, %k)
        %Y_view[ind(%IDX1[%m, %k]), ind(%IDX2[%m, %k])] {
            variables_space_set   = #var_space_set,
            variables_space_order = #var_space_order
        } : memref<64x64xf16>, memref<64x64xi32>, memref<64x64xi32> -> !ktdp.access_tile<64x64xindex>

    // (4) Load X[m, k]
    %X_data = ktdp.load %X_access_tile : !ktdp.access_tile<64x64xindex> -> tensor<64x64xf16>

    // (5) Scatter: Y[IDX1[m,k], IDX2[m,k]] = X[m, k]
    ktdp.store %X_data, %Y_access_tile : tensor<64x64xf16>, !ktdp.access_tile<64x64xindex>

    return
  }
}
```

---

### 2. `paged-tensor-write.mlir` — paged write (stretch goal)

The scatter dual of `paged-tensor-copy.mlir`. That example reads from a paged layout into a contiguous output:
```
Y[b, tkv, h, dkv] = X[Idx[b, tkv/Ptkv], h, tkv%Ptkv, dkv]
```
This example writes from a contiguous input into a paged layout:
```
Y[Idx[b, tkv/Ptkv], h, tkv%Ptkv, dkv] = X[b, tkv, h, dkv]
```
Same shapes: `X` is `[4, 2048, 8, 128] f16` (contiguous, dim order `{Nb, Ntkv, Nh, Ndkv}`); `Y` is `[10000, 8, 64, 128] f16` (paged, dim order `{Npages, Nh, Ptkv, Ndkv}`); `Idx` is `[4, 32] i32`.

```mlir
// (shapes, coord sets, var space sets identical to paged-tensor-copy.mlir —
//  X and Y roles are swapped)

// (2) Direct access tile for X[b, tkv, h, dkv] — load source contiguously
%X_access_tile = ktdp.construct_access_tile %X_mem_view[%c0, %c0, %c0, %c0] {
    access_tile_set   = #Y_var_space_set,
    access_tile_order = #Y_var_space_order
} : memref<4x2048x8x128xf16> -> !ktdp.access_tile<4x8x2048x128xindex>

// (3) Indirect access tile for Y[Idx[b, tkv/Ptkv], h, tkv%Ptkv, dkv] — scatter destination
%Y_access_tile = ktdp.construct_indirect_access_tile
    intermediate_variables(%b, %h, %tkv, %dkv)
    %Y_mem_view[ind(%Idx_mem_view[%b, %tkv floordiv 64]), (%h), (%tkv mod 64), (%dkv)] {
        variables_space_set   = #X_var_space_set,
        variables_space_order = #X_var_space_order
    } : memref<10000x8x64x128xf16>, memref<4x32xi32> -> !ktdp.access_tile<4x8x2048x128xindex>

// (4) Load X[b, tkv, h, dkv]
%X_data = ktdp.load %X_access_tile : !ktdp.access_tile<4x8x2048x128xindex> -> tensor<4x8x2048x128xf16>

// (5) Scatter: Y[Idx[b, tkv/Ptkv], h, tkv%Ptkv, dkv] = X[b, tkv, h, dkv]
ktdp.store %X_data, %Y_access_tile : tensor<4x8x2048x128xf16>, !ktdp.access_tile<4x8x2048x128xindex>
```

---

cc: @nwang-ibm

WarningRan:
## Summary

Closes #44.

`ktdp.store` previously raised `NotImplementedError` for `IndirectAccessTile`. This PR adds `MemoryOps.indirect_store`, mirroring `MemoryOps.indirect_load`, and wires it into `ktdp__store`.

### Main: scatter support (Issue #44)

- `MemoryOps.indirect_store` — enumerates `variables_space_set`, resolves each per-dim subscript (`direct` / `direct_expr` / `indirect`), and delegates to `MemoryOps.store` with the resolved coordinate list.
- `ktdp__store` dispatches to `indirect_store` for `IndirectAccessTile`.
- Source-tile shape is validated against IAT shape on entry (raises `ValueError` instead of an opaque downstream NumPy error).
- Negative indirect indices raise `IndexError` (NumPy fancy-index assignment would otherwise silently wrap from the tail).
- New MLIR fixtures:
  - `examples/rfc/indirect-scatter.mlir` — 64×64 scatter dual of `indirect-access-copy.mlir`.
  - `examples/rfc/paged-tensor-write.mlir` — scatter dual of `paged-tensor-copy.mlir` (LX-overflow `xfail`, same constraint as gather).

### Additional fix: `indirect_load` `vso` path

While writing the scatter mirror I noticed `indirect_load`'s non-identity `variables_space_order` branch was broken: it does `points = [vso.eval(pt) for pt in points]`, which permutes the components of `pt` in place — this corrupts the downstream `pt[sub["var_index"]]` and `eval_subscript_expr(..., pt)` bindings, and does not reorder the enumeration either.

Grep of `examples/` and `tests/` confirms **no fixture exercises non-identity `variables_space_order` on an IAT** today. (The non-identity `affine_map` in `paged-tensor-copy.mlir` is on `access_tile_order` of a `construct_access_tile`, which is a different operand on a different op.) So this is dead-but-broken code, not a regression risk.

Two changes:
- Tightened to `raise NotImplementedError(...)` with a clear message, guarded by `not vso.is_identity()` as belt-and-suspenders against parser changes (parser currently nullifies identity `vso`, but the explicit check is free).
- Backported the per-dim resolver pre-compilation (loop-invariant hoisting) from `indirect_store` to keep the two paths structurally aligned.

Happy to split this into a separate PR if reviewers prefer.

### Tests

- `test_small_indirect_scatter` — 4×4 bijection scatter, verifies data correctness (X rotated 180°).
- `test_small_indirect_scatter_collision` — all source elements map to `Y[0,0]`; pins current last-writer-wins behavior with a docstring noting it is implementation-defined (see open question below).
- `test_indirect_scatter_rfc` — 64×64 smoke test (zero-seeded); verifies the RFC fixture parses and runs end-to-end. Correctness is covered by the 4×4 tests above.
- `test_paged_tensor_indirect_scatter` — `xfail(strict=True)` for the production-sized paged variant; same LX-overflow constraint as the gather counterpart.

## Open question: coordinate collision semantics

When two source points `(m, k)` map to the same destination `Y[r, c]`, this PR currently exhibits **last-writer-wins** behavior, inherited directly from NumPy fancy-index assignment. This is documented as *implementation-defined* in the `indirect_store` docstring and locked by `test_small_indirect_scatter_collision` (with an explicit comment that the test pins NumPy behavior, not a spec guarantee).

Worth flagging: NumPy itself documents the duplicate-index assignment case as **undefined** — the CPython implementation is stable last-writer-wins, but not contractual. The collision test therefore depends on a NumPy implementation detail; the docstring's *implementation-defined* hedge reflects that.

RFC 0682 does not specify collision semantics. I picked last-writer-wins for this PR because:

1. The intended use cases (permutation, reordering) are bijections — collisions don't occur in well-formed inputs.
2. Reduction variants (add / max / error-on-collision) are user-driven design choices; without a concrete need, picking one risks over-design.
3. Hardware support for atomic-reduction scatter on Spyre is unclear — I'd rather not commit interpreter semantics ahead of the hardware story.

**Which option do you prefer for this PR?**

- (a) raise on collision (strictest, catches bugs early)
- (b) commit to last-writer-wins as spec behavior (drop the *implementation-defined* hedge in the docstring)
- (c) add a `reduction=` kwarg now with `overwrite` as default (forward-compat, but more surface area)
- (d) stay with the current "implementation-defined, currently last-writer-wins" stance and revisit when a user need surfaces

**Update**: See #60 for the ongoing discussion; this PR implements (d) as a placeholder pending that decision.

## Suggested follow-ups (not in this PR)

While implementing this PR I noticed several issues — `MemoryOps` interface limitations and gather-side cleanups. None are caused by `indirect_store` (the architectural ones exist symmetrically in `indirect_load`); none are triggered by current MLIR fixtures (all rectangular, identity-vso). Flagging them here for reviewer judgment: **if you agree any of these are worth addressing, I'm happy to open follow-up issues to track them.** Otherwise feel free to tell me to drop any that aren't worth the bookkeeping.

I'd suggest two follow-up issues — one architectural, one tactical — since each group shares a coherent scope.

### Architectural (one umbrella issue; shared root: `MemoryOps`'s "per-point coords + dense rectangular shape" contract is a poor fit for non-trivial iteration patterns)

1. **Non-identity `variables_space_order` proper support.** Currently `raise NotImplementedError` in both `indirect_load` and `indirect_store`. Correct handling: keep `vss.enumerate` order unchanged and apply `vso.eval(pt)` on the **tile-side** index (post-gather permutation for load, dual for store) — `vso` is a tile-layout map, not a rebinding of variables. **Tracked in #58.**

2. **`MemoryOps.bulk_read` / vectorized index gather.** Both `indirect_load` and `indirect_store` instantiate a `_MemAccessor` and call `.read()` once per inner-loop iteration. For a 64×64 IAT with one indirect dim that's 4096 per-point reads; for production paged kernels it would be millions. A bulk read API that returns the full index ndarray once, with NumPy fancy indexing in the loop, is the right fix — must touch gather and scatter together. **Covered by #52.**

3. **Mask / non-rectangular `variables_space_set` contract.** If `vss.enumerate(shape)` returns fewer points than `shape.prod()` (e.g. a triangular constraint), `MemoryOps.store(coords=...)`'s `flat[offsets] = tile.data.flatten()` raises `ValueError: shape mismatch` — and even if it didn't, the positional alignment between the dense source tile and the masked coords is undefined. The same issue exists symmetrically in `MemoryOps.load` (`reshape(out_shape)` on the gathered values). All current MLIR uses rectangular sets so this never trips, but the contract needs a defined semantic. **Tracked in #56.**

### Gather-side cleanup (one paired issue)

4. **Symmetric negative-index check for `indirect_load` + Idx/Y address overlap in `paged-tensor-copy.mlir`.** Two independent gather-side bugs that pair naturally:
   - **Negative-index check on `indirect_load`**: mirror of the store-side check this PR adds. `indirect_load` reads `raw_idx` at `memory_ops.py:381` with no guard against negative values, so a negative index from the lookup would silently wrap from the tail (NumPy fancy-indexing semantics). One-line `raise IndexError` fix, same shape as the store-side guard at `memory_ops.py:459-462`. **Directly addressed in this PR (commit `dd84f61`).**
   - **Idx/Y address overlap in `paged-tensor-copy.mlir`**: latent test-fixture bug. `Y_start_address = 10000000`, `Idx_start_address = 20000000`, but Y's footprint (`128 × 8 × 4 × 2048` f16 elements ≈ 16 MB) extends past 20000000, so Y physically overlaps Idx. The current zero-seed masks it (zeros are benign in either tensor); a non-zero seed would let Y writes corrupt the gather indices. Fix is to relayout the fixture's address constants. **Tracked in #57.**

cc  @fabianlim @kiszk  @lchu6 @nwang-ibm

### DIFF
--- a/examples/rfc/paged-tensor-write.mlir
+++ b/examples/rfc/paged-tensor-write.mlir
@@ -7,17 +7,17 @@
 // Idx input = 2D tensor, shape {Nb, Ntkv/Ptkv} = {4, 32} i32
 // Y output = paged 4D tensor, shape {Npages, Nh, Ptkv, Ndkv} = {10000, 8, 64, 128} f16
 //
-// Affine declarations are kept verbatim from paged-tensor-copy.mlir; the
-// `var_space` qualifier refers to the iteration variable layout, not to which
-// tensor the IAT acts on. So `#X_var_space_*` continues to describe the
+// The `var_space` declarations are kept verbatim from paged-tensor-copy.mlir;
+// the `var_space` qualifier refers to the iteration variable layout, not to
+// which tensor the IAT acts on. So `#X_var_space_*` continues to describe the
 // indirect-access (b, h, tkv, dkv) iteration even though X is now the
 // (contiguous) source.
 
-#X_coord_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 9999 >= 0,
+#paged_coord_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 9999 >= 0,
                                               d1 >= 0, -d1 + 7 >=0,
                                               d2 >= 0, -d2 + 63 >= 0,
                                               d3 >= 0, -d3 + 127>= 0)>
-#Y_coord_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 3 >= 0,
+#contiguous_coord_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 3 >= 0,
                                               d1 >= 0, -d1 + 2047 >=0,
                                               d2 >= 0, -d2 + 7 >= 0,
                                               d3 >= 0, -d3 + 127>= 0)>
@@ -65,7 +65,7 @@ module {
         %X_str_Nb = arith.muli %X_str_Ntkv, %Ntkv : index
         %X_mem_view = ktdp.construct_memory_view %X_start_address,
                         sizes: [%Nb, %Ntkv, %Nhkv, %Ndkv], strides: [%X_str_Nb, %X_str_Ntkv, %Ndkv, 1] {
-                        coordinate_set = #Y_coord_set,
+                        coordinate_set = #contiguous_coord_set,
                         memory_space = #ktdp.spyre_memory_space<HBM>
         } : memref<4x2048x8x128xf16>
 
@@ -73,7 +73,7 @@ module {
         // dim order (outermost to innermost) {Npages, Nh, Ptkv, Ndkv}
         %Y_mem_view = ktdp.construct_memory_view %Y_start_address,
                         sizes: [%Npages, %Nhkv, %Ptkv, %Ndkv], strides: [65536, 8192, %Ndkv, 1] {
-                        coordinate_set = #X_coord_set,
+                        coordinate_set = #paged_coord_set,
                         memory_space = #ktdp.spyre_memory_space<HBM>
         } : memref<10000x8x64x128xf16>
 
--- a/examples/rfc/paged-tensor-write.mlir
+++ b/examples/rfc/paged-tensor-write.mlir
@@ -1,0 +1,105 @@
+// Scatter dual of paged-tensor-copy.mlir.
+// Same semantics, X and Y roles swapped:
+//   paged-tensor-copy:  Y[b][tkv][h][dkv] = X[Idx[b][tkv/Ptkv]][h][tkv%Ptkv][dkv]
+//   paged-tensor-write: Y[Idx[b][tkv/Ptkv]][h][tkv%Ptkv][dkv] = X[b][tkv][h][dkv]
+//
+// X input  = contiguous 4D tensor, shape {Nb, Ntkv, Nh, Ndkv} = {4, 2048, 8, 128} f16
+// Idx input = 2D tensor, shape {Nb, Ntkv/Ptkv} = {4, 32} i32
+// Y output = paged 4D tensor, shape {Npages, Nh, Ptkv, Ndkv} = {10000, 8, 64, 128} f16
+//
+// Affine declarations are kept verbatim from paged-tensor-copy.mlir; the
+// `var_space` qualifier refers to the iteration variable layout, not to which
+// tensor the IAT acts on. So `#X_var_space_*` continues to describe the
+// indirect-access (b, h, tkv, dkv) iteration even though X is now the
+// (contiguous) source.
+
+#X_coord_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 9999 >= 0,
+                                              d1 >= 0, -d1 + 7 >=0,
+                                              d2 >= 0, -d2 + 63 >= 0,
+                                              d3 >= 0, -d3 + 127>= 0)>
+#Y_coord_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 3 >= 0,
+                                              d1 >= 0, -d1 + 2047 >=0,
+                                              d2 >= 0, -d2 + 7 >= 0,
+                                              d3 >= 0, -d3 + 127>= 0)>
+#X_var_space_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 3 >= 0,
+                                                  d1 >= 0, -d1 + 7 >= 0,
+                                                  d2 >= 0, -d2 + 2047>= 0,
+                                                  d3 >= 0, -d3 + 127 >= 0)>
+#Y_var_space_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 3 >= 0,
+                                                  d1 >= 0, -d1 + 2047>= 0,
+                                                  d2 >= 0, -d2 + 7 >= 0,
+                                                  d3 >= 0, -d3 + 127 >= 0)>
+#X_var_space_order = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+#Y_var_space_order = affine_map<(d0, d1, d2, d3) -> (d0, d2, d1, d3)>
+
+module {
+  func.func @paged_tensor_write_1core() {
+        %Nb = arith.constant 4 : index
+        %Ntkv = arith.constant 2048 : index
+        %Ptkv = arith.constant 64 : index
+        %Ndkv = arith.constant 128 : index
+        %Nhkv = arith.constant 8 : index
+        %Npages = arith.constant 10000 : index
+        %Ntkv_Ptkv = arith.divui %Ntkv, %Ptkv : index
+
+        // Memory layout: addresses are spaced so all three tensors are byte-disjoint.
+        //   X  spans [10M, 26.78M)  — 4*2048*8*128 f16 = 16.78 MB (contiguous source)
+        //   Idx spans [30M, 30M+512] — 4*32 i32 = 512 B (page table)
+        //   Y  starts at 40M         — 10000*8*64*128 f16 = 1.31 GB (paged dest)
+        // (paged-tensor-copy.mlir uses overlapping addresses; we don't replicate
+        // that here — see test_spec_gaps.py for the matching seed values.)
+        %X_start_address = arith.constant 10000000 : index   // X (contiguous source)
+        %Idx_start_address = arith.constant 30000000 : index // Idx (page table)
+        %Y_start_address = arith.constant 40000000 : index   // Y (paged destination)
+
+        // (1) Construct memory view for Index tensor
+        %Idx_mem_view = ktdp.construct_memory_view %Idx_start_address,
+                        sizes: [%Nb, %Ntkv_Ptkv], strides: [%Ntkv_Ptkv, 1] {
+                        coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + 3 >= 0, d1 >= 0, -d1 + 31 >= 0)>,
+                        memory_space = #ktdp.spyre_memory_space<HBM>
+        } : memref<4x32xi32>
+
+        // (1) Construct memory view for input X (contiguous)
+        // dim order (outermost to innermost) {Nb, Ntkv, Nh, Ndkv}
+        %X_str_Ntkv = arith.muli %Ndkv, %Nhkv : index
+        %X_str_Nb = arith.muli %X_str_Ntkv, %Ntkv : index
+        %X_mem_view = ktdp.construct_memory_view %X_start_address,
+                        sizes: [%Nb, %Ntkv, %Nhkv, %Ndkv], strides: [%X_str_Nb, %X_str_Ntkv, %Ndkv, 1] {
+                        coordinate_set = #Y_coord_set,
+                        memory_space = #ktdp.spyre_memory_space<HBM>
+        } : memref<4x2048x8x128xf16>
+
+        // (1) Construct memory view for output Y (paged)
+        // dim order (outermost to innermost) {Npages, Nh, Ptkv, Ndkv}
+        %Y_mem_view = ktdp.construct_memory_view %Y_start_address,
+                        sizes: [%Npages, %Nhkv, %Ptkv, %Ndkv], strides: [65536, 8192, %Ndkv, 1] {
+                        coordinate_set = #X_coord_set,
+                        memory_space = #ktdp.spyre_memory_space<HBM>
+        } : memref<10000x8x64x128xf16>
+
+        // (2) Direct access tile for X[b, tkv, h, dkv] — load source contiguously
+        // access_tile_set/order use the (b, tkv, h, dkv) iteration so flatten()
+        // of the loaded tile aligns with the (b, h, tkv, dkv) IAT enumeration.
+        %c0 = arith.constant 0 : index
+        %X_access_tile = ktdp.construct_access_tile %X_mem_view[%c0, %c0, %c0, %c0] {
+            access_tile_set   = #Y_var_space_set,
+            access_tile_order = #Y_var_space_order
+        } : memref<4x2048x8x128xf16> -> !ktdp.access_tile<4x8x2048x128xindex>
+
+        // (3) Indirect access tile for Y[Idx[b, tkv/Ptkv], h, tkv%Ptkv, dkv]
+        %Y_access_tile = ktdp.construct_indirect_access_tile
+            intermediate_variables(%b, %h, %tkv, %dkv)
+            %Y_mem_view[ind(%Idx_mem_view[%b, %tkv floordiv 64]), (%h), (%tkv mod 64), (%dkv)] {
+                variables_space_set   = #X_var_space_set,
+                variables_space_order = #X_var_space_order
+            } : memref<10000x8x64x128xf16>, memref<4x32xi32> -> !ktdp.access_tile<4x8x2048x128xindex>
+
+        // (4) Load X[b, tkv, h, dkv]
+        %X_data_tile = ktdp.load %X_access_tile : !ktdp.access_tile<4x8x2048x128xindex> -> tensor<4x8x2048x128xf16>
+
+        // (5) Scatter: Y[Idx[b, tkv/Ptkv], h, tkv%Ptkv, dkv] = X[b, tkv, h, dkv]
+        ktdp.store %X_data_tile, %Y_access_tile : tensor<4x8x2048x128xf16>, !ktdp.access_tile<4x8x2048x128xindex>
+
+        return
+  }
+}
--- a/ktir_cpu/ops/memory_ops.py
+++ b/ktir_cpu/ops/memory_ops.py
@@ -347,19 +347,15 @@ class MemoryOps:
         (direct dims use the variable value, indirect dims look up the
         index in an index memref), then delegates to :meth:`load`.
 
-        Raises NotImplementedError if ``variables_space_order`` is non-identity
-        (rebinding intermediate variables under a permutation is unimplemented).
+        Raises NotImplementedError if ``variables_space_order`` is non-identity.
         """
         vss = iat.variables_space_set
         vso = iat.variables_space_order
 
-        # Reject permuted vso: applying it in-place would corrupt the variable
-        # bindings consumed below and would not reorder the enumeration anyway.
         if vso is not None and not vso.is_identity():
             raise NotImplementedError(
-                "indirect_load: non-identity variables_space_order is not yet "
-                "supported (rebinding intermediate variables under a permutation "
-                "is unimplemented)"
+                "indirect_load: output tile permutation via non-identity "
+                "variables_space_order is not yet implemented"
             )
 
         # Pre-compile per-dim resolvers so branch dispatch and constant lookups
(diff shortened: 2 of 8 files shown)


WarningRan:

ktir_cpu/ops/memory_ops.py
```diff
@@ -330,36 +330,145 @@ def indirect_load(
...
+            coords.append(tuple(coord))
+
+        # Make the C-order contract with MemoryOps.store explicit; no-op when
+        # the source is already contiguous.
+        src_data = np.ascontiguousarray(tile.data)
+        if src_data is not tile.data:
+            src_tile = Tile(src_data, tile.dtype, tile.shape, tile.unique_sticks)
+        else:
+            src_tile = tile
+
```
fabianlim: This contiguity normalisation isn't needed. `MemoryOps.store` with `coords=` uses a read-modify-write scatter (`flat[offsets] = tile.data.flatten()`); `flatten()` always returns a new contiguous copy, so the `ascontiguousarray` dance here is a no-op at best and an extra full-tile copy at worst. Safe to drop the whole block and pass `tile` directly to `MemoryOps.store`.

ktir_cpu/ops/memory_ops.py
```diff
@@ -330,36 +330,145 @@ def indirect_load(
...
+                    raw_idx = int(_MemAccessor(context, idx_view, addr).read(1, idx_view.dtype)[0])
+                    # NumPy fancy-index assignment silently wraps negative
+                    # indices; reject them here. Use raise so the check
+                    # survives python -O.
+                    if raw_idx < 0:
+                        raise IndexError(
+                            f"indirect index {raw_idx} from {idx_view} is negative"
+                        )
+                    coord.append(raw_idx)
+            coords.append(tuple(coord))
```
fabianlim, line 740: The negative-index guard added to `indirect_store` is missing here in `indirect_load`. NumPy fancy-indexing silently wraps negative values on reads too, so a corrupt index tensor would silently produce wrong gathered data rather than raising. The fix is symmetric — one line, same shape as the store-side guard:
```python
if raw_idx < 0:
    raise IndexError(f"indirect index {raw_idx} from {idx_view} is negative")
```

ktir_cpu/ops/memory_ops.py
```diff
@@ -330,36 +330,145 @@ def indirect_load(
...
+        # Reject permuted vso: applying it in-place would corrupt the variable
+        # bindings consumed below and would not reorder the enumeration anyway.
+        if vso is not None and not vso.is_identity():
+            raise NotImplementedError(
+                "indirect_load: non-identity variables_space_order is not yet "
+                "supported (rebinding intermediate variables under a permutation "
+                "is unimplemented)"
+            )
+
+        # Pre-compile per-dim resolvers so branch dispatch and constant lookups
```
fabianlim, line 361: The error message says "rebinding intermediate variables under a permutation is unimplemented" — but that's not quite the right framing. `vso` doesn't rebind variables; it controls which slot in the output tile a given variable-space point maps to. The variables themselves stay fixed; only the output position changes.

I think this is actually fixable without too much work. The key insight is that `vso` only affects the tile layout, not the subscript resolution:

```
For each pt in vss.enumerate():
  coord  = resolve_subscripts(pt)   ← always uses original pt, unchanged
  out_pos = vso.eval(pt)            ← where this element lives in the tile
```

Concretely for load:
```
variable space     subscript resolution     output tile
(d0=0,d1=0) ──► X[ IDX1[0,0], IDX2[0,0] ] ──► tile[ vso(0,0) ]
(d0=0,d1=1) ──► X[ IDX1[0,1], IDX2[0,1] ] ──► tile[ vso(0,1) ]
...
```

With identity vso (current), enumeration order = tile order, so the gathered values land in the right positions automatically. With non-identity vso, you'd gather in enumeration order but then permute the result array using `vso.eval(pt)` as the destination index before returning the tile.

Happy to leave this as a follow-up issue — just wanted to correct the diagnosis so whoever implements it starts from the right place. Suggest updating the error message to something like: "output tile permutation via variables_space_order is not yet implemented".

examples/rfc/paged-tensor-write.mlir
```diff
@@ -0,0 +1,105 @@
...
+// indirect-access (b, h, tkv, dkv) iteration even though X is now the
+// (contiguous) source.
+
+#X_coord_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 9999 >= 0,
+                                              d1 >= 0, -d1 + 7 >=0,
+                                              d2 >= 0, -d2 + 63 >= 0,
+                                              d3 >= 0, -d3 + 127>= 0)>
+#Y_coord_set = affine_set<(d0, d1, d2, d3) : (d0 >= 0, -d0 + 3 >= 0,
+                                              d1 >= 0, -d1 + 2047 >=0,
+                                              d2 >= 0, -d2 + 7 >= 0,
```
fabianlim, line 22: The coord set names are inverted relative to the tensors that use them: `%X_mem_view` (the contiguous source, shape `{4,2048,8,128}`) gets `#Y_coord_set` (bounds `{Nb=4,Ntkv=2048,Nh=8,Ndkv=128}`), while `%Y_mem_view` (the paged destination, shape `{10000,8,64,128}`) gets `#X_coord_set` (bounds `{Npages=10000,Nh=8,Ptkv=64,Ndkv=128}`). The bounds are correct; only the names are swapped. The comment at the top of the file explains why (names were kept from `paged-tensor-copy.mlir` where X/Y roles were different), but it's still easy to misread. Consider renaming to something role-neutral like `#contiguous_coord_set` / `#paged_coord_set`.

ktir_cpu/ops/memory_ops.py
fabianlim, line 740: Rebase note: the  #38 refactors `_MemAccessor` away — `load`/`store` now dispatch directly via `tile_ref.memory_space` rather than through an accessor object. The two call sites added in this PR:
```python
_MemAccessor(context, idx_view, addr).read(1, idx_view.dtype)
```
will need to be updated on rebase. The fix should be straightforward once the new helper shape is clear from that branch.

fabianlim:
Thanks for the detailed follow-up notes.

**Bulk read (#2):** covered by #52 — the indirect-side fast path there addresses the per-point index-tensor reads.

**Non-rectangular `variables_space_set` contract (#3):** worth a separate issue. The current behaviour (passing a masked coord list to `MemoryOps.store`'s `flat[offsets] = tile.data.flatten()`) silently misaligns source elements against destination coords when the set is non-rectangular — the positional pairing between the dense tile and the masked coord list is undefined. There are also tests that just return a flat list for non-rectangular sets which is likely wrong. Can we open an issue for this?

**Gather-side cleanup (#4):** agreed, worth tracking. Can we open a second issue covering the negative-index guard on `indirect_load` and the `paged-tensor-copy.mlir` Y/Idx address overlap?

examples/rfc/paged-tensor-write.mlir
WarningRan: Done. Renamed `#X_coord_set` → `#paged_coord_set` and `#Y_coord_set` → `#contiguous_coord_set` — both definitions and references. The lines using them now read naturally (`%X_mem_view` (contiguous source) gets `#contiguous_coord_set`; `%Y_mem_view` (paged dest) gets `#paged_coord_set`).

Also scoped the top-of-file "kept verbatim" comment to the `var_space_*` declarations only, since `coord_set` is no longer verbatim.

ktir_cpu/ops/memory_ops.py
WarningRan: Thanks - your model of `vso` makes much more sense than what I had.

In this PR I just cleaned up the comments and error messages that came from the wrong model (the "rebinding variables" framing) and adopted your suggested wording for the `NotImplementedError`. Will file the follow-up issue for actually implementing the post-gather permutation step once this PR lands.

ktir_cpu/ops/memory_ops.py
WarningRan: Done. Mirrored the store-side guard to `indirect_load` — same shape, same error message. Comment wording flipped (`fancy-index reads` vs `fancy-index assignment`) so the two sites read symmetrically when diffed against each other.

This also closes the gather-side cleanup item I had originally planned to file as a separate follow-up issue (it's now in this PR alongside the store-side guard).

ktir_cpu/ops/memory_ops.py
```diff
@@ -294,6 +294,12 @@ def store(
         to those local coordinates via a read-modify-write on the allocation.
         When *coords* is None, stores the full tile (contiguous or strided).
 
+        Source data layout: ``tile.data`` is read in C-order via
```
WarningRan, line 307: Drive-by addition while addressing the `ascontiguousarray` comment below on `indirect_store`.

The contract that `tile.data` is read in C-order via `flatten()` was previously implicit — that's exactly the gap that led me to add the redundant `ascontiguousarray` block in `indirect_store`. Documenting it here so the next caller doesn't repeat the same defensive workaround.

Reasonable to keep in this PR, or would you prefer a separate docstring-only PR for scope cleanliness?

ktir_cpu/ops/memory_ops.py
WarningRan: Done. Confirmed `MemoryOps.store` line 319 (`flat[offsets] = tile.data.flatten()`) always returns a contiguous C-order copy — the `ascontiguousarray` block was redundant. Removed.

WarningRan:
Thanks for the thorough review, @fabianlim.

PR description updated; review feedback addressed in `2e67ba3` and `dd84f61`. Follow-up issues filed: #56, #57, #58.

**Scope note on #57:** the "gather-side cleanup" pair originally also covered a symmetric negative-index guard on `indirect_load`. That guard has been **directly addressed in this PR** (commit `dd84f61`), so #57 is scoped to just the `paged-tensor-copy.mlir` Y/Idx address overlap.

**Rebase after #38:** acknowledged — planning to handle this once everything else is settled.

---

**Still awaiting your input:** the collision-semantics question in the PR description (options a/b/c/d, recommendation (d)) is a spec/RFC-level decision and would benefit from an explicit reviewer call.

ktir_cpu/ops/memory_ops.py
fabianlim: its ok to keep this in this PR

fabianlim:
@WarningRan i had read your responses, they look good and I have resolved the converstations. For the collission semantics can we open a discussion issue to track it. you are good to rebase and tag me again for final review.

WarningRan:
Collision semantics tracked in #60.

ktir_cpu/ops/memory_ops.py
WarningRan: @fabianlim rebased onto main, thanks!

fabianlim (commented):
Two small notes:

**Test gap**: the negative-index guard (`IndexError` on negative `raw_idx`) exists in both `indirect_load` and `indirect_store` but isn't exercised by any test. Worth adding a short parametrized case.

**Duplication**: the resolver-building loop and point-enumeration loop in `indirect_load` / `indirect_store` are ~60 lines of near-verbatim copy. Not blocking, but #52 (bulk-read / vectorized index gather) will need to touch both anyway — could be a natural time to extract a shared helper.

WarningRan:
> **Test gap**: the negative-index guard (`IndexError` on negative `raw_idx`) exists in both `indirect_load` and `indirect_store` but isn't exercised by any test. Worth adding a short parametrized case.

Thank you for this comments, @fabianlim ! Added the negative-index test — `test_negative_indirect_index_raises` parametrized over `indirect_load` and `indirect_store`, asserts `IndexError` when IDX1 has a negative entry. Reuses the existing 4×4 MLIR fixtures, no new ones.

On placement: kept it at the end of the file rather than splitting into two tests (one per section), since the parametrized form crosses both gather and scatter.

fabianlim (commented):
**Latency accounting gap: index tensor sticks not counted**

`indirect_load` and `indirect_store` read each index element via `_MemAccessor`, but the sticks touched by those reads are never accumulated into `unique_sticks`. The `Tile` returned by `indirect_load` (and the load inside `indirect_store`) gets its `unique_sticks` from `_flat_memory_offsets` on the data tensor only — the index tensor's HBM accesses are invisible to the latency model.

For a 64×64 IAT with two index tensors, that's up to 4096 × 2 index reads whose sticks go unaccounted, which will undercount latency estimates.

The direct case (`MemoryOps.load`) already handles this correctly via `_flat_memory_offsets`. The indirect paths need to accumulate the index-read sticks and add them into the final `unique_sticks` on the returned tile (or thread them through as a separate accounting field).

Not blocking for this PR — the latency model is approximate and this is a pre-existing gap in `indirect_load` that `indirect_store` inherits — but worth tracking explicitly.

fabianlim (approved):
Approved. thank you for your contribution. Please follow up with the additional issues, and especially there seems to be some divergence with how indirect load / store uses `MemAccessor` with the direct case (tracked in https://github.com/torch-spyre/ktir-cpu/issues/61).

PR #54, opened 2026-05-15
[Fix] Support 1-D grid attribute in parser (grid = [X])

### LINKED ISSUE #41: [Bug] Regex Parser does not support 1-D grid attribute (grid = [X])
## Summary

`_parse_grid_dimensions` in `parser.py` only matched 2- and 3-element `grid` attributes:

```
grid = [X, Y]
grid = [X, Y, Z]
```

A 1-D form `grid = [X]` silently fell through to the `(1, 1, 1)` default instead of returning `(X, 1, 1)`.

**Affected code:** `ktir_cpu/parser.py:_parse_grid_dimensions`

## Fix

Extended the regex to accept an optional second and third element, defaulting missing dims to 1. Fixed in commit 1a9e332.

Before:
```python
grid_match = re.search(r'grid\s*=\s*\[(\d+),\s*(\d+)(?:,\s*(\d+))?\]', func_header)
y = int(grid_match.group(2))
```

After:
```python
grid_match = re.search(
    r'grid\s*=\s*\[(\d+)(?:,\s*(\d+))?(?:,\s*(\d+))?\]',
    func_header,
)
y = int(grid_match.group(2)) if grid_match.group(2) else 1
```

## Test

`TestModuleParser::test_parser_1d_grid` added to cover the regression:

```python
def test_parser_1d_grid(self):
    # grid = [X] — single element; Y and Z should default to 1
    parser = KTIRParser()
    module = parser.parse_module("""
    module {
      func.func @single() attributes { grid = [4] } {
        return
      }
    }
    """)
    assert module.get_function("single").grid == (4, 1, 1)
```

fabianlim:
## Summary

- Extends `_parse_grid_dimensions` in `ktir_cpu/parser.py` to accept a single-element `grid = [X]` form, treating missing Y and Z as 1
- Previously, `grid = [X]` silently fell through to the `(1, 1, 1)` default instead of returning `(X, 1, 1)`
- Adds `TestModuleParser::test_parser_1d_grid` to cover the regression

Fixes #41

## Test plan

- [ ] `uv run pytest tests/test_dialects_parse.py -v -k test_parser_1d_grid` passes
- [ ] Full test suite `uv run pytest tests/ -v` passes

### DIFF
--- a/ktir_cpu/parser.py
+++ b/ktir_cpu/parser.py
@@ -194,11 +194,15 @@ class KTIRParser(KTIRParserBase):
         Returns:
             (x, y, z) grid dimensions
         """
-        # Pattern: grid = [X, Y, Z] or grid = [X, Y]
-        grid_match = re.search(r'grid\s*=\s*\[(\d+),\s*(\d+)(?:,\s*(\d+))?\]', func_header)
+        # Pattern: grid = [X] or grid = [X, Y] or grid = [X, Y, Z].
+        # Missing dims default to 1.
+        grid_match = re.search(
+            r'grid\s*=\s*\[(\d+)(?:,\s*(\d+))?(?:,\s*(\d+))?\]',
+            func_header,
+        )
         if grid_match:
             x = int(grid_match.group(1))
-            y = int(grid_match.group(2))
+            y = int(grid_match.group(2)) if grid_match.group(2) else 1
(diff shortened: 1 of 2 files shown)


fabianlim:
CC @kiszk @lchu6

nwang-ibm (approved):
LGTM

PR #59, opened 2026-05-15
Generator-based scheduler: fixes cross-core communication correctness

### LINKED ISSUE #50: Real RingBackend with working cross-core reduce
# Plan: Real RingBackend with working cross-core reduce

Supersedes the "Phase 1 remainder" + "Phase 2" sections of `PLAN_core.md`.
Phase 0 (`construct_distributed_memory_view`) and the first half of Phase 1
(per-core LX routing, `RingBackend` ABC, `DirectLXBackend`) are done in
`PR_core_lx_routing.md`. This issue covers what's left to make `ktdp.reduce`
actually run end-to-end.

cc @nwang-ibm @WarningRan

## Layering: who knows what

Three layers, strictly separated. Each only talks to the one directly
below it.

```
┌─────────────────────────────────────────────────────────────────┐
│ Dialect handler (ktdp.reduce, ktdp.transfer)                    │
│   knows: how to extract MLIR operands/attributes                │
│   doesn't know: scheduling, algorithms, transport               │
│   → returns a generator from CommOps.*                          │
├─────────────────────────────────────────────────────────────────┤
│ Collective algorithm (CommOps.reduce, CommOps.transfer)         │
│   knows: the ring algorithm — N-1 rounds, who-sends-to-whom     │
│   doesn't know: when other cores run, how send/recv physically  │
│                 happen, latency                                 │
│   → calls ctx.send_to(...) and yield ('recv', src)              │
├─────────────────────────────────────────────────────────────────┤
│ Driver (GridExecutor._run_scheduler)                            │
│   knows: how to step generators, route ('recv', src) requests   │
│          against ring messages, detect deadlock                 │
│   doesn't know: what reduce is, what transfer is                │
│   → invokes ctx.send_to which dispatches to RingBackend         │
├─────────────────────────────────────────────────────────────────┤
│ Transport (RingBackend: DirectLXBackend, RingTransferBackend)   │
│   knows: how a send physically happens — direct memcpy?         │
│          hop-counted latency? multi-cast?                       │
│   doesn't know: anything about reduce / transfer / scheduling   │
└─────────────────────────────────────────────────────────────────┘
```

### CommOps vs. RingBackend — do they clash?

They look like both implement "distributed semantics," but they live on
different axes and don't compete *if the layering is enforced*:

- `CommOps.reduce` = **algorithm**. Which cores send what to whom, in
  what order. Independent of transport.
- `RingBackend` = **transport**. When a core says "send to B," what
  physically happens — direct buffer write, or hop-counted with latency,
  or multi-cast.

Same reduce algorithm runs on either backend with different latency
profiles. Same backend serves any number of collective algorithms.

**The clash exists today.** Current `CommOps.reduce` (signature:
`reduce(context, tile, core_group, reduce_fn, ring: RingNetwork)`)
reaches *past* `CoreContext` and calls `ring.send` / `ring.recv_from`
directly. This bypasses the `RingBackend` ABC entirely and pins the
reduce algorithm to one specific transport. K2 is a symptom — the
algorithm papers over empty receive buffers (`if received is not None`)
instead of blocking properly, because the transport it's wired to is
synchronous and non-blocking.

**PR-B fixes the layering.** `CommOps.reduce` loses its `ring` parameter
entirely:

```python
# Before — reaches into transport
def reduce(ctx, tile, group, fn, ring):
    ring.send(ctx.core_id, next_core, result)
    received = ring.recv_from(prev_core, ctx.core_id)  # may be None!

# After — talks only to ctx
def reduce(ctx, tile, group, fn):
    ctx.send_to(next_core, result)
    received = yield ('recv', prev_core)               # always a tile
```

After the change, `CommOps.reduce` doesn't import `RingNetwork` or
`RingBackend`. It only knows `context`. The transport choice
(`DirectLXBackend` vs. `RingTransferBackend`) is invisible to the
algorithm — that's the point.

## Why not just finish Phase 1 as written

Adding `RingTransferBackend` underneath the existing `CommOps.reduce` does
not make reduce work. `CommOps.reduce` runs a log₂N `send`/`recv` loop
*inside one core's handler invocation* — sibling cores haven't run yet, so
`recv` finds an empty buffer. K1/K2 in gap analysis describe exactly this.
The scheduler is the load-bearing fix; the latency-accurate transport
backend is layered on top.

### PR-A: Cross-core reduce example + xfail test

Pin the contract before changing transport.

- `examples/ktir/ring_reduce.mlir` — 4 cores, each holds a 1×K row in its
  own LX, `ktdp.reduce` to sum into core 0's LX.
- `tests/test_ring_reduce.py` — load known inputs, run, assert core 0's LX
  contains the sum. Mark `xfail` until PR-B.
- No code changes to the interpreter.

Size: ~1 day.

### PR-B: Generator-based scheduler (fixes K1 + K2)

The real work. Replaces `GridExecutor.execute_with_communication`'s BSP
replay with a single-pass generator-driven scheduler.

**Handler shape.** `ktdp.reduce` and `ktdp.transfer` handlers become
generators. Blocking calls suspend by yielding a request:

```python
def reduce_handler(ctx, tile, group, fn):
    for r in range(rounds):
        ctx.send_to(next_core, partial)
        msg = yield ('recv', prev_core)   # suspends here
        partial = fn(partial, msg)
    return partial
```

Sync handlers (everything else) stay sync. `_execute_operation` wraps a
sync handler's return value in a one-shot generator so the driver loop is
uniform.

**Driver.** Per-core generator + a wait-set keyed by `(src, dst)`. Loop:

1. For each runnable core, `.send(last_recv_value)`. If it yields
   `('recv', src)`, park it on `(src, self.core_id)`. If it `StopIteration`s,
   mark done.
2. After every step, drain `RingNetwork`'s pending messages: for each
   `(src, dst)` with a queued tile, if a core is parked on that key, wake
   it with the popped tile.
3. If no core is runnable and any are parked → deadlock. Raise with the
   wait-set printed (`core 3 waiting on recv from core 1; core 1 done`).
4. Stop when all cores done.

**Latency.** Each `.send(...)` step on a generator counts as one tick
against that core's clock. The final per-core latency is the max across
cores (existing model). Transfer latency is added by PR-C; for PR-B the
ring is still `DirectLXBackend`-style (zero transport cost), but reduce
now produces the correct value.

**Files touched (sketch).**

| File | Change |
|------|--------|
| `ktir_cpu/grid.py` | Replace `execute_with_communication` with `_run_scheduler`; per-core generator state; wait-set; deadlock detection |
| `ktir_cpu/ops/comm_ops.py` | `CommOps.reduce` and `CommOps.transfer` rewritten as generators yielding `('recv', src)` |
| `ktir_cpu/dialects/ktdp_ops.py` | `ktdp.reduce` / `ktdp.transfer` handlers return the generator from `CommOps.*` instead of running it inline |
| `ktir_cpu/dialects/registry.py` | `_execute_operation` adapter: wrap sync handler returns as `iter([result])` |
| `ktir_cpu/interpreter.py` | Drop `max_rounds` argument; call `_run_scheduler` |
| `tests/test_ring_reduce.py` | Remove `xfail` |
| `tests/test_ktir_cpu.py` | Un-xfail the BSP multi-round tests xfailed in PR_core_lx_routing |
| `docs/gap_analysis.md` | K1, K2 → ✅ |

**Test plan.**

- PR-A's `test_ring_reduce` passes.
- New `tests/test_scheduler.py`:
  - 2-core ping-pong (both send, both recv) completes.
  - True cycle (A waits on B, B waits on A, no sends) raises `Deadlock`
    with both cores in the printed wait-set.
  - Reduce on 8 cores (non-power-of-2 group) produces correct sum.
- Existing test suite: 613 + 7 previously-xfailed → all green.

Size: ~3-4 days.

### PR-C: RingTransferBackend + hop-count latency

Now that correctness is solid, add real transport modelling.

- `RingNetwork.hop_count(src, dst) -> int` =
  `min(|src-dst|, num_cores - |src-dst|)`.
- `RingTransferBackend(RingBackend)` — `send`/`recv` delegate to
  `RingNetwork`; latency decorator on `send_to` charges
  `hop_count(src, dst) * tile_bytes / ring_bandwidth`.
- `CoreContext.send_to(dst, tile)` and `CoreContext.recv_from_core(src)`
  — the only public ring API for handlers.
- Interpreter selects backend via existing config (default
  `RingTransferBackend`; tests can opt into `DirectLXBackend`).
- Drop `env.reduce_backend` from `PLAN_core.md` — it was a correctness
  workaround; PR-B makes it unnecessary.

**Test plan.**

- `test_ring_reduce` still passes (correctness invariant).
- New: ring-reduce latency on 4 cores ≈ `log2(4) * (1+1+1) hops *
  bytes/bw` ± rounding.
- Hop count: `(0→5, 32 cores) == 5`; `(0→27, 32 cores) == 5`;
  `(0→16, 32 cores) == 16`.

Size: ~2 days.

## Out of scope

- K3 (multi-cast load latency). No kernel needs it yet.
- Async I/O, real concurrency. The scheduler is single-threaded by
  design — generators give us suspension without threads.
- Backpressure / bounded ring buffers. Current `RingNetwork` is unbounded;
  a real model would cap per-link in-flight bytes. Defer until a kernel
  exposes the gap.

## Open questions

1. Where should the generator state live — on `CoreContext`, or in a
   side-table inside the scheduler? Side-table keeps `CoreContext` clean
   but adds a lookup. Lean side-table.
2. Should `ktdp.transfer` block until delivered, or fire-and-forget?
   Spec is silent. Fire-and-forget matches hardware DMA; the receiver
   blocks on its `recv`. Lean fire-and-forget.
3. Deterministic core scheduling order — fixed (`range(num_cores)`) or
   randomized for fuzz testing? Start fixed; add a seed knob later if
   needed.

fabianlim:
## Summary

Replaces the BSP round-replay loop with a cooperative generator scheduler that correctly handles multi-round and cyclic communication (ring reduce, bidirectional exchange).

- Each core runs as a Python generator (`CoreExecutionStack`) — the only generator-aware layer
- Blocking `recv` ops suspend via `yield RecvRequest(src)` and resume when the tile arrives
- `send_to` is fire-and-forget; senders never block, preventing deadlock in symmetric patterns
- Deadlock detected when no core can make progress

**Changes:**
- `grid.py` — `RecvRequest` dataclass, `CoreContext.send_to`, `CoreExecutionStack`, new single-pass `execute_with_communication`
- `comm_ops.py` — `CommOps.reduce`/`transfer` rewritten as generators
- `dialects/registry.py` — `ring_backend` removed from `ExecutionEnv`
- `interpreter.py` — `_execute_op` rename, wired to new scheduler
- `tests/test_grid_scheduler.py` — scheduler harness + 8 correctness tests (single send/recv, bidirectional, single/double `scf.for`, deadlock flat/loop, 4-core ring)
- `tests/test_examples.py` — `TestRingReduceExecution` xfail class (pending parser support for new `ktdp.reduce` attributes)
- `docs/cross_core_scheduling.md` — architecture, control flow, generator design, `RingBackend` integration path
- `docs/gap_analysis.md` — K1 and K2 marked resolved

Closes #50 (partial — PR-C for `RingTransferBackend` + hop-count latency is follow-on).


fabianlim:
@nwang-ibm @WarningRan FYI — tagging you for review.

fabianlim:
## Stack TOC

This is the bottom of a 3-PR stack refactoring how cores communicate during execution. Each PR has a single, narrow goal; together they replace ad-hoc cross-core wiring with a documented, pluggable protocol.


image


1. **#59 — A correct multi-core execution model.** Replaces the previous BSP-replay approximation with a coroutine scheduler that drives each core to completion, parking on blocking receives and resuming on message arrival. Adds explicit deadlock detection. This is the foundation; the next two PRs are only meaningful on top of it.

2. **#62 — Clean up the previous comm path.** Deletes the dead message-buffer infrastructure that was bypassed by #59. Pure removal — no behavior change — but it clears space so #63's abstractions don't have to coexist with vestigial parallel paths.

3. **#63 — Make the comm primitives pluggable and the lifecycle explicit.** Reduction and remote-memory access become backend abstractions, so future variants (different reduction algorithms, ring-hop transfers) slot in without touching the scheduler. Per-run scheduler bindings on each core gain a real attach/detach API instead of an implicit field mutation.

Each PR targets `main` because GitHub doesn't support cross-fork stacked PRs cleanly; diff against the parent branch on each PR shows only that PR's changes.

cc @nwang-ibm @WarningRan

ktir_cpu/grid.py
```diff
@@ -322,27 +419,68 @@ def get_cores_in_group(self, grid_coords: Tuple[int, int, int]) -> List[int]:
...
+            _advance(core_id, tile)
+            return True
+
+        for core in self.cores:
+            core._send_fn = lambda dst, tile, _src=core.core_id: _enqueue(_src, dst, tile)
+            stacks[core.core_id] = CoreExecutionStack(core, operations, input_ptrs, execute_op)
+            _advance(core.core_id)
+
+        while stacks:
+            if not any(_try_deliver(c) for c in list(stacks)):
```
kiszk: More lightweight approach

```suggestion
            if not any(_try_deliver(c) for c in tuple(stacks)):
```

ktir_cpu/grid.py
```diff
@@ -234,6 +264,73 @@ def untrack_lx(self, name: str):
...
+        """Step the generator.  Returns the final value when done."""
+        self.waiting_on = None
+        try:
+            request = self._gen.send(send_val) if send_val is not None else next(self._gen)
+            if not isinstance(request, RecvRequest):
+                raise TypeError(f"Comm op yielded unexpected type {type(request)!r}; expected RecvRequest")
+            self.waiting_on = request.src
+            return None
+        except StopIteration as e:
+            return e.value
```
kiszk, line 328: For ease of debugging

```suggestion
            return e.value
        except Exception as e:
            raise RuntimeError(
                f"Core {self.core.core_id}: Unexpected exception during execution"
            ) from e
```

ktir_cpu/ops/comm_ops.py
```diff
@@ -179,84 +179,99 @@ def transfer(
         context: CoreContext,
         tile: Tile,
         dst_cores: List[int],
-        ring_backend: RingBackend,
-    ):
-        """Transfer a tile to destination cores via the ring backend."""
+    ) -> Generator:
```
kiszk, line 182:
```suggestion
    ) -> Tile:
```

tests/test_grid_scheduler.py
```diff
@@ -0,0 +1,428 @@
...
+# spec above and add it to the parametrize list with a readable id.
+# ---------------------------------------------------------------------------
+
+@pytest.mark.parametrize(
+    "spec",
+    [
+        pytest.param(SPEC_RING_REDUCE_2_CORES, id="2_cores"),
+        pytest.param(SPEC_RING_REDUCE_4_CORES, id="4_cores"),
+        pytest.param(SPEC_RING_REDUCE_4X4_ROWS, id="4x4_rows"),
+        pytest.param(SPEC_RING_REDUCE_4X4_COLS, id="4x4_cols"),
```
kiszk: How about using consistent naming?

```suggestion
        pytest.param(SPEC_RING_REDUCE_2X1X1_COLS, id="4x4_cols"),
        pytest.param(SPEC_RING_REDUCE_4X1X1_COLS, id="4x4_cols"),
        pytest.param(SPEC_RING_REDUCE_4X4X1_COLS, id="4x4_cols"),
        pytest.param(SPEC_RING_REDUCE_4X4X1_COLS, id="4x4_cols"),
```

ktir_cpu/grid.py
```diff
@@ -322,27 +419,68 @@ def get_cores_in_group(self, grid_coords: Tuple[int, int, int]) -> List[int]:
...
-
-        Returns:
-            List of results from each core
+    def execute_with_communication(
+        self,
+        operations: List["Operation"],
+        input_ptrs: Dict[str, Any],
+        execute_op: Callable[["Operation", "CoreContext"], Any],
+    ) -> List[Any]:
+        """Drive all cores to completion via the generator scheduler.
```
kiszk, line 428: How about adding validations against inputs

```suggestion
        """Drive all cores to completion via the generator scheduler.

        if not operations:
            return [None] * self.num_cores

        if not callable(execute_op):
            raise ValueError(
                f"execute_op must be callable, got {type(execute_op)!r}"
            )

        if input_ptrs is None:
            input_ptrs = {}
```

ktir_cpu/grid.py
fabianlim: Apologies for the bare `except StopIteration` — you're right that unexpected exceptions silently lose the core ID context here. The wrapping belongs in `_advance()` in #63, which is the single call site that drives `resume()` for all cores:

```python
def _advance(core_id: int, send_val: Any = None) -> None:
    stack = stacks[core_id]
    result = stack.resume(send_val)
```

We'll add the `except Exception` wrapper with the core ID there, so every path through the scheduler gets consistent error context.

ktir_cpu/grid.py
fabianlim: Fair point — sorry for the missing guards. In #63 the signature becomes explicit (`operations`, `input_ptrs`, `execute_op`, `transfer_backend`) so the right place to add them is there:

```python
def execute_with_communication(
    self,
    operations: List["Operation"],
    input_ptrs: Dict[str, Any],
    execute_op: Callable[["Operation", "CoreContext"], Any],
    transfer_backend: Optional["TransferBackend"] = None,
) -> List[Any]:
```

We'll add the callable check on `execute_op` and the empty-list guard on `operations` at the top of that method.

ktir_cpu/ops/comm_ops.py
fabianlim: Sorry for the stray space. In #63 `CommOps.transfer` is removed entirely as part of the `RingBackend` → `TransferBackend` refactor, and `CommOps.reduce` gets a cleaner signature:

```python
def reduce(
    context: CoreContext,
    tile: Tile,
    core_group: List[int],
    backend: ReduceBackend,
) -> Generator:
```

We'll make sure the annotation style is clean throughout.

ktir_cpu/grid.py
fabianlim: Good call — applied, using `tuple(stacks)` as suggested.

tests/test_grid_scheduler.py
fabianlim: Good point on consistency — renamed all specs and test IDs to encode the full grid shape:

```python
pytest.param(SPEC_RING_REDUCE_2X1X1, id="2x1x1"),
pytest.param(SPEC_RING_REDUCE_4X1X1, id="4x1x1"),
pytest.param(SPEC_RING_REDUCE_4X4X1_ROWS, id="4x4x1_rows"),
pytest.param(SPEC_RING_REDUCE_4X4X1_COLS, id="4x4x1_cols"),
```

ktir_cpu/grid.py
fabianlim: Addressed in #63 — exception wrapping with core ID context added to `_advance()`: https://github.com/torch-spyre/ktir-cpu/pull/63

ktir_cpu/grid.py
fabianlim: Addressed in #63 — callable check on `execute_op` and empty-list guard on `operations` added at the top of `execute_with_communication`: https://github.com/torch-spyre/ktir-cpu/pull/63

ktir_cpu/ops/comm_ops.py
fabianlim: Addressed in #63 — `CommOps.transfer` is removed as part of the `RingBackend` → `TransferBackend` refactor, and annotation style is clean throughout: https://github.com/torch-spyre/ktir-cpu/pull/63

kiszk (approved):

PR #62, opened 2026-05-17
[Refactor] Delete RingNetwork and dead RingBackend send/recv methods

fabianlim:
## Summary

- Delete `RingNetwork`, `RingBackend.send/.recv`, and `DirectLXBackend.send/.recv` — none of them are on the live scheduler path. Cross-core comm goes through `CoreContext.send_to` + `RecvRequest`, owned by `GridExecutor`'s message queue.
- Slim `RingBackend` and `DirectLXBackend` to remote-LX-peek (`get_lx`) only, with refreshed docstrings.
- Drop the now-dead `RingNetwork` allocation from `interpreter._prepare_execution`, the `RingNetwork` export from `ktir_cpu.ops`, and `TestRingNetwork` from `tests/test_ops.py`.
- Update `docs/cross_core_scheduling.md` to remove the "Marked for deletion" section and reflect the slimmer `RingBackend`.

> **Stacked on #59** — this branch is built on top of `grid-network`. Will rebase onto `main` once #59 merges. Review the diff against `grid-network` for just the deletion changes.

## Test plan

- [x] `uv run pytest tests/` — 813 passed, 7 skipped, 6 xfailed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #63, opened 2026-05-17
[Refactor] ReduceBackend / TransferBackend abstractions; attach_scheduler lifecycle

fabianlim:
## Summary

- Introduce `ReduceBackend` ABC and move ring-reduction into `RingReduceBackend`. `CommOps.reduce` becomes a passthrough.
- Add `@register_reduce_backend(op_name, BackendCls)` / `get_reduce_backend(op_name)` so dialect handlers declare their backend at the decorator and resolve it explicitly — same registry pattern as parsers/handlers, no introspection.
- Rename `RingBackend` → `TransferBackend`, `DirectLXBackend` → `InstantTransferBackend` (method `get_lx` → `run`). The names now describe what they do, not where they notionally live.
- Replace the implicit `core._send_fn = ...` mutation with explicit `CoreContext.attach_scheduler(send_fn, transfer_fn)` / `detach_scheduler()` lifecycle. Both injected functions live together and are bound for the duration of one run.
- `transfer_backend` is supplied as a direct parameter to `GridExecutor.execute_with_communication` (not via `ExecutionEnv` — `ExecutionEnv` is for handler-visible services, and no handler reads the backend). Scheduler curries it into each core's `transfer_fn`.
- Delete `ktdp.transfer` dialect handler (no callers remained).
- Three new deadlock tests in `test_grid_scheduler.py` — each monkey-patches `RingReduceBackend.run` with a deliberately broken protocol (recv-only, wrong destination, extra recv) and asserts the scheduler raises `RuntimeError("Deadlock detected: ...")`.

> **Stacked on #62, which is stacked on #59.** Opens against `main` because GitHub doesn't support cross-fork stacked PRs cleanly. Will rebase once #59 and #62 merge. See the table of contents on #59.

## Test plan

- [x] `uv run pytest tests/` — 813 passed, 7 skipped, 6 xfailed.
- [x] `tests/test_grid_scheduler.py` — 4 ring-reduce variants + 3 deadlock variants (all pass).

### DIFF
--- a/ktir_cpu/ops/comm_ops.py
+++ b/ktir_cpu/ops/comm_ops.py
@@ -20,7 +20,7 @@ dialect handlers in ``ktir_cpu.dialects``.
 """
 
 from abc import ABC, abstractmethod
-from typing import Callable, Generator, List
+from typing import Callable, Dict, Generator, List, Type
 from ..ir_types import Tile
 from ..grid import CoreContext, RecvRequest
 from ..memory import LXScratchpad, SpyreMemoryHierarchy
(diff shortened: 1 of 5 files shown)


fabianlim:
cc @nwang-ibm @WarningRan — stacked on #62 (which stacks on #59). See the stack TOC on #59 for review order.

ktir_cpu/ops/comm_ops.py
```diff
@@ -19,129 +19,153 @@
...
-            round 3:
-              core 0:  recv 2; acc = 8+2 = 10
-              core 1:  recv 3; acc = 7+3 = 10
-              core 2:  recv 4; acc = 6+4 = 10
-              core 3:  recv 1; acc = 9+1 = 10
-
-        Cores not in *core_group* return their input tile unchanged
-        without communicating.
-        """
+    ) -> Generator[RecvRequest, Tile, Tile]:
```
kiszk: Do you want to add `Tile` since `RingReduceBackend` returns `Tile`? Or, do you want to change `RingReduceBackend.run()` to return `Generator[]`?
```suggestion
    ) -> Union[Tile, Generator[RecvRequest, Tile, Tile]]:
```

tests/test_grid_scheduler.py
```diff
@@ -426,3 +433,92 @@ def run_spec(spec: dict, execute_fn) -> GridExecutor:
...
+    yield RecvRequest(src=prev)
+    return tile
+
+
+def _broken_extra_recv(self, ctx, tile, group):
+    """Run N-1 ring rounds correctly, then dangle one extra recv. The
+    final recv has no matching send → deadlock after the algorithm
+    has otherwise made progress.
+    """
+    if ctx.core_id not in group:
```
kiszk, line 493: Is it better to refactor three initializations by using common setup code like?
```
def _setup_ring_context(ctx, tile, group):
    if ctx.core_id not in group:
        return None, tile
    n = len(group)
    my_idx = group.index(ctx.core_id)
    next_core = group[(my_idx + 1) % n]
    prev_core = group[(my_idx - 1) % n]
    return (n, my_idx, next_core, prev_core), tile

def _broken_extra_recv(self, ctx, tile, group):
    setup, tile = _setup_ring_context(ctx, tile, group)
    if setup is None:
        return tile
    n, my_idx, next_core, prev_core = setup
    ...
```

ktir_cpu/ops/comm_ops.py
fabianlim: Good catch — fixed with `Union[Tile, Generator[RecvRequest, Tile, Tile]]`. The early return for non-participating cores makes `run` a plain function in that path, so `Generator` alone was wrong.

tests/test_grid_scheduler.py
fabianlim: Counted ~8–9 lines saved net (helper adds ~5, removes 12). Not enough to justify the indirection for what are intentionally standalone readable stubs — keeping them as-is.

kiszk (approved):

PR #64, opened 2026-05-18
[Test] test_paged_tensor_indirect_access: derive seed addresses from IR

### LINKED ISSUE #57: [Bug] paged-tensor-copy.mlir: Y physical footprint overlaps Idx
## Background

`examples/rfc/paged-tensor-copy.mlir` has a latent fixture-layout bug: Y's physical footprint extends into Idx's address range.

## Problem

Address constants in the fixture:

- `Y_start_address = 10000000`
- `Idx_start_address = 20000000`

Y's footprint = `Nb × Ntkv × Nhkv × Ndkv × sizeof(f16)` = `4 × 2048 × 8 × 128 × 2` = 16,777,216 bytes ≈ 16 MiB, which extends past `20000000` and physically overlaps Idx (overlap region: `[20000000, 26777216)`, ≈ 6.46 MiB).

The current zero-seed test masks the bug (zeros in either tensor are benign for the gather output). A non-zero seed would let the layout corrupt the gather indices, producing wrong test results.

## Fix

Relayout the fixture's address constants so Y's footprint does not extend into Idx's range. Test-fixture-only change; no production code changes.

## Context

- Identified during review of #49.
- Latent — current zero-seed tests pass; would surface with non-zero seeds.

WarningRan:
Close #57.

Refactor `test_paged_tensor_indirect_access` to read seed addresses from the parsed IR instead of hardcoding `20000000` / `30000000`. The test now tracks whatever `arith.constant` values appear in `paged-tensor-copy.mlir`.

### DIFF
--- a/tests/test_spec_gaps.py
+++ b/tests/test_spec_gaps.py
@@ -104,11 +104,13 @@ def test_paged_tensor_indirect_access(path, func_name, entry):
     def _prepare_and_seed(grid_shape):
         _orig(grid_shape)
         hbm = interp.memory.hbm
-        # Idx tensor (4×32 i32): all zeros so every indirect page ID resolves to 0.
-        hbm.write(addrs["Idx_start_address"], np.zeros(4 * 32, dtype=np.int32))
-        # X tensor: only page 0 needs data (65 536 f16 elements = one page).
+        # Idx tensor (Nb × Ntkv/Ptkv i32): all zeros so every indirect page ID resolves to 0.
+        idx_n_elements = addrs["Nb"] * (addrs["Ntkv"] // addrs["Ptkv"])
+        hbm.write(addrs["Idx_start_address"], np.zeros(idx_n_elements, dtype=np.int32))
+        # X tensor: only page 0 needs data (one page = Nhkv * Ptkv * Ndkv f16 elements).
         # Because Idx is all zeros every gather access lands on page 0.
-        hbm.write(addrs["X_start_address"], np.zeros(65536, dtype=np.float16))
+        x_page_n_elements = addrs["Nhkv"] * addrs["Ptkv"] * addrs["Ndkv"]
+        hbm.write(addrs["X_start_address"], np.zeros(x_page_n_elements, dtype=np.float16))
     interp._prepare_execution = _prepare_and_seed
     interp.execute_function(func_name)
 
--- a/tests/test_spec_gaps.py
+++ b/tests/test_spec_gaps.py
@@ -90,8 +90,10 @@ def test_paged_tensor_indirect_access(path, func_name, entry):
     """
     interp = KTIRInterpreter()
     interp.load(path)
-    # Read seed addresses from the parsed IR so the test tracks the .mlir if
-    # its arith.constant values change.
+    # Build a name→value map of every arith.constant in the function so we can
+    # look up addresses by their SSA name (e.g. %Idx_start_address) instead of
+    # hardcoding the literal integers.  If the .mlir changes the address values,
+    # the test automatically picks them up.
     func = interp.module.get_function(func_name)
     addrs = {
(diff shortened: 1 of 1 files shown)


WarningRan:
cc @fabianlim for your review

tests/test_spec_gaps.py
```diff
@@ -90,15 +90,23 @@ def test_paged_tensor_indirect_access(path, func_name, entry):
     """
     interp = KTIRInterpreter()
     interp.load(path)
+    # Read seed addresses from the parsed IR so the test tracks the .mlir if
```
fabianlim: Suggestion: expand this comment to explain what the dict contains and why it avoids hardcoding:

```python
# Build a name→value map of every arith.constant in the function so we can
# look up addresses by their SSA name (e.g. %Idx_start_address) instead of
# hardcoding the literal integers.  If the .mlir changes the address values,
# the test automatically picks them up.
```

tests/test_spec_gaps.py
```diff
@@ -90,15 +90,23 @@ def test_paged_tensor_indirect_access(path, func_name, entry):
...
+        op.result.lstrip("%"): op.attributes["value"]
+        for op in func.operations
+        if op.op_type == "arith.constant" and op.result
+    }
     _orig = interp._prepare_execution
     def _prepare_and_seed(grid_shape):
         _orig(grid_shape)
         hbm = interp.memory.hbm
-        # Idx tensor at 20000000: shape 4x32 i32, all zeros → every page ID = 0
-        hbm.write(20000000, np.zeros(4 * 32, dtype=np.int32))
```
fabianlim, line 98: Suggestion: separate the shape/dtype fact from the zeroing rationale:

```python
# Idx tensor (4×32 i32): all zeros so every indirect page ID resolves to 0.
hbm.write(addrs["Idx_start_address"], np.zeros(4 * 32, dtype=np.int32))
# X tensor: only page 0 needs data (65 536 f16 elements = one page).
# Because Idx is all zeros every gather access lands on page 0.
hbm.write(addrs["X_start_address"], np.zeros(65536, dtype=np.float16))
```

WarningRan: applied verbatim, thanks

tests/test_spec_gaps.py
WarningRan: applied verbatim, thanks

tests/test_spec_gaps.py
```diff
@@ -90,15 +90,25 @@ def test_paged_tensor_indirect_access(path, func_name, entry):
...
-        # X tensor at 30000000: seed page 0 only (65536 f16 elements)
-        # All accesses land on page 0 because Idx is all zeros.
-        hbm.write(30000000, np.zeros(65536, dtype=np.float16))
+        # Idx tensor (4×32 i32): all zeros so every indirect page ID resolves to 0.
```
fabianlim: actually looking at this again. im wondering now if its better to derive the the data sizes (`4 * 32` and  `65536`) from the sizes also, since even though the size hardcodes are removed, the data sizes are hardcoded.

WarningRan: Done — applied. Note both sizes derive from the same arith.constant
dict; `4*32` was pure Idx shape repetition, while `65536` was the
"one page" test-scope choice (Idx all zeros → only page 0 ever
gathered). The derived `Nhkv * Ptkv * Ndkv` makes that scope choice
visible at the call site.

WarningRan:
FYI — `test_paged_tensor_indirect_scatter` (added in #49) follows the same hardcoded-address-and-size pattern. Planning to apply the same IR-introspection treatment to it in a separate follow-up PR to keep this one focused, but happy to fold in here if you'd prefer.

fabianlim (approved):
thank you, now the hardcoding is removed from the test and its more portable

Issue #61, opened 2026-05-15
[Bug] Latency accounting gap: index tensor sticks not counted in indirect_load / indirect_store


WarningRan:
## Background

In `indirect_load` and `indirect_store`, each index element is read via a per-element `_MemAccessor.read(1, ...)` call. The HBM sticks touched by those reads are never accumulated into the `unique_sticks` field on the `Tile` returned by `indirect_load` (and used internally by `indirect_store`). The latency model only sees stick traffic for the **data** tensor — index-tensor HBM accesses are charged at flat dense bytes (`np.prod(idx.shape) * bytes_per_elem(idx.dtype)`) rather than at stick granularity.

For a 64×64 IAT with two index tensors, that's up to 4096 × 2 = 8192 index reads whose sticks go unaccounted, which can undercount latency estimates for scattered access patterns. Worst case (each idx element landing on its own stick) the charge should be `n_idx_reads × STICK_BYTES`, not `n_idx_elements × bytes_per_elem(idx.dtype)`.

The direct case (`MemoryOps.load`) handles this correctly via `_flat_memory_offsets`. The indirect paths need to do the same — accumulate the index-read sticks into the final `unique_sticks` on the returned tile, or thread them through as a separate accounting field.

## Context: surfaced during PR #49

This is a **pre-existing gap in `indirect_load` that `indirect_store` (added in #49) inherits** — both walk the same per-element index resolution loop:

​```python
# ktir_cpu/ops/memory_ops.py — inside indirect_load / indirect_store
for sub in iat.dim_subscripts:
    if sub["kind"] == "indirect":
        idx_view = iat.index_views[sub["index_view_idx"]]
        addr = idx_view.byte_address + offset * bpe
        raw = _MemAccessor(
            context, idx_view.memory_space, addr, idx_view.lx_core_id,
        ).read(1, idx_view.dtype)
​```

Flagged by @fabianlim on PR #49 as **non-blocking** but worth tracking explicitly. Original review comment: https://github.com/torch-spyre/ktir-cpu/pull/49#pullrequestreview-4301684659.

## Proposed direction

Two thread-through options (both raised in the original review comment):

1. **Accumulate into `Tile.unique_sticks`** — fold index-read sticks into the same field. Simplest, but conflates data-side and index-side stick semantics; `Tile.coalescing_efficiency` (currently `data.nbytes / (unique_sticks * STICK_BYTES)`) would need redefinition.
[...]

PR #65, opened 2026-05-20
[Fix] idx tensor stick accounting in indirect_load / indirect_store

WarningRan:
# [Fix] Make HBM stick accounting consistent across load and store

Addresses #61.

## The fix

Stores never returned the unique-stick count they computed. The latency model fell back to either source-tile `nbytes` (which undercounts non-stick-aligned scatter) or a no-IO mirror (which only existed because `indirect_store` discarded its own count).

The fix is structural: every store entry point — `MemoryOps.store`, `indirect_store`, `distributed_store` — returns an `int` unique-stick count, the `ktdp.store` handler propagates it as the op result, and `LatencyTracker._data_size` charges `result * STICK_BYTES`. KTIR-level semantics are unchanged (stores still produce no IR result; the `int` is an interpreter-only convention). All `v.data.nbytes` fallbacks are deleted on both load and store paths; any Tile or IAT operand whose unique-stick field is unpopulated raises `RuntimeError` so handler regressions surface immediately.

LX targets return `0`, not `None`. "Stick count is ill-defined for LX" stays a helper-level concept (`_MemAccessor.count_sticks` still returns `None`); the public-op return is "this op generated zero HBM sticks" — same type-shape as HBM, no `None` in the sideband channel.

`_MemAccessor.read_scattered` was rewritten to batch unique addresses into contiguous runs — one `sim.read` per run; runs may cross stick boundaries since DMA descriptors are `(start, length)`. The `_idx_unique_sticks_no_reads` helper is deleted; it existed only because `indirect_store` discarded its count.

## Where the change lives

- `ktir_cpu/ops/memory_ops.py` — `read_scattered` rewritten to contiguous-run batching; `MemoryOps.{store,indirect_store,distributed_store}` signatures `-> int`; `_idx_unique_sticks_no_reads` deleted; `_resolve_idx_reads` skips views whose enumerated address list is empty so a degenerate (zero-extent) variable space stops short-circuiting through a misleading post-loop invariant message.
- `ktir_cpu/dialects/ktdp_ops.py` — `ktdp.store` handler returns the `int` from each branch instead of `return None`.
- `ktir_cpu/latency.py` — `_data_size` dispatches on `isinstance(result, int)` first; `v.data.nbytes` fallback removed; strict raises pin the handler contract.
- `tests/test_latency.py` and `tests/test_latency_modeling.py` — see Tests.

## Concerns we worked through

**`read_scattered` cross-allocation invariant.** All addresses must lie within a single HBM allocation; cross-allocation calls would silently zero-fill because `HBMSimulator._read_flat` reads one allocation only. The invariant is declared in the docstring; hard-guarding requires the simulator to expose allocation extent and is tracked in #67.

**Run boundaries don't split at stick edges.** `read_scattered` cuts on `diff > bpe` only. Each run maps to a single DMA block transfer; `HBMSimulator._read_flat` already handles cross-stick reads within one allocation. Call count = run count, bounded by `unique_sticks` (best case = 1 for fully dense). Pinned by `test_read_scattered_run_crosses_stick_boundary_one_call`.

**LX = 0 sticks at the public-op boundary, not None.** Every store handler returns `int`, so the strict raises in `_data_size` only fire on genuine handler bugs, never on legitimate LX paths. The "ill-defined for LX" semantics stay one layer down where they're useful for LX-vs-HBM dispatch.

**`distributed_store` returns `int` even though the original review didn't request it.** All three store variants share the same dialect-handler branch and the same `_data_size` consumer; one-off behavior on the distributed path would force the handler to dispatch on store type before deciding whether to return. Symmetric `int` keeps the contract uniform.

## Open question

- **Paged-attention E2E fixture.** We owe a paged-attention E2E (same shape family as RFC §5 Example 2, `(4, 8, 2048, 128)`). Reuse `examples/rfc/indirect-access-copy.mlir` if representative enough, or add a new `.mlir`. You know this repo's tolerance for new fixtures better than I do.

## Tests

In `tests/test_latency.py`:

- **`read_scattered` contract** — `test_read_scattered_*` (8 parametrize cases: dense/scattered/repeated/cross-stick/subset/LX) lock call count = run count, run boundaries cut on `diff > bpe`, intra-byte propagated, values reassembled in caller order.
- **Int-sideband semantics** — `test_data_size_int_sideband_charges_stick_bytes`, `test_data_size_int_sideband_direct_store_64x64_scatter` (C3 regression pin: `100 * 128 = 12800` vs the prior `64 * 64 * 2 = 8192` logical-bytes proxy — 4608-byte undercount).
- **Strict-raise guards** — `test_data_size_rejects_{tile,iat}_operand_with_none_result`, `test_data_size_raises_when_tile_result_missing_unique_sticks`, `test_data_size_raises_when_iat_load_result_missing_index_unique_sticks`. All use `pytest.raises(RuntimeError, match=...)`.
- **LX = 0 contract** — `test_store_returns_zero_for_lx_destination`, `test_indirect_store_returns_zero_for_all_lx`, `test_distributed_store_returns_zero_for_all_lx`.
- **Load ↔ store symmetry** — `test_indirect_store_index_sticks_mirrors_load` (`store_total_sticks - load_data_sticks == load_idx_sticks` on the same IAT), plus the preserved `test_indirect_load_index_sticks_simple_gather`.
- **Zero-extent IAT skip** — `test_resolve_idx_reads_zero_extent_skips_view`.
- Preserved load-side IAT pins: `test_data_size_charges_index_unique_sticks`, `test_data_size_iat_load_skips_operand_branch_when_result_field_set[…]`.

In `tests/test_latency_modeling.py`:

- `test_work_splitting_{elementwise,transcendental}[4|8]` — expected store cost changed from `tile_bytes` to `sticks_per_op * STICK_BYTES`. The earlier parametrization was pinning the C3 bug; under the fix the correct expectation is stick-granular.

### DIFF
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -32,6 +32,7 @@ import numpy as np
 from .ir_types import AccessTile, IndirectAccessTile, MemRef, Tile, TileRef
 from .dtypes import bytes_per_elem
 from .memory import HBMSimulator
+from .ops.memory_ops import _idx_unique_sticks_no_reads
 
 
 from .dialects.registry import get_latency_category
@@ -298,11 +299,27 @@ class LatencyTracker:
                 total += result.unique_sticks * HBMSimulator.STICK_BYTES
             else:
                 total += result.data.nbytes
+            if result.index_unique_sticks is not None:
+                total += result.index_unique_sticks * HBMSimulator.STICK_BYTES
         for v in operands:
             if isinstance(v, IndirectAccessTile):
-                for iv in v.index_views:
-                    if iv.memory_space == "HBM":
-                        total += int(np.prod(iv.shape)) * bytes_per_elem(iv.dtype)
+                # Load case (result is a Tile): idx sticks already charged
+                # via result.index_unique_sticks above. The IAT-load
+                # contract requires the handler to set the field to an
+                # integer (0 for an all-LX IAT, positive for HBM); a
+                # surviving None means the handler skipped
+                # ``_resolve_idx_reads``, which would silently undercount.
+                if isinstance(result, Tile):
+                    if result.index_unique_sticks is None:
+                        raise RuntimeError(
+                            "IAT operand with Tile result must populate "
+                            "index_unique_sticks; got None. This indicates "
+                            "the op handler skipped _resolve_idx_reads."
+                        )
+                    continue
+                # Store case: no result Tile carries the count, so
+                # recompute from the IAT. Always int (0 for all-LX).
+                total += _idx_unique_sticks_no_reads(v) * HBMSimulator.STICK_BYTES
             elif isinstance(v, Tile):
                 if result is not None:
                     raise ValueError(
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -308,20 +308,17 @@ class LatencyTracker:
 
         total = 0
         if isinstance(result, Tile):
-            if result.unique_sticks is not None:
-                total += result.unique_sticks * HBMSimulator.STICK_BYTES
-            else:
-                total += result.data.nbytes
+            if result.unique_sticks is None:
+                raise RuntimeError(
+                    "Tile result on HBM path must populate unique_sticks; "
+                    "got None. Load handlers must set unique_sticks for "
+                    "stick-granular HBM accounting."
+                )
+            total += result.unique_sticks * HBMSimulator.STICK_BYTES
             if result.index_unique_sticks is not None:
                 total += result.index_unique_sticks * HBMSimulator.STICK_BYTES
         for v in operands:
             if isinstance(v, IndirectAccessTile):
-                # Only the load path reaches here — store ops returned
-                # via the int sideband above. The IAT-load contract
-                # requires the handler to set ``index_unique_sticks`` to
-                # an integer (0 for an all-LX IAT, positive for HBM); a
-                # surviving None means the handler skipped
-                # ``_resolve_idx_reads``, which would silently undercount.
                 if not isinstance(result, Tile):
                     raise RuntimeError(
                         "IAT operand without Tile result and without int "
@@ -342,12 +339,6 @@ class LatencyTracker:
                         f"_data_size: Tile in operands but result is also "
                         f"{type(result).__name__}; no ktdp op should produce both"
                     )
-                # Tile-operand-without-result reaches here only if a memory
-                # op handler returned None instead of the int from its
-                # underlying MemoryOps call. Reject rather than fall back
-                # to v.data.nbytes — the source tile's logical size is not
-                # a valid proxy for HBM traffic on a stick-addressed bus
-                # (would undercount any non-stick-aligned scatter).
                 raise RuntimeError(
                     "Tile operand with None result: store handler must "
                     "propagate MemoryOps.store's int return as op result "
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -32,7 +32,6 @@ import numpy as np
 from .ir_types import AccessTile, IndirectAccessTile, MemRef, Tile, TileRef
 from .dtypes import bytes_per_elem
 from .memory import HBMSimulator
-from .ops.memory_ops import _idx_unique_sticks_no_reads
 
 
 from .dialects.registry import get_latency_category
@@ -287,12 +286,26 @@ class LatencyTracker:
     def _data_size(result: Any, operands: List[Any]) -> int:
         """Estimate bytes transferred by a memory operation.
 
-        Counts result bytes (loads), operand Tile bytes (stores), and
-        index tensor bytes (indirect access loads and scatter stores).
-
         HBM traffic is always charged at stick granularity:
         ``unique_sticks * HBMSimulator.STICK_BYTES``.
+
+        Two carriers convey ``unique_sticks`` from the op handler:
+
+        * **Loads** stamp ``unique_sticks`` (data) and
+          ``index_unique_sticks`` (idx, when an IAT is involved) on the
+          result :class:`Tile`. ``_data_size`` reads them off the result.
+        * **Stores** have no result Tile — the dialect handler instead
+          returns the int from ``MemoryOps.store`` /
+          ``indirect_store`` / ``distributed_store`` as the op result.
+          ``_data_size`` consumes it via ``isinstance(result, int)``.
+          For an indirect store, the int already aggregates both the
+          parent destination's sticks and the idx-side sticks.
         """
+        # Store sideband: the handler propagated MemoryOps.{store,
+        # indirect_store, distributed_store}'s int return as op result.
+        if isinstance(result, int):
+            return result * HBMSimulator.STICK_BYTES
+
         total = 0
         if isinstance(result, Tile):
             if result.unique_sticks is not None:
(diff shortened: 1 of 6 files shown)


WarningRan:
@fabianlim

ktir_cpu/ops/memory_ops.py
```diff
@@ -72,12 +73,228 @@ def __init__(
...
+        ambiguous in a read context (zero-traffic vs caller bug); use
+        :meth:`count_sticks` directly for the pure query.
+        """
+        if not byte_addresses:
+            raise ValueError("read_scattered called with empty address list")
+        unique_sticks = type(self).count_sticks(self._memory_space, byte_addresses)
+        np_dtype = _to_np_dtype(dtype)
+        cache: Dict[int, Any] = {}
+        for a in set(byte_addresses):
+            if self.stick_bytes is not None:
```
fabianlim, line 152: This still issues one `_sim.read` per unique address, which is expensive when LX transfers are modelled — each call goes through `context.get_lx(lx_core_id)` and the ring backend.

Suggested approach: group `byte_addresses` into contiguous runs first, then issue one `sim.read` per run. Each run maps to a single DMA block transfer, which is both faster and physically more accurate.

Two examples:

**Dense access** — 16 i32 elements in a 4×4 IDX tensor, all contiguous:
```
addresses: [0, 4, 8, ..., 60]   (one contiguous run)
→ sim.read(base=0, n=16)        (one call)
→ pick elements by offset
```

**Scattered access** — 4 elements spanning two sticks:
```
addresses: [0, 4, 128, 132]
→ run 1: sim.read(base=0,   n=2)   → values [a, b]
→ run 2: sim.read(base=128, n=2)   → values [c, d]
→ assemble in caller order
```

Number of `sim.read` calls = number of contiguous runs ≤ `unique_sticks`. Best case (dense) = 1 call; worst case (fully scattered) = `unique_sticks` calls — bounded by sticks, not by number of points.

ktir_cpu/ops/memory_ops.py
```diff
@@ -687,56 +869,18 @@ def indirect_store(
...
-                    # survives python -O.
-                    if raw_idx < 0:
-                        raise IndexError(
-                            f"indirect index {raw_idx} from {idx_view} is negative"
-                        )
-                    coord.append(raw_idx)
-            coords.append(tuple(coord))
+        # Resolve every idx-tensor read up front (mirror of indirect_load).
+        # Stick count is discarded here — ``_data_size`` recomputes it from
+        # the IAT for store ops because there is no result Tile to carry
```
fabianlim: This discard + `_idx_unique_sticks_no_reads` recompute can be eliminated. A few options were considered:

1. **Return a carrier object** (`StoreResult(unique_sticks, index_unique_sticks)`) — works but adds a new type just to shuttle two ints out.
2. **Return an int directly** — `indirect_store` returns `idx_unique_sticks`, `store` returns `unique_sticks` (already computed by `_flat_memory_offsets`, currently discarded). The dialect handler propagates it as the op result, and `_data_size` handles it with a simple `isinstance(result, int)` branch.

Option 2 is the cleanest: no new type, no double enumeration, `_idx_unique_sticks_no_reads` goes away entirely. The same fix also closes the direct-store gap where `_data_size` currently charges `v.data.nbytes` instead of stick-granular bytes.

ktir_cpu/latency.py
```diff
@@ -298,11 +299,27 @@ def _data_size(result: Any, operands: List[Any]) -> int:
...
+                            "IAT operand with Tile result must populate "
+                            "index_unique_sticks; got None. This indicates "
+                            "the op handler skipped _resolve_idx_reads."
+                        )
+                    continue
+                # Store case: no result Tile carries the count, so
+                # recompute from the IAT. Always int (0 for all-LX).
+                total += _idx_unique_sticks_no_reads(v) * HBMSimulator.STICK_BYTES
             elif isinstance(v, Tile):
                 if result is not None:
```
fabianlim, line 337: This branch handles the direct store case (`ktdp.store` with a non-indirect access tile). It charges `v.data.nbytes` — raw tensor bytes — which is wrong for HBM.

For a 64×64 f16 tile scattered across 100 sticks the actual HBM traffic is `100 * 128 = 12800` bytes, but this charges `64*64*2 = 8192` bytes. Undercount or overcount depending on scatter pattern.

The load path handles this correctly: `MemoryOps.load` stamps `unique_sticks` on the returned `Tile` and `_data_size` reads it. The store path breaks because the handler returns `None` so there is no object to carry the count.

Fix: have `MemoryOps.store` return `unique_sticks` (it already computes it via `_flat_memory_offsets` and discards it with `offsets, _ = ...`). The dialect handler propagates it, and `_data_size` handles it with `isinstance(result, int)`. This is the same fix proposed for `indirect_store` above — both store paths become consistent with the load path.

ktir_cpu/ops/memory_ops.py
WarningRan: `read_scattered` now sorts unique addresses, walks once to merge any pair with `diff == bpe` into a contiguous run, and issues one `sim.read(start, n=run_len, dtype, intra_byte)` per run. Calls = run count: 1 for fully dense, `unique_sticks` worst case. Runs may span stick boundaries -- a DMA descriptor is (start, length), and `_read_flat` already handles cross-stick reads within one allocation. Locked by `test_read_scattered_*` (8 cases: dense_within_stick / mid_stick_run_intra_byte_propagated / scattered_two_runs_of_two / two_isolated_points / repeated address / subset / cross-stick / lx_returns_none).

ktir_cpu/ops/memory_ops.py
WarningRan: You're right -- the "forced by architecture" framing was wrong. `MemoryOps.indirect_store` now returns `data_sticks + idx_unique_sticks` as an int, the dialect handler propagates it as the op result, and `_data_size` consumes via `isinstance(result, int)`. The `_idx_unique_sticks_no_reads` helper is gone -- it existed only because `indirect_store` previously discarded the count it had already computed. Tests: `test_data_size_int_sideband_charges_stick_bytes`, `test_indirect_store_index_sticks_mirrors_load`, `test_indirect_store_returns_zero_for_all_lx`.

ktir_cpu/latency.py
WarningRan: `MemoryOps.store` (and `distributed_store` for symmetry) now also return int unique-sticks, propagated through the same sideband as indirect_store. `_data_size`'s `v.data.nbytes` fallback is gone; a Tile operand with a None result now raises a strict `RuntimeError` so handler regressions surface immediately instead of silently undercounting. Pinned by `test_data_size_int_sideband_direct_store_64x64_scatter` (the 64x64 f16 / 100 sticks example: charges 100*128 = 12800, asserts that differs from the logical 8192). LX targets return 0 (no stick concept; zero HBM sticks) so the new strict guards don't fire on legitimate LX paths -- covered by `test_store_returns_zero_for_lx_destination`, `test_distributed_store_returns_zero_for_all_lx`.

ktir_cpu/ops/memory_ops.py
```diff
@@ -117,13 +126,28 @@ def read_scattered(
             raise ValueError("read_scattered called with empty address list")
         unique_sticks = type(self).count_sticks(self._memory_space, byte_addresses)
         np_dtype = _to_np_dtype(dtype)
+        bpe = _bytes_per_elem(dtype)
+
+        sorted_unique = sorted(set(byte_addresses))
+        runs: List[List[int]] = [[sorted_unique[0]]]
```
fabianlim, line 141: is there a danger of having nothing in the `sorted_unique` list? if so we should guard, and if there is something above that already guards this, we should have a comment

fabianlim:
Reviewed carefully — 871/871 tests pass. No blockers. A few items worth addressing:

### Correctness

**Silent zero-fill if scattered addresses cross an allocation boundary** (`ktir_cpu/ops/memory_ops.py:131-137`).

The merge step joins addresses whose sorted gap equals `bpe` into one batched read. Concretely, with `bpe=4`:

- input addresses `[0x1000, 0x1004, 0x1008]` → one run, one `sim.read(0x1000, 3, ...)`
- input addresses `[0x1000, 0x100C]` (gap 12 ≠ bpe) → two runs, two reads

The hazard: if two distinct HBM allocations happen to sit exactly `bpe` apart (e.g. allocation A ends at `0x1100`, allocation B begins at `0x1104`, both i32), an input like `[0x10FC, 0x1104]` merges into one run. That run is then served by a single `_read_flat` call which finds *one* containing allocation and zero-pads anything past its end (`memory.py:159-165`). The element from B silently reads as zero — wrong values, no error.

Today's only caller (`_resolve_idx_reads`) issues one accessor per `idx_view`, and every address it passes is `idx_view.byte_address + offset * bpe` — all within that one view's allocation. So the bug isn't reachable today. But `read_scattered` reads as a public-ish helper with no precondition stated. Two ways to close it:

1. (cheap) Document the invariant: "all addresses must lie within a single allocation."
2. (safer) Cap each run on either `diff > bpe` *or* `start + len(run)*bpe > alloc_end` — would require a way to query allocation extent, which doesn't currently exist on the simulator.

I'd take (1) for this PR and file a follow-up if (2) ever becomes needed.

**LX-load `result.data.nbytes` fallback is dead code** (`ktir_cpu/latency.py:313-314`).

`_data_size` is only reached for HBM ops — LX ops short-circuit at line 202-203. And on the HBM path, `MemoryOps.load` always populates `unique_sticks` (lines 469-475). So `else: total += result.data.nbytes` never runs. Either delete it, or replace with a strict raise to match the new tile-operand-with-None-result contract right below it. Leaving it inconsistent is the worst of both.

**Empty IAT enumeration would raise with a misleading message** (`ktir_cpu/ops/memory_ops.py:125-126`).

If `iat.variables_space_set.enumerate(iat.shape)` ever returns `[]` (e.g. a zero-extent dim), `_resolve_idx_reads` calls `read_scattered([], …)` which raises `ValueError("read_scattered called with empty address list")` — a caller-error message for a legitimate degenerate case. Probably not reachable from well-formed IR. Either gate `if not addrs: continue` in `_resolve_idx_reads`, or accept empty in `read_scattered` and return `(empty_array, 0)`.

### Doc/comment hygiene

A few docstrings/comments read like commit messages — worth trimming:

- `memory_ops.py:118-119` — "Full migration to a tuple-returning `read()` is a follow-up" belongs in an issue, not a docstring.
- `memory_ops.py:184-188` — "Performance note … paged-attention scale" — the hoist's purpose is self-evident; one sentence.
- `latency.py:332-339` and `latency.py:354-359` — multi-line block comments above raises whose error messages already say the same thing.
- `tests/test_latency.py` — recurring "C3 regression" / "pre-fix bug" / "pre-sideband" framing (e.g. ~lines 961, 1179, 1190, 1220, 1227, 1238). Tests should pin invariants, not narrate fix history.

### Things I confirmed are correct

- Int sideband doesn't escape into IR/env: `interpreter.py:210` gates `set_value` on `op.result` truthy, and `ktdp.store` has no result.
- Caller order preservation in `read_scattered`: `cache[a] = block[i]` keyed by run address, then `np.fromiter` rehydrates in input order — works for non-monotonic and repeated addresses.
- Stick-count fast-path `ceil((bp+n*bpe)/sb) - floor(bp/sb)` is correct for any base_ptr alignment, not just stick-aligned.
- `_resolve_idx_reads` ↔ `_build_indirect_coords` ordering: both iterate pt-major dim-minor with the same `kind=="indirect"` filter, so per-view consumption order matches address insertion order. Shared-view case works.
- Spec: no `ktdp` op syntax/operand/attribute changes; indirect coord assembly preserves "not combined multiplicatively"; `construct_memory_view` still non-allocating.

cc @kiszk @lchu6 @nwang-ibm

ktir_cpu/ops/memory_ops.py
WarningRan: Good catch — yes, sorted_unique would be empty iff `byte_addresses` is empty, which is guarded upstream by the `ValueError` at the top of `read_scattered`. Added a one-line comment above `sorted_unique[0]` pointing at the upstream guard so the reader doesn't have to re-derive it:
```
sorted_unique = sorted(set(byte_addresses))
# sorted_unique non-empty: byte_addresses guarded above.
runs: List[List[int]] = [[sorted_unique[0]]]
```

WarningRan:
Addressed all four in the latest push. Thank you! @fabianlim 

**Correctness**

1. read_scattered cross-allocation hazard — went with option (1), docstring invariant. Option (2) needs a simulator alloc_extent API; filed #67.
2. LX-load result.data.nbytes fallback — replaced with a strict raise, matching the contract below. Pinned by test_data_size_raises_when_tile_result_missing_unique_sticks.
3. Empty IAT enumeration — `if not addrs: continue` in _resolve_idx_reads. _build_indirect_coords iterates the same enumeration so the zero-extent view consumes nothing on the assembly side. Pinned by test_resolve_idx_reads_zero_extent_skips_view.

**Hygiene**

Trimmed all four: docstring follow-up line and perf note in memory_ops.py; block comments above both raises in latency.py; "C3 regression" / "pre-fix" / "pre-sideband" framing in test docstrings.

fabianlim (approved):
all my comments are addressed

Issue #58, opened 2026-05-15
[Feature] Implement non-identity variables_space_order in indirect_load / indirect_store


WarningRan:
## Background

`MemoryOps.indirect_load` and `MemoryOps.indirect_store` currently `raise NotImplementedError` when `variables_space_order` (`vso`) is non-identity. This was a deliberate Phase 1 measure in #49 to replace silently-wrong code with explicit rejection — the original `points = [vso.eval(pt) for pt in points]` shortcut both corrupted variable bindings and didn't actually reorder enumeration.

## Correct implementation approach

(Per review feedback in #49.)

`vso` controls **the position each variable-space point occupies in the output tile** (load) or source tile (store). It does not affect subscript resolution.

```
For each pt in vss.enumerate():
  coord    = resolve_subscripts(pt)   # always uses original pt, unchanged
  out_pos  = vso.eval(pt)             # where this element lives in the tile
```

Concretely for `indirect_load`:

```
variable space     subscript resolution     output tile
(d0=0,d1=0) ──► X[ IDX1[0,0], IDX2[0,0] ] ──► tile[ vso(0,0) ]
(d0=0,d1=1) ──► X[ IDX1[0,1], IDX2[0,1] ] ──► tile[ vso(0,1) ]
...
```

With identity `vso` (current behaviour), enumeration order = tile order, so gathered values land in the right positions automatically. With non-identity `vso`, gather in enumeration order, then permute the result array using `vso.eval(pt)` as the destination index before returning the tile.

`indirect_store` is the dual by symmetry — read from `tile[vso.eval(pt)]` rather than the enumeration position when writing back. (The review worked through the load case explicitly; the store side is inferred.)

## Scope

- Implement the post-gather permutation step in `indirect_load`.
- Implement the dual write-side permutation in `indirect_store`.
- Remove `raise NotImplementedError` in both.
- Add test fixtures with non-identity `vso` (no fixture exercises this path today).

## Context

- Fix does not require changing `vss.enumerate` — enumeration order stays unchanged.
- Identified during review of #49 (the original error message and rationale comments were rewritten in commit `dd84f61` to use the correct framing).

fabianlim:
# Discussion: Issue #58 — variables_space_order in indirect_load / indirect_store

@WarningRan I conceptualized what is needed to fix this issue.

## What `vso` means

`vso` controls **iteration order** over the variable space — the sequence in
which points are visited, and therefore the order in which gathered values fill
[...]

Issue #52, opened 2026-05-15
[Feature] BoxSet rectangular fast paths in direct ktdp.load/store and indirect_load


fabianlim:
## Background

`BoxSet` — the axis-aligned specialisation of `AffineSet` — was introduced alongside `ktdp.construct_distributed_memory_view`. The distributed paths (`MemoryOps.distributed_tile_access`, `distributed_load`, `distributed_store`) take advantage of the box structure end-to-end: O(ndim) intersection at access-tile construction, and a rectangular sub-`TileRef` slice at load/store time — replacing per-coord scatter with one NumPy slice paste per surviving partition.

The single-allocation paths (`ktdp.load`, `ktdp.store`, `indirect_load`, and once #49 lands `indirect_store`) only get a partial benefit:

- Parse-time lowering still returns a `BoxSet` for axis-aligned `coordinate_set` / `access_tile_set` / `variables_space_set` attributes, so `BoxSet.enumerate(shape)` runs without an AST walk (purely structural O(∏shape)).
- But the call sites then materialise the enumerated point list and feed it to `MemoryOps.load(coords=...)` / `store(coords=...)` / the indirect-access enumeration — same code path as if the set had been a non-axis-aligned `AffineSet`.

## Problem

`ktdp.load` / `ktdp.store` and the indirect-access paths don't have a structural fast path analogous to the rectangular sub-`TileRef` shortcut in `distributed_load` / `distributed_store`. When the access region is an axis-aligned box (the common case), the current code still:

1. Enumerates every point in the box into a `List[Tuple[int, ...]]`.
2. Linearises that list to flat element offsets via `_flat_memory_offsets`.
3. Reads the surrounding `span` of bytes and gathers / scatters per offset.

For a contiguous rectangular access this collapses to a strided slice that NumPy can do in one operation; the per-coord coord-list step is wasted work.

## Scope

Add a structural fast path to direct and indirect access:

1. **`ktdp.load` (direct).** When `access_tile.coordinate_set` is `BoxSet` and `coordinate_order` is identity (or absent):
   - Build a sub-`TileRef` covering the box (inheriting parent strides; `base_ptr` shifted by `box.lo · strides · bpe`).
   - Call `MemoryOps.load(context, sub_tile_ref)` and use the returned `Tile` directly — no coord-list path.
   - Fall through to the existing point-list path when `coordinate_order` is non-identity (the order-permutation path is rare and not box-shaped after permutation).
[...]

PR #66, opened 2026-05-20
Support non-identity variables_space_order in indirect_load/store

WarningRan:
Addresses #58.

## The change

Replaces the `NotImplementedError` for non-identity `variables_space_order` (vso) in `MemoryOps.indirect_load` / `MemoryOps.indirect_store` with a sort-then-gather implementation per RFC 0682 §473: enumerate the indirect access tile in declared order, then reorder by `vso.eval` before each gather/scatter.

## Where the change lives

- `ktir_cpu/ops/memory_ops.py` — drop `NotImplementedError`; add a permutation guard on vso at both `indirect_load` and `indirect_store` entries (rejects non-bijective maps with `ValueError("... must permute its input dimensions ...")`); introduce a module-level `_enumerate_in_vso_order(iat)` helper that returns the variable-space points sorted by `vso.eval` (identity / absent vso falls through to the natural row-major enumeration). The two #65 helpers `_resolve_idx_reads` and `_build_indirect_coords` each route their `points = ...` line through this helper, so the indirect-side and the coord-build side stay lockstep without changing #65's public helper API.
- `ktir_cpu/affine.py` — replace probe-based `is_identity` / `is_permutation` with structural AST checks (`_expr_as_single_dim` reusing `_constraint_to_linear`). The probe-based versions accepted false positives such as `(d0+d1-2, d0+d1-1)` evaluating to `(1, 2)` on probe `[1, 2]`. The structural check inspects each output expression at the AST level and admits only `1 * d_i + 0` forms.
- `tests/test_affine.py` — `TestAffineMapIsPermutation` (10 cases including 1-D edge, 3-cycle, shear/many-to-one/non-square/linear-combination/constant-offset rejections, trivial-wrapper acceptance) and `TestAffineMapIsIdentity` (4 cases pinning the probe-bug regressions).
- `tests/test_indirect_access.py` — swap-vso load+store, 3-cycle vso load+store on a 2×2×2 fixture (chosen because non-involution distinguishes correct sort-then-gather from any apply-then-load implementation), and a parameterised non-permutation-rejected case with `match="must permute its input dimensions"`.

## What is NOT changing

- **`coordinate_space_order` (cso) direct-path semantics.** This PR is scoped to vso on the indirect path; cso is untouched. See "Deferred" below.
- **Dynamic-shape / symbolic vso.** `AffineMap` parsing rejects symbol forms today; vso here is dim-only. Symbolic bounds will compose with the existing `n_syms` plumbing (PR #42) when needed.
- **ODS-level verification.** The bijection check lives in the runtime path; the dialect verifier in `ktir-mlir-frontend` is unchanged.

## Concerns we worked through

- **Sort-then-gather vs apply-then-load.** RFC §473's iteration-order wording, as clarified in #58's worked example, is satisfied by either implementation when vso is an involution (e.g. a single swap), but they diverge for non-involutions. The 2×2×2 3-cycle fixture pins this: expected outputs are derived by direct simulation, not closed-form, so any future implementation that drifts to apply-then-load will fail the test rather than silently match.
- **Why structural over probe for `is_identity` / `is_permutation`.** Any cheap probe has linear-combination collisions for general n; the structural check reuses the existing `_constraint_to_linear` primitive and admits only `1 * d_i + 0` per output.
- **Guard non-redundancy.** Identity is a permutation, so `not vso.is_identity() and not vso.is_permutation()` reduces to `not vso.is_permutation()`. Kept the single check; identity remains the fast path that skips the sort.
- **Store-side symmetry.** Store applies the same sort key as load. This matters for non-involution vso: the scatter-side ordering must match the gather-side ordering produced by an upstream `indirect_load`, otherwise a load→store round-trip with the same vso is not a no-op. Pinned by `test_indirect_store_with_3cycle_vso`.
- **Rebased on #65; sort placement.** #65 split the indirect path into `_resolve_idx_reads` + `_build_indirect_coords`, which both consume a separate `enumerate(iat.shape)`. A literal "hoist the sort before the helpers" form would have required widening #65's public helper signatures to take a pre-computed `points` argument. Instead, both helpers route their enumeration through `_enumerate_in_vso_order(iat)`, which is the single source of truth for vso ordering — same external behavior, no API churn on the just-merged #65 surface.

## Tests added

- `tests/test_affine.py` — 14 cases on `is_permutation` + `is_identity` (1-D edge through 3-cycle, plus probe-collision regressions: `(d0+d1-2, d0+d1-1)` and `(d1-1, d0+1)`).
- `tests/test_indirect_access.py` — 4×4 swap-vso load+store; 2×2×2 3-cycle load+store (expected from direct simulation, distinguishes sort-then-gather from apply-then-load on non-involutions); parameterised non-permutation rejection with `match="must permute its input dimensions"`.

Full suite went 775 → 795 passed (+20) on top of the post-#65 main baseline (1 skipped, 7 xfailed unchanged); the structural `is_identity` tightening did not break any existing fixture.

## Deferred to follow-ups

- **`coordinate_space_order` direct-path consistency.** RFC §394 (cso) and §473 (vso) use the same lexicographic-traversal language, but `MemoryOps.{load, store}` apply cso pointwise on the direct path while the indirect path here uses sort-then-gather. Either §394's text means something subtly different from §473, or the direct path has a latent inconsistency. Will file a separate issue once we agree on the reading — happy to draft it. Flag if you read §394 differently and this is a non-issue.
- Bulk vectorisation of the sort+gather path is scoped to #52 (bulk-read refactor across load/store); not in this PR.
- ODS-level vso bijection verifier in `ktir-mlir-frontend` (no second non-identity-permutation consumer yet).
- Symbolic / dynamic-shape vso composing with `n_syms` (no current consumer).

### DIFF
--- a/ktir_cpu/affine.py
+++ b/ktir_cpu/affine.py
@@ -98,7 +98,7 @@ class AffineMap:
         return eval_affine_map(self, dims)
 
     def is_identity(self) -> bool:
-        """Return True if this map is the identity: output == input for all inputs.
+        """Return True if this map is the identity: output[i] == d_i for every i.
 
         Used at parse time to detect trivial coordinate-order maps.  When
         ``access_tile_order`` is an identity map it has no effect on which
@@ -106,16 +106,51 @@ class AffineMap:
         ``coordinate_order`` to ``None``.  This allows load/store to skip
         the per-coord ``cso.eval()`` calls and, combined with a full
         ``coordinate_set``, enables the contiguous fast path entirely.
+
+        Implemented structurally: each output expression must flatten to
+        ``1 * d_i + 0`` with output position ``i`` matching the dim index.
+        A probe-based ``eval(probe) == probe`` check would accept maps
+        like ``(d0, d1) -> (d1 - 1, d0 + 1)`` (probe ``[1,2]`` → ``(1,2)``),
+        which are not identity.
         """
-        # Quick structural check: same number of inputs and outputs, and each
-        # output expression is just the corresponding input dimension variable.
         if len(self.exprs) != self.n_dims:
             return False
-        # Verify by evaluation: identity map satisfies eval(dims) == dims for
-        # a non-trivial probe vector.  Using [1, 2, ..., n_dims] avoids false
-        # positives from constant-zero expressions.
-        probe = list(range(1, self.n_dims + 1))
-        return list(self.eval(probe)) == probe
+        for i, expr in enumerate(self.exprs):
+            idx = _expr_as_single_dim(expr, self.n_dims)
+            if idx != i:
+                return False
+        return True
+
+    def is_permutation(self) -> bool:
+        """Return True if this map permutes its input dimensions.
+
+        A permutation map is square (output count equals input count) and
+        each output expression is exactly one dim variable, with every dim
+        index appearing exactly once.  Accepts coordinate permutations
+        like ``(d0, d1, d2) -> (d2, d0, d1)``; rejects shears, scalings,
+        constant offsets, and many-to-one collapses.
+
+        Used by ops whose semantics require iteration over the input space
+        in a permuted order (e.g. ``variables_space_order`` on indirect
+        access tiles): the implementation sorts enumerated points by the
+        map's image, which is well-defined only when the image is a
+        permutation of the original points.
+
+        Implemented structurally on the parsed AST.  A probe-based
+        ``sorted(eval(probe)) == probe`` check would accept linear
+        combinations such as ``(d0, d1) -> (d0 + d1 - 2, d0 + d1 - 1)``
+        (probe ``[1,2]`` → ``(1,2)``), which are not coordinate
+        permutations.
+        """
+        if len(self.exprs) != self.n_dims:
+            return False
+        seen = set()
+        for expr in self.exprs:
+            idx = _expr_as_single_dim(expr, self.n_dims)
+            if idx is None or idx in seen:
+                return False
+            seen.add(idx)
+        return True
 
 
 @dataclass(frozen=True)
@@ -323,6 +358,29 @@ class BoxSet:
         if any(v is None for v in los) or any(v is None for v in his):
             return None
         return cls(lo=tuple(los), hi=tuple(his))  # type: ignore[arg-type]
+
+
+def _expr_as_single_dim(node: "_Node", n_dims: int) -> Optional[int]:
+    """Return ``i`` if *node* is structurally equivalent to ``1 * d_i + 0``.
+
+    Returns ``None`` for any expression that flattens to a linear form
+    with a non-zero constant, a non-unit coefficient, or coefficients on
+    more than one dim variable.  Used by :meth:`AffineMap.is_identity`
+    and :meth:`AffineMap.is_permutation` for structural checks that
+    cannot be fooled by linear combinations whose evaluation on a
+    specific probe happens to coincide with the probe (e.g.
+    ``d0 + d1 - 2`` evaluates to ``1`` on probe ``[1, 2]``).
+    """
+    lin = _constraint_to_linear(node, n_dims)
+    if lin is None:
+        return None
+    coeffs, const = lin
+    if const != 0:
+        return None
+    nz = [i for i, k in enumerate(coeffs) if k != 0]
+    if len(nz) != 1 or coeffs[nz[0]] != 1:
+        return None
+    return nz[0]
 
 
 def _constraint_to_linear(node: "_Node", n_dims: int) -> Optional[Tuple[List[int], int]]:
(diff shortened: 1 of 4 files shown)


WarningRan:

ktir_cpu/ops/memory_ops.py
```diff
@@ -690,10 +698,10 @@ def indirect_store(
```
fabianlim, line 908: This still uses the per-element read loop from the original `indirect_load`. PR #65 refactors this into `_resolve_idx_reads` + `read_scattered` (one accessor per index view, contiguous-run batching). Depending on merge order this PR may need to be rebased on top of #65 or the two need to be coordinated.

What's the plan — does this merge before or after #65?

WarningRan: Good catch — I missed this when scoping #66. Agreed the per-element loop here will be gone after #65 lands, so the sort placement question is different in the post-refactor structure (sort hoisted before `_resolve_idx_reads`, since vso only affects enumeration order, not per-point resolution).

Plan: merge #66 *after* #65. Once #65 lands I'll rebase, move the `sorted(points, key=vso.eval)` ahead of `_resolve_idx_reads` so both `read_scattered` and `_build_indirect_coords` consume the same reordered enumeration, and re-run the 3-cycle test to confirm it still distinguishes sort-then-gather from apply-then-load under the new batching.

Holding this PR until then.

WarningRan:
Rebase this PR, please review, thank you @fabianlim !

ktir_cpu/affine.py
```diff
@@ -325,6 +360,29 @@ def try_from_affine_set(cls, aset: "AffineSet") -> Optional["BoxSet"]:
         return cls(lo=tuple(los), hi=tuple(his))  # type: ignore[arg-type]
 
 
+def _expr_as_single_dim(node: "_Node", n_dims: int) -> Optional[int]:
```
fabianlim: The helper name `_expr_as_single_dim` was hard for me to parse without context — "as single dim" reads like a coercion but the function is really a structural matcher that returns the dim index or `None`. Suggest renaming to **`_match_pure_dim_ref`**: "match" signals the None-on-no-match return convention, and "pure dim ref" captures the "coefficient 1, constant 0, exactly one dim" constraint.

Also worth expanding the docstring with a small examples table — the flattening behavior (via `_constraint_to_linear`) means algebraically-equivalent forms like `d0 + 0` or `d0 + d1 - d1` collapse to the same result, which isn't obvious from the prose. Something like:

```
Examples (n_dims=3):
    d0           -> 0
    d2           -> 2
    d1 + 0       -> 1     # zero constant ok
    d0 + 1       -> None  # non-zero constant
    2 * d0       -> None  # non-unit coefficient
    d0 + d1      -> None  # more than one dim
    -d0          -> None  # coefficient -1, not 1
```

Optional: a test pinning the cancellation behavior (e.g. `(d0, d0 + d1 - d1)` → `is_identity()`) would lock in the flatten-form semantics so a future syntactic rewrite doesn't silently change it.

fabianlim:
Re: the deferred cso/vso direct-vs-indirect divergence noted in the description — agree deferring is the right call. The direct-path fix isn't a one-liner: it needs the same sort-then-gather vs apply-then-load reading settled for §394, a `cso.is_permutation()` guard, and updates to existing fixtures that pin the current pointwise semantics. Better to settle the spec question in a separate issue first.

ktir_cpu/affine.py
WarningRan: Thanks — addressed all three in the latest commit:

- Renamed `_expr_as_single_dim` → `_match_pure_dim_ref`; both call sites in `is_identity` / `is_permutation` updated.
- Expanded the docstring with your examples table verbatim, plus a sentence noting that the flatten path through `_constraint_to_linear` is what makes algebraically-equivalent forms (`d0 + 0`, `d0 + d1 - d1`) collapse to the same match.
- Added `test_identity_through_cancellation` in `TestAffineMapIsIdentity`. Small adjustment to your literal example: `(d0, d0 + d1 - d1)` flattens to `(1*d0 + 0, 1*d0 + 0)` — the second output reduces to `d0`, not `d1`, so the map isn't identity. The test uses `(d0 + d1 - d1, d1 + d0 - d0)` instead, which exercises subtractive cancellation in both slots and *is* identity. Same intent (pin flatten-form so a future syntactic-only matcher can't silently regress this).

WarningRan:
Thanks for confirming the deferral. I'll file a separate issue, scoped to the three sub-tasks you listed:

1. settle §394 vs §473 reading (sort-then-gather vs pointwise / apply-then-load) at the spec level first;
2. add `cso.is_permutation()` guard symmetric with the vso guard from this PR;
3. migrate the fixtures that currently pin pointwise cso semantics.

Will link the new issue back from the §394 references and from the "Deferred" section of this PR's description once it's open.

fabianlim (approved):
all my comments have been addressed

PR #69, opened 2026-05-28
Feature: BoxSet support for symbolic dynamic shapes

### LINKED ISSUE #51: [Feature] BoxSet support for dynamic shapes (symbolic n_syms)
## Background

`BoxSet` — the axis-aligned specialisation of `AffineSet` — was introduced alongside `ktdp.construct_distributed_memory_view`. Parse-time lowering puts distributed memory views on an O(ndim) fast path when both the partition extent `B_i` and the access set `A` are `BoxSet`; the slow path (full `AffineSet.enumerate` brute force) is preserved for non-axis-aligned sets.

PR #42 (`[Fix] Fix dynamic shapes: memref<?>, SSA sizes as operands, symbolic affine dims`) added symbolic affine sets via `AffineSet.n_syms` and a `symbols` argument on `contains` / `enumerate`.

## Problem

`BoxSet` does not yet compose with `n_syms` / `symbols`. `BoxSet.try_from_affine_set` rejects symbolic sets (`n_syms > 0`, via `getattr` for backwards compatibility) and lets them fall through to the `AffineSet` branch.

Concretely, for distributed memory views:

- A distributed memory view whose partition `coordinate_set`s reference a symbolic dim (e.g. dynamic batch size) parses to `AffineSet` instead of `BoxSet`.
- `MemoryOps.distributed_tile_access` then takes the slow path: brute-force `B_i.enumerate(view_shape)` per partition, point-by-point membership test against `x + A`, and a per-coord scatter inside `distributed_load` / `distributed_store`.
- The BoxSet fast-path performance characteristics (~58,000× on `distributed_tile_access`, ~315× end-to-end on a 256×512 fixture) are forfeited for any dynamic-shape distributed view.

The TODO is documented in `ktir_cpu/affine.py`'s module docstring.

## Scope

This issue is narrowly about lifting the `n_syms > 0` rejection so symbolic distributed views land on the BoxSet fast path. Concretely:

1. **Symbolic `BoxSet.lo` / `BoxSet.hi`.** Accept either `int` or a symbolic expression node (matching `AffineSet`'s constraint AST atoms). Mixed concrete/symbolic axes are allowed.
2. **`symbols` argument on `BoxSet.contains` / `enumerate` / `is_full` / `is_empty` / `lower_bounds`.** Mirrors `AffineSet`'s post-#42 signature so call sites can pass through a `symbols` tuple uniformly. When all axes are concrete, the argument is ignored.
3. **`BoxSet.try_from_affine_set` accepts `n_syms > 0`.** Lowering succeeds when every constraint still has the form `c * d_i + k(syms) >= 0` with `c ∈ {+1, -1}`, where `k(syms)` is a linear combination of integer constants and symbol references. The resulting `BoxSet.lo[i]` / `hi[i]` carries the symbolic expression directly.
4. **`BoxSet.translate` / `intersect` accept symbolic operands.** Translation produces a new symbolic box; intersection with another (symbolic or concrete) box produces a box whose bounds are `max(lo_a, lo_b)` / `min(hi_a, hi_b)` — represented symbolically when either side is symbolic.
5. **`distributed_tile_access` fast path keeps firing on symbolic inputs.** When `B_i` and `(x + A)` are symbolic boxes, `B_i.intersect(translate(x + A))` resolves at runtime against the bound `symbols` tuple, producing a concrete `BoxSet` for the survivor.

## Out of scope

- Non-convex symbolic regions (unions of boxes).
- BoxSet fast paths in direct `ktdp.load`/`ktdp.store` and `indirect_load` — tracked separately.

## Test plan

- Parametrize the existing distributed-view fast/slow path tests over a `?`-dim variant. The symbolic version must:
  - Round-trip through `parse_affine_set` as a symbolic `BoxSet` (not `AffineSet`).
  - Produce identical output to a fully-concrete reference run after binding `symbols` to concrete values.
  - Assert that `survivor.coordinate_set` is `BoxSet` (regression guard against silent fall-through to `AffineSet`).
- Add a dynamic-shape RFC example or test fixture exercising a `memref<?x64xf16>` distributed view with two partitions whose row counts are symbolic.
- Existing concrete-shape tests in `tests/test_affine.py` and `tests/test_distributed_view.py` must continue to pass unchanged.

## References

- PR #42: introduced `AffineSet.n_syms` and the `symbols` argument; the model `BoxSet` needs to compose with.
- `ktir_cpu/affine.py` module docstring — TODO is recorded inline next to `BoxSet`'s definition.

WarningRan:
## Summary

Resolves #51 (in-scope item) — partition coordinate sets carrying symbolic dynamic dims now lower to a symbolic `BoxSet` and resolve through the construct-memory-view boundary, keeping `distributed_tile_access` on the O(ndim) fast path under dynamic shapes.

## What changes

- **`BoxSet`** carries `Bound = int | tuple` per-axis; cached `_all_concrete` flag short-circuits hot-path queries on fully-concrete boxes (no isinstance tax).
- **`try_from_affine_set`** lifts the `n_syms > 0` reject; constraints with linear symbol terms become symbolic `lo` / `hi` AST expressions (assumes `s_i >= 0`, matching dim-size semantics).
- **`construct_memory_view` handler** eagerly specialises symbolic coordinate sets per the ODS contract — symbols bind positionally to the dynamic operands listed in `sizes:` (KtdpOps.td:137-139), not to the resolved shape tuple. `memref<64x?xf16>` correctly binds `s_0` to the trailing dynamic dim, not the leading static `64`.
- **`distributed_tile_access`** keeps a single specialise-at-boundary contract: callers supply concrete coordinate sets; symbol resolution is the construction-time handler's job.
- **`construct_access_tile`** raises `NotImplementedError` on symbolic `access_tile_set` — defensive fail-fast at the boundary.

## Design decisions

1. **Eager specialise at the construct-memory-view boundary** rather than threading symbols through `distributed_tile_access`. Both produce the same concrete result on the dialect path; eager keeps the resolution as a single architectural layer and lets the partition-loop inner intersect run on plain ints.
2. **Symbol binding source = `sizes:` operands, not full shape.** ODS spec `KtdpOps.td:137-139` defines this verbatim. The previous reading (positional against the resolved full shape) silently miscompiled `memref<64x?xf16>` by stealing symbol index 0 for the static leading dim. The trailing-dynamic test in `tests/test_dialects_exec.py` is the regression guard.
3. **Single constraint flattener with a dim-only wrapper.** `_constraint_to_linear_syms` is the canonical sym-aware walker; `_constraint_to_linear` is a thin wrapper that calls it with `n_syms=0` so any `sym` atom trips the guard and returns `None`, preserving the dim-only "reject symbols" contract. Initial cut kept these as parallel walkers; consolidated in cleanup pass per review feedback (`_match_pure_dim_ref` does call `_constraint_to_linear` via `AffineMap.is_identity` / `is_permutation`, but those are parse-time / op-entry, not inner loops, so the wrapper overhead is in the noise).
4. **MVP simplification on `sym_max` / `sym_min`**: concrete fold, same-`SymRef` idempotence, `+0` absorption, `-(-x)` collapse. No nested absorption / commutativity / canonicalisation. Per-axis candidate count is ≤ 2 in realistic IR, so deeper folding would not reduce live AST shape; canonicalisation would also need a structural-equality engine outside this PR's scope.
5. **`contains` uses inline `eval_bound`; `enumerate` / `is_full` / `lower_bounds` route through `specialize`.** Trade-off: single-point queries avoid AST allocation; whole-box consumers amortise one specialise. Deliberate, not a leak.
6. **Symbols `s_i >= 0`** (matches dim-size semantics). Stated in `try_from_affine_set` docstring. If signed symbols become a requirement, lowering needs a guard or fallback to AffineSet.

## Out of scope (deferred follow-ups)

- **Nested-absorption / canonicalisation for `sym_max` / `sym_min`.** Triggered when realistic IR produces > 2 candidates per axis.
- **Non-distributed `ktdp.load` / `ktdp.store` BoxSet fast path.** Issue #51 explicitly out of scope.
- **`AffineSet.is_full(symbols=...)`** signature alignment. Held back to keep this PR one logical change; not blocking any current consumer.

## Tests

New coverage:

- Sym-bound helper unit tests (`test_ast.py::TestSymBoundHelpers`).
- Symbolic `BoxSet` round-trip tests (`test_affine.py::TestSymbolicBoxSet`).
- Parametrised distributed-view geometry tests on symbolic-derived partitions (`test_distributed_view.py::test_distributed_tile_access_dynamic_shape_emits_box_set`).
- Dialect-handler tests for symbol-binding correctness, including the trailing-dynamic `Nx?` regression guard (`test_dialects_exec.py::TestKtdp::test_construct_memory_view_specializes_symbolic_coord_set_*`).
- `pytest.raises(NotImplementedError)` lock-down on the symbolic-`access_tile_set` path.

## RFC / spec conformance

- `coordinate_set` symbol binding: matches the `ktdp.construct_memory_view` ODS contract (KtdpOps.td:137-139) — "symbols ... bound to variables in the `sizes` operand, from left-to-right in a one-to-one fashion".
- No new ops, attributes, or semantic changes to existing ops.

## Test plan checklist (issue #51 §6)

- [x] Round-trip parse to symbolic `BoxSet`, not `AffineSet`.
- [x] Bound to concrete symbol values, output bytewise-equal to all-concrete reference (geometry test parametrised over `partition_rows`).
- [x] Regression guard: `assert isinstance(survivor.coordinate_set, BoxSet)`.
- [x] Dynamic-shape fixture: `memref<?x4xf16>` and `memref<64x?xf16>` distributed-view-style fixtures (handler level).
- [x] Existing concrete-shape tests unchanged.

---

### DIFF
--- a/ktir_cpu/affine.py
+++ b/ktir_cpu/affine.py
@@ -46,13 +46,17 @@ axis-aligned, fully-pinned, unit-coefficient, non-symbolic sets to
 ``AffineSet``.  ``parse_affine_set_raw`` skips the lowering for tests
 that need to inspect the AST directly.
 
-TODO: BoxSet does not yet compose with AffineSet's ``n_syms`` /
-``symbols`` parameters (added in PR #42 for dynamic-shape symbolic
-bounds).  ``try_from_affine_set`` currently rejects symbolic sets
-(``n_syms > 0``) so they stay on the AffineSet branch.  When symbolic
-boxes are needed, ``BoxSet.lo``/``hi`` will need to accept symbolic
-expressions and ``contains``/``enumerate`` will need to take a
-``symbols`` argument.
+Symbolic bounds: ``BoxSet.lo`` / ``hi`` may carry an integer constant
+(concrete leaf, fast path) or an AST node (a :data:`Bound` over symbol
+variables).  Per-axis operations that depend on a concrete value
+(``contains``, ``is_empty``, ``is_full``, ``enumerate``) take a
+``symbols`` argument that resolves the symbolic bounds; pure-int boxes
+short-circuit on the ``_all_concrete`` flag and ignore ``symbols``.
+``try_from_affine_set`` lowers symbolic axis-aligned sets — see its
+docstring for the accepted constraint shape.  Geometric operations
+(``intersect``, ``translate``) use :func:`parser_ast.sym_max` /
+``sym_min`` / ``sym_add`` so concrete-on-concrete cases stay O(ndim)
+without constructing AST nodes.
 
 TODO: ``parse_affine_set_raw`` (in ``parser_ast.py``) exists only because
 ``parse_affine_set`` now lowers axis-aligned inputs to ``BoxSet`` by
@@ -68,12 +72,18 @@ until after C4 to keep this commit faithful to backup.
 from __future__ import annotations
 
 import itertools
-from dataclasses import dataclass
-from typing import TYPE_CHECKING, List, Optional, Sequence, Tuple
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, List, Optional, Sequence, Tuple, Union
 
 if TYPE_CHECKING:
     # Avoid circular import at runtime; parser_ast imports nothing from here.
     from .parser_ast import _Node
+
+
+# A bound on a single axis of :class:`BoxSet`.  ``int`` is the concrete leaf
+# (fast path); a tuple is an AST node over symbol variables, evaluated by
+# :func:`parser_ast.eval_bound`.
+Bound = Union[int, tuple]
 
 
 @dataclass(frozen=True)
@@ -226,27 +236,60 @@ class BoxSet:
     visible at each call site via ``isinstance`` dispatch, not hidden
     behind polymorphism.  Mixed-type operations — e.g.
     ``BoxSet.intersect(aset: AffineSet)`` — raise ``TypeError``.
+
+    Symbolic bounds: ``lo[d]`` / ``hi[d]`` may be an ``int`` (concrete)
+    or an AST node tuple over symbol variables (see :data:`Bound` and
+    :func:`parser_ast.eval_bound`).  Whether the box is fully concrete
+    is cached in :attr:`_all_concrete` at construction so the hot path
+    (per-axis ``contains`` / ``is_empty``) skips ``isinstance`` checks
+    on every element.
     """
-    lo: Tuple[int, ...]   # inclusive
-    hi: Tuple[int, ...]   # exclusive
+    lo: Tuple["Bound", ...]   # inclusive
+    hi: Tuple["Bound", ...]   # exclusive
+    # Cached at __post_init__: True iff every entry in lo/hi is a plain int.
+    # ``compare=False`` keeps equality / hashing keyed off the bounds alone
+    # (the flag is a derived property).  ``init=False`` so callers don't pass it.
+    _all_concrete: bool = field(default=False, init=False, compare=False, repr=False)
 
     def __post_init__(self) -> None:
         if len(self.lo) != len(self.hi):
             raise ValueError(
                 f"BoxSet: lo/hi length mismatch: lo={self.lo} hi={self.hi}"
             )
+        all_concrete = all(isinstance(b, int) for b in self.lo) and all(
+            isinstance(b, int) for b in self.hi
+        )
+        # Frozen dataclass: bypass the descriptor to set the cached flag once.
+        object.__setattr__(self, "_all_concrete", all_concrete)
 
     @property
     def n_dims(self) -> int:
         return len(self.lo)
 
-    def contains(self, point: Sequence[int]) -> bool:
-        """True iff ``lo[d] <= point[d] < hi[d]`` for every dim."""
+    def contains(self, point: Sequence[int], symbols: Sequence[int] = ()) -> bool:
+        """True iff ``lo[d] <= point[d] < hi[d]`` for every dim.
+
+        ``symbols`` is required to resolve symbolic bounds; concrete
+        boxes ignore it (the ``_all_concrete`` flag short-circuits the
+        AST walk).  Passing too few symbols on a symbolic box raises
+        ``IndexError`` (from :func:`parser_ast.eval_bound`) — the
+        contract matches :meth:`AffineSet.contains`.
+        """
         if len(point) != self.n_dims:
             return False
-        return all(self.lo[d] <= point[d] < self.hi[d] for d in range(self.n_dims))
+        if self._all_concrete:
+            return all(self.lo[d] <= point[d] < self.hi[d] for d in range(self.n_dims))
+        from .parser_ast import eval_bound
+        return all(
+            eval_bound(self.lo[d], symbols) <= point[d] < eval_bound(self.hi[d], symbols)
+            for d in range(self.n_dims)
+        )
 
-    def enumerate(self, shape: Optional[Tuple[int, ...]] = None) -> List[Tuple[int, ...]]:
+    def enumerate(
+        self,
+        shape: Optional[Tuple[int, ...]] = None,
+        symbols: Sequence[int] = (),
+    ) -> List[Tuple[int, ...]]:
         """Return all integer points in the box in row-major order.
 
         ``shape`` is accepted for signature parity with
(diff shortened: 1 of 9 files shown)


WarningRan:
cc @fabianlim — issue #51 follow-up; comments welcome on the eager-specialise-vs-thread choice (design decision 1) and the scope cut on symbolic `access_tile_set` (see comment on #51).

fabianlim (commented):
...

fabianlim (commented):
...

fabianlim (commented):
...

fabianlim (commented):
...

ktir_cpu/parser_ast.py
```diff
@@ -496,6 +506,108 @@ def enumerate_affine_set(aset: AffineSet, shape: Tuple[int, ...], symbols: Seque
...
+# ``sym_*`` constructors apply the minimum constant folding needed to keep
+# ``intersect`` / ``translate`` from accumulating trivially redundant AST
+# nodes per partition (concrete fold of two ints; idempotence on the same
+# ``("sym", k)`` reference; additive identity).  Deeper canonicalisation
+# (commutativity, nested absorption) is intentionally out of scope — it
+# would need a structural-equality engine and the realistic candidate
+# count per axis is ≤ 2, so deep nesting does not arise.
+# ---------------------------------------------------------------------------
+
+Bound = Union[int, tuple]
```
fabianlim, line 527: **Single `Bound` definition.** This is also defined in `ktir_cpu/affine.py:86`. Pick one home (parser_ast is the natural one since the AST helpers live here) and import from it.

ktir_cpu/affine.py
```diff
@@ -405,6 +521,19 @@ def _constraint_to_linear(node: "_Node", n_dims: int) -> Optional[Tuple[List[int
     dimension variables and integer constants (e.g. contains ``sym`` or
     ``ref`` atoms).  The constraint represents
     ``sum(coeffs[i] * d_i) + const >= 0``.
+
+    Symbolic constraints (``sym`` atoms) take the AffineSet branch — see
+    :func:`_constraint_to_linear_syms` for the variant that accepts them.
+
+    .. warning::
```
fabianlim: **Consolidate flatteners.** Drop this; replace with a thin wrapper around `_constraint_to_linear_syms`. The "byte-identical hot path" justification is unfounded — `_match_pure_dim_ref` (in `parser_ast.py`) doesn't call `_constraint_to_linear` at all. The only caller is `BoxSet.try_from_affine_set`, parse-time only.

```python
def _constraint_to_linear(node, n_dims):
    """Flatten a dim-only constraint AST into ``(coeffs, const)``.

    Wrapper over :func:`_constraint_to_linear_syms` with ``n_syms=0`` —
    any ``sym`` atom in the AST trips the ``j >= n_syms`` guard and
    returns ``None``, preserving the original "reject symbols" contract.
    """
    result = _constraint_to_linear_syms(node, n_dims, n_syms=0)
    if result is None:
        return None
    dim_coeffs, _sym_coeffs, const = result
    return dim_coeffs, const
```

Removes the duplicated walker and the misleading docstring warning that ships with it.

ktir_cpu/affine.py
```diff
@@ -226,78 +236,174 @@ class BoxSet:
...
+        On symbolic boxes the comparison can only be answered for
+        concrete shapes after resolving against ``symbols`` (a tuple
+        equality on AST nodes would always fail structurally even when
+        the runtime extent matches).
+        """
         if len(shape) != self.n_dims:
             return False
-        return self.lo == (0,) * self.n_dims and tuple(self.hi) == tuple(shape)
+        if self._all_concrete:
+            return self.lo == (0,) * self.n_dims and tuple(self.hi) == tuple(shape)
```
fabianlim: **`is_full` is not translation-invariant.** Current `lo == (0,) and hi == shape` returns `False` on a translated box even when it covers `[x, x + shape)`. Latent today but easy to trip. Replace with extent comparison, and apply the same fix to any other site that does `lo == 0 and hi == shape`:

```python
def is_full(self, shape, symbols=()):
    if len(shape) != self.n_dims:
        return False
    spec = self if self._all_concrete else self.specialize(symbols)
    return all(spec.hi[d] - spec.lo[d] == shape[d] for d in range(self.n_dims))
```

fabianlim (commented):
Overall: LGTM with cleanup. Eager-specialise at `construct_memory_view` is the right design; the trailing-`?` test is a real regression guard.

Inline comments below cover the required changes. Two notes that don't pin to a single line:

- **Drop the gap_analysis rows (36a, 36b).** `BoxSet` is a perf specialisation, not a spec concept — RFC 0682 defines `coordinate_set` as `IntegerSetAttr`. Symbolic sets already worked pre-PR via `AffineSet`; this PR is a perf change, not a conformance change. Track via #51.
- **End-to-end MLIR test gap (follow-up):** parse a `?`-shape distributed view and execute through the dialect handlers all the way to `ktdp.load`. Current tests stop at the ops layer (`access_tile_set=None`) or the handler boundary in isolation — nothing exercises the full path on symbolic input.

Smaller follow-ups: `lower_bounds` docstring claims "always tuple of int" but field is typed `Tuple[Bound, ...]` and `# type: ignore` is doing real work; `test_translate_concrete_offset_preserves_symbols` only asserts the concrete side; `_build_sym_term` lacks coverage for negative-coefficient symbols (e.g. `d_i - s_0 >= 0`).

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -64,15 +65,35 @@ def ktdp__construct_memory_view(op, context, env):
...
+    # SSA operands listed in ``sizes:`` (the dynamic dims), not the
+    # static-int entries — for ``memref<64x?xf16>`` the static leading
+    # ``64`` must not steal symbol index 0 from the dynamic tail dim.
+    # Pre-resolution, dynamic entries appear as SSA-name strings in
+    # ``shape_attr``; we filter on that to recover the dynamic operands
+    # in declaration order, then resolve each through ``context``.
+    # Substituting here keeps downstream consumers (find_partition /
+    # distributed_tile_access / load) on the concrete fast path — they
+    # never see a symbolic set.
+    if isinstance(coordinate_set, BoxSet) and not coordinate_set._all_concrete:
```
fabianlim: **Expose `is_concrete` instead of leaking `_all_concrete`.** Reaching into the dataclass internal couples the dialect handler to `BoxSet`'s implementation. A `BoxSet.is_concrete` property is one line. Same applies at `:152`.

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -116,10 +137,32 @@ def ktdp__construct_access_tile(op, context, env):
...
+    # ``$symbol_operands`` argument list — independent from
+    # ``$indices`` and from the access tile shape.  The ktir-cpu Python
+    # parser does not yet surface that operand list on the
+    # ``Operation`` object; wiring it through and threading the
+    # resolved symbols here is tracked as an issue-#51 follow-up
+    # (gap_analysis row 36b).  Fail fast at this boundary rather than
+    # leaking a symbolic set into ``distributed_tile_access`` where it
+    # would surface as an opaque ``IndexError`` from ``eval_bound``.
+    if isinstance(coordinate_set, BoxSet) and not coordinate_set._all_concrete:
+        raise NotImplementedError(
```
fabianlim, line 153: On the deferral: the handler side is trivial (~5 lines mirroring the memory-view case). The real work is the parser — `parse_construct_access_tile` currently calls `find_ssa_names` indiscriminately and folds everything into `indices`, with no separation between index operands and symbol operands. Adding that needs a surface-syntax decision and will likely touch other parsers that share the same SSA-collection helpers. Worth scoping before the follow-up PR opens.

ktir_cpu/parser_ast.py
WarningRan: Done. `Bound` now lives in `parser_ast.py` (next to `eval_bound` / `sym_*` helpers); `affine.py` references it only via string annotations and a `TYPE_CHECKING` import — no runtime dependency, no circular import.

ktir_cpu/affine.py
WarningRan: Adopted — replaced the duplicated walker with your snippet. `_constraint_to_linear` is now ~10 lines delegating to `_constraint_to_linear_syms` with `n_syms=0`; the misleading docstring warning is gone, and the sibling docstring reflects the new "canonical flattener" role.

One small clarification on the call graph for the record: `_match_pure_dim_ref` actually does go through `_constraint_to_linear` (in `affine.py`, inside its body), and it's used by `AffineMap.is_identity` / `is_permutation`, which are called from `mlir_frontend/parser.py`, `ktdp_ops.py`, and `memory_ops.py`. But those calls happen at parse-time / op-entry rather than in any inner loop, so the wrapper's overhead (one extra call + an empty `sym_coeffs` list) is in the noise — your consolidation conclusion still holds. Mentioning it just so the call graph is documented for future readers.

ktir_cpu/affine.py
WarningRan: Took a closer look at the call sites before adopting the extent form, and I think we want to keep the strict `[0, shape)` semantics here. The call at `ktdp_ops.py:parse_construct_access_tile` uses `coordinate_set.is_full(access_shape)` as the trigger to drop `coordinate_set` to `None`, and dropping it puts `ktdp.load` / `ktdp.store` on the contiguous fast path that assumes a zero origin. With the extent form, a translated box would slip through:

```python
>>> BoxSet(lo=(5,), hi=(15,)).is_full((10,))      # extent form: True
>>> parse_affine_set_raw(
...   "affine_set<(d0) : (d0 - 5 >= 0, -d0 + 14 >= 0)>"
... ).is_full((10,))                              # AffineSet:    False
```

The peer `AffineSet.is_full` is also `[0, shape)` (vertex check on the corners of the box), so I'd like to keep the two implementations agreeing on the contract since `coordinate_set` is typed `BoxSet | AffineSet | None`.

Adopted the rest of your suggestion: plumbed `symbols=()` through so symbolic boxes can be queried, and unified the concrete/symbolic branches with `spec = self if self.is_concrete else self.specialize(symbols)` (using the new public property from your other comment). The docstring now spells out the asymmetry-vs-extent and points at the fast-path contract. If translation-invariant extent comparison is useful elsewhere, I'd lean toward a separate `BoxSet.has_extent(shape)` helper rather than overloading `is_full` — happy to add that or revisit `is_full`'s contract if you have a use case in mind.

ktir_cpu/dialects/ktdp_ops.py
WarningRan: Done. Added a one-line `BoxSet.is_concrete` property; both call sites in `ktdp_ops.py` now read the property. Internal callers inside `BoxSet` keep using the cached field directly (no extra attribute lookup on the hot path).

ktir_cpu/dialects/ktdp_ops.py
WarningRan: Thanks for the heads-up on the parser side — that's a much bigger surface than just the handler. Tracked as #73 with the parser scope spelled out (the indiscriminate `find_ssa_names` split, the surface-syntax decision, and the sibling-parser audit).

Also dropped the `gap_analysis row 36b` reference from the `NotImplementedError` message (matching the gap-analysis cleanup you flagged in the summary); the message now points to a follow-up.

WarningRan:
Thanks for the review! Cleanup pass is up: e65530f covers the five inline comments plus the summary feedback. Quick rundown:

- ① `Bound` unified to `parser_ast.py`; `affine.py` imports under `TYPE_CHECKING` only.
- ② `_constraint_to_linear` is now a thin wrapper over `_constraint_to_linear_syms` (your snippet); duplicate walker and the docstring warning removed.
- ③ `BoxSet.is_full` — kept `[0, shape)` semantics to stay aligned with `AffineSet.is_full` and the zero-origin fast path that `ktdp_ops.parse_construct_access_tile` uses to drop `coordinate_set` to `None`; adopted your `symbols=()` plumbing and `spec`-based unified branch. Repro and details inline (open to revisiting if there's a translation-invariant use case I'm missing).
- ④ `BoxSet.is_concrete` property added; both `ktdp_ops.py` sites updated.
- ⑤ `construct_access_tile` defer message no longer references `gap_analysis row 36b`; the parser-side work you flagged is now scoped in #73 (linked below).
- ⑥ Dropped rows 36a / 36b from `docs/gap_analysis.md` — agreed this is a perf change tracked by issue #51, not an RFC-conformance gap.
- ⑦ End-to-end MLIR symbolic test (`?`-shape view → `ktdp.load`) is tracked as #74 (linked below).
- ⑧ Three small follow-ups also addressed: `lower_bounds` docstring + `cast` (replacing `# type: ignore`), symbolic-side assertion added in `test_translate_concrete_offset_preserves_symbols`, and a new `test_negative_symbol_coefficient_in_bound` exercising `_build_sym_term` on `d_i - s_0 >= 0` (the negated-symbol path).

Test status: 814 passed / 1 skipped / 7 xfailed (baseline 813 / 1 / 7; the +1 is the new negative-coefficient test).

---

Follow-up issues opened for the two items above:

- #73 — `[Feature] Wire $symbol_operands through parse_construct_access_tile and resolve symbolic access_tile_set` (the parser surface-syntax decision + handler wiring; replaces the current `NotImplementedError` defer).
- #74 — `[Test] End-to-end MLIR test gap for symbolic distributed views` (parse a `?`-shape distributed view and execute through the dialect handlers all the way to `ktdp.load`, including a `memref<64x?xf16>` trailing-dynamic case).

ktir_cpu/affine.py
fabianlim: The `→ None` normalization is the wrong move here. `None` means "load everything contiguously from base" — it can't represent a translated slice. The right fix is to normalize to a `BoxSet` covering `[0, shape)` instead of `None`, and retire the `None` sentinel. Then the `isinstance(cs, BoxSet)` fast path in load/store handles all rectangular cases uniformly, including translated slices. `is_full` becomes unnecessary.

Separating this out — fits in #52.

ktir_cpu/dialects/ktdp_ops.py
fabianlim: @WarningRan i may have misunderstood the issue here, I posted a comment on #73 saying that we may not need to support this case. All the dynamic symbols should be resolved in the memory views. The access tiles should already be concrete.

WarningRan: Confirmed via grep across `examples/` and `tests/` — every access_tile_set is dim-only; `vector_add_dynamic_ktir`.mlir (the only `?`-shape e2e example) keeps a concrete `access_tile_set` even though `coordinate_set` is symbolic. Closing #73 as not-planned. Keeping the `NotImplementedError` as a defensive guard.

ktir_cpu/affine.py
WarningRan: Agreed — retiring the `None` sentinel is the right direction. Once `coordinate_set` normalizes to `BoxSet([0, shape))` instead of `None`, the `isinstance(cs, BoxSet)` fast path in load/store handles translated slices uniformly and `is_full` falls out.

Keeping the current `[0, shape)` semantics in this PR to preserve the existing `None`-sentinel contract until that lands. Tracked in #52.

fabianlim (approved):
all my issues have been resolved and followup issues created.

PR #70, opened 2026-05-30
[Feature] affine eq node: first-class == constraint in affine sets

fabianlim:
Introduces `("eq", expr)` as a first-class AST node for equality constraints in affine sets, replacing the previous parse-time rewrite (from `inter-tile-reduce` / `4bdf137`) that expanded `==` into two `>=` inequalities.

## Changes

- **Tokeniser**: `==` added to `_TOKEN_RE` before `>=` (longest-match)
- **`parse_constraint_list`**: `lhs == rhs` emits `("eq", ("sub", lhs, rhs))` — one node, normalised to `expr == 0` form (mirrors how `>=` constraints normalise to `expr >= 0`)
- **`_eval_node`**: `eq` tag returns the inner expression value
- **`affine_set_contains`**: checks `== 0` for `eq` nodes, `>= 0` for everything else
- **`BoxSet.try_from_affine_set`**: unified pin logic handles `eq` (sets both `lo` and `hi`) alongside existing inequality bounds; duplicated linearise-then-extract code removed

## Spec examples (all parse and evaluate correctly)

- `affine_set<(g) : (g == 0)>`
- `affine_set<(i) : (i == 0)>`
- `affine_set<(p)[c] : (p - c + 2 == 0)>`
- `affine_set<(p)[c, g] : (p + c - 8*g - 3 == 0)>`

## Coordination with PR #69

`test_eq_multi_dim_rejected` in `tests/test_affine.py` is marked with a `TODO` and will need to be revisited when PR #69 lands.

PR #71, opened 2026-05-31
parser utilities for inter-tile communication

fabianlim:
**Stacked on [#70](https://github.com/torch-spyre/ktir-cpu/pull/70) (`affine-eq-node`).** Branch off `affine-eq-node`'s tip; rebase before merge once #70 lands. The PR diff currently includes `f0cfdbf` (the `affine-eq-node` commit) until #70 merges into `main`.

Three pure-additive parser-side helpers. No callers in this PR — they're prerequisites for the inter-tile-reduce work coming next.

## Commits

- **`a207238` `parser`: split flush condition into two-stage evidence/veto.**
  Tokenizer's "is the previous op done" reframed as Stage 1 (positive evidence: type-terminal OR new SSA assignment) + Stage 2 veto (a bare `{` is a region-body opener, never an op header). Lets ops whose type annotation lands on the *last attribute line* and whose region opens on the *next line* keep their region attached. Behaviour-preserving for every existing kernel.

- **`b494048` `parser_utils`: add `extract_named_attr`.**
  Public, caller-driven extractor for the bare `key = value` form some ops use *outside* a `{ ... }` block. Sibling to `parse_attr_block`. Caller names the key it expects.

- **`39c0416` `parser_ast`: add `enumerate_membership_keys`.**
  Generic affine-set primitive: \"for which keys `k ∈ domain` does `family(k)` admit `point`?\" Two-line implementation via `domain.enumerate` + `family.contains`, so it works for both `AffineSet` and `BoxSet`. Naming uses `family` / `domain` / `key` rather than the loaded affine-grammar word \"symbol.\"

## Why now

The follow-up `inter-tile-reduce` PR uses all three:

- ops whose attributes live on the op header (no `{ ... }` block) → `extract_named_attr`.
- ops whose region opens on the line after the type annotation → two-stage flush.
- \"which group does this core belong to?\" lookup in the consume handler → `enumerate_membership_keys`.

Landing them as a focused parser PR keeps the inter-tile review surface smaller and gives the helpers room to be reused by other dialects.

## Tests

\`820 passed, 1 skipped, 7 xfailed\`. No new tests — the helpers have no callers yet; their consumers ship in the inter-tile-reduce PR with full coverage.

### DIFF
--- a/ktir_cpu/parser_utils.py
+++ b/ktir_cpu/parser_utils.py
@@ -238,6 +238,57 @@ def parse_attr_list(op_text: str, aliases: Optional[Dict] = None,
     return result
 
 
+def extract_named_attr(op_text: str, key: str,
+                       aliases: Optional[Dict] = None) -> Optional[str]:
+    """Extract a single bare ``key = value`` attribute from ``op_text``.
+
+    Sibling to :func:`parse_attr_block` for ops that carry attributes
+    *outside* a ``{ ... }`` block — e.g. ops whose attribute list lives
+    on the op header itself rather than wrapped in braces:
+    ``producer_tiles_per_group = #all_tiles, groups = #one_group``.
+
+    Returns the resolved value string (alias-resolved when applicable),
+    or ``None`` if not found.  Caller-driven: caller names the key it
+    expects.
+    """
+    m = re.search(rf'\b{re.escape(key)}\s*=\s*', op_text)
+    if not m:
+        return None
+    rest = op_text[m.end():].lstrip()
+
+    # #alias reference
+    if rest.startswith('#'):
+        end = re.search(r'[,\n}]|\s+:|\s*->', rest)
+        raw = rest[:end.start()].strip() if end else rest.strip()
+        return aliases.get(raw, raw) if aliases else raw
+
+    # keyword<...> values — count <> depth, skip >= and ->
+    kw_m = re.match(r'[\w.]+<', rest)
+    if kw_m:
+        i = kw_m.end() - 1
+        depth = 0
+        while i < len(rest):
+            ch = rest[i]
+            if ch == '>' and i + 1 < len(rest) and rest[i + 1] == '=':
+                i += 2
+                continue
+            if ch == '-' and i + 1 < len(rest) and rest[i + 1] == '>':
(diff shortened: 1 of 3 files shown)


fabianlim:
cc @kiszk @lchu6 @nwang-ibm — stacked on #70.

fabianlim:
@kiszk can you help me take a look at this PR? This is eventually required for the #72 inter-tile PR

ktir_cpu/parser_utils.py
```diff
@@ -238,6 +238,57 @@ def parse_attr_list(op_text: str, aliases: Optional[Dict] = None,
...
+        return None
+    rest = op_text[m.end():].lstrip()
+
+    # #alias reference
+    if rest.startswith('#'):
+        end = re.search(r'[,\n}]|\s+:|\s*->', rest)
+        raw = rest[:end.start()].strip() if end else rest.strip()
+        return aliases.get(raw, raw) if aliases else raw
+
+    # keyword<...> values — count <> depth, skip >= and ->
```
kiszk, line 265: It would be good to create a common utility function based on part of `_extract_attr_value()`. Not mandatory.

ktir_cpu/parser_utils.py
```diff
@@ -238,6 +238,57 @@ def parse_attr_list(op_text: str, aliases: Optional[Dict] = None,
...
+    Sibling to :func:`parse_attr_block` for ops that carry attributes
+    *outside* a ``{ ... }`` block — e.g. ops whose attribute list lives
+    on the op header itself rather than wrapped in braces:
+    ``producer_tiles_per_group = #all_tiles, groups = #one_group``.
+
+    Returns the resolved value string (alias-resolved when applicable),
+    or ``None`` if not found.  Caller-driven: caller names the key it
+    expects.
+    """
+    m = re.search(rf'\b{re.escape(key)}\s*=\s*', op_text)
```
kiszk, line 254: What happens when a key is empty (i.e. `key=`).

kiszk (approved):
While I left two comments, they are not blockes.

fabianlim:
@kiszk i will address those 2 items in the next PR #72

PR #72, opened 2026-06-01
[Experimental] Inter tile reduce

fabianlim:
cc: @tnakaike @Prasanth-Chatarasi @lchu6 @mudhakar 

Stacked on:
  1. [ktir-cpu#70](https://github.com/torch-spyre/ktir-cpu/pull/70)
     `affine-eq-node` — `==` as a first-class AST node.
  2. [ktir-cpu#71](https://github.com/torch-spyre/ktir-cpu/pull/71)
     `parser-utils-flush-fix` — three parser-side helpers
     (`parse.py` two-stage flush, `parser_utils.extract_named_attr`,
     `parser_ast.enumerate_membership_keys`).

Merge order: #70 → #71 → this PR.  Until #70 and #71 land on `main`,
the diff against `main` includes their commits; review against
`#71` is the focused view.

Spec: [ktir-mlir-frontend#23](https://github.com/torch-spyre/ktir-mlir-frontend/pull/23)
(🧪 unmerged — surface may shift)

## Summary

Implement the **reduce path** of the four-op design: `ktdp.inter_tile_produce` + `ktdp.inter_tile_reduce`. Runs a 4-core all-reduce end-to-end with per-core ring-bandwidth accounting. All new ops are 🧪 experimental.

**In:** two ops + `yield_partial`/`yield_reduced` terminators, `!ktdp.tile_future<T_p>` (N=1), per-core `TileFuture`, `CommPlan` (producers/consumers/deps; plan-aware fold), end-to-end bandwidth pipeline (`bytes_moved` → `Tile.comm_bytes` → `LatencyReport`), rewritten `ring_reduce.mlir`, two latency tests.

**Out (deferred):** `inter_tile_consume`, `inter_tile_reduce_scatter`, `producer_dependency_per_consumer` runtime (full-barrier semantics for now), N>1 partials, spec-invariant verifier checks.

## Latency traces and scores

`HardwareConfig()` defaults: `clock_ghz = 1.0`,
`hbm_bandwidth_tb_s = 1.0` (per-core: 31.25 B/cyc on the 32-chip
default), `ring_bandwidth_tb_s = 4.0` (4000 B/cyc),
`simd_elements_per_cycle = 64`.  Both kernels load a
`tensor<1x128xf16>` row per core (256 bytes = `8.192` HBM cycles).

**Combiner work is charged per fold, on the core that runs it.**
`reduce_fn` (the user's combiner region) executes inside
`RingReduceBackend.run` once per *in-plan* round — not once per ring
round.  Each call routes through ``execute_region → _execute_op →
record_op``, so every `linalg.add` (`128/64 = 2.0` cycles), every
`tensor.empty`, and every `arith.cmpi` lands in *that core's*
``compute_cycles`` bucket.  Out-of-plan rounds still send / receive
on the wire (`bytes_moved` accumulates), but the fold is skipped, so
no compute charge accrues for them.  The single-`scf.if` writer adds
`8.192` HBM cycles on writer cores only.

### Single-group, 4 cores (`examples/ktir/ring_reduce.mlir`)

```
kernel_cycles: 22.5916           ← max core (= core 0, the writer)
core 0:  total=22.5916  compute=6.0156  memory=16.3840  comm=0.1920   (writer)
core 1:  total=14.3996  compute=6.0156  memory= 8.1920  comm=0.1920
core 2:  total=14.3996  compute=6.0156  memory= 8.1920  comm=0.1920
core 3:  total=14.3996  compute=6.0156  memory= 8.1920  comm=0.1920
```

`comm = 0.192 = (4 − 1) × 256 / 4000`.  Three messages per core,
one per ring round.

### Multi-group, 16 cores in 4 groups of 4 (`examples/latency/ring_reduce_multi_group.mlir`)

```
kernel_cycles: 23.3596           ← max core (any of the four group writers)
core  0 (writer    ): total=23.3596  compute=6.0156  memory=16.3840  comm=0.9600
core  1 (non-writer): total=15.1676  compute=6.0156  memory= 8.1920  comm=0.9600
core  4 (writer    ): total=23.3596  compute=6.0156  memory=16.3840  comm=0.9600
core  5 (non-writer): total=15.1676  compute=6.0156  memory= 8.1920  comm=0.9600
core  8 (writer    ): total=23.3596  compute=6.0156  memory=16.3840  comm=0.9600
core 12 (writer    ): total=23.3596  compute=6.0156  memory=16.3840  comm=0.9600
```

`comm = 0.96 = (16 − 1) × 256 / 4000`.  **Five times** the
single-group value — exactly the ratio `(num_cores − 1)` scales
to.  The four group writers all share `kernel_cycles = 23.3596`;
core 0 has no special status now that the workgroup partitions
the writer responsibility evenly.

The writer/non-writer total-cycle gap is `8.1920` — the extra HBM
store on writers, identical between the two kernels.

## Architecture

Design lives in
[`docs/cross_core_scheduling.md`](docs/cross_core_scheduling.md)
§"Inter-tile communication" — that's the canonical reference.  The
shape, in one breath:

> `TileFuture` is per-core, holding this core's partial.  The
> consume handler builds a `CommPlan` (producers, consumers,
> optional `producer_dependency_per_consumer`) and dispatches to a
> `ReduceBackend`.  The backend's ring spans the **whole
> workgroup** — every core sends and receives every round — and the
> plan acts as a *fold mask*: only tiles originating from
> `plan.producers` get folded into a consumer's accumulator,
> non-producers seed with `identity`, and non-consumers return
> `None`.  Per-core `bytes_moved` lands on `Tile.comm_bytes`; the
> latency tracker reads it to charge ring bandwidth.  Read the doc
> for diagrams.

Two non-obvious bits worth surfacing here:

- **`_wrap_latency_counter` (interpreter).** Comm handlers return a
  generator; `record_op` fires *after* the generator returns its
  real value (with `comm_bytes` set), not at handler entry.  Wrap
  is opt-in — only when the latency tracker is on.

- **Workgroup size injection.** `CoreContext.attach_scheduler`
  takes a `num_cores` kwarg and exposes it via `ctx.num_cores`.
  Backends size their protocols from this (the live IR-grid count),
  not from `HardwareConfig.num_cores` (the chip-spec constant —
  see "Known issue" below).

## Cleanups

Behaviour-preserving refactor alongside the architecture work.

- **`attach_reshape` + backend-stamps-`comm_bytes`.**  Handler's
  inline `_reshape` + `_gen` closures move into `ktdp_helpers.py`
  as `reshape_tile_to_target` and `attach_reshape`.  Backend stamps
  `result.comm_bytes = self.bytes_moved` itself, keeping the
  fill / publish on the same side of the contract.  Consume-handler
  tail:
  ```python
  plan = CommPlan.for_reduce(...)
  backend = _select_reduce_backend(plan, op.attributes)
  target_shape = op.attributes.get("_result_shape")
  raw = backend.run(context, local_tile, plan, reduce_fn, identity)
  return attach_reshape(raw, target_shape)
  ```

## Tests

`826 passed, 1 skipped, 6 xfailed`. New:
- `test_examples.py::TestRingReduceExecution::test_ring_reduce_sum`
- `test_latency.py::TestRingReduceLatency::test_ring_reduce_per_core_breakdown`
- `test_latency.py::TestRingReduceMultiGroupLatency::test_multi_group_correctness_and_bytes`

## Follow-ups

COMMENT: compact this

- **Track [ktir-mlir-frontend#23](https://github.com/torch-spyre/ktir-mlir-frontend/pull/23).**
  The four-op spec is upstream-unmerged; reconcile any op / attr
  drift before merging this on `main`.  The 🧪 decoration in
  `ktdp_ops.py` and `gap_analysis.md` calls it out.
- **Spec-invariant verification.** Three checks are deferred until
  the upstream spec stabilises (see the "Verification — deferred"
  block in `ktir_cpu/dialects/ktdp_ops.py`):
  - **A2.** ``groups`` match between produce and consume.  Bounded
    enumeration over ``ctx.num_cores`` — no polyhedral set-difference
    machinery needed.
  - **B1.** ``producer_dependency_per_consumer ⊆
    producer_tiles_per_group`` (subset).  Post-check on
    ``CommPlan.for_reduce``.
  - **B2.** Every producer is named by some consumer's deps
    (coverage).  Same place as B1.

  The single-use invariant on ``%fut`` (§2.3) is *not* on the list:
  the per-core ``TileFuture`` instance is the def-use edge for our
  purposes, so the simulator can't double-consume even silently —
  see ``TileFuture``'s docstring and
  ``docs/cross_core_scheduling.md`` §"Def-use is simulated by
  TileFuture, not walked".
- **Broadcast / reduce-scatter delivery ops.** Substrate in place;
  each needs its own handler (and a `BroadcastBackend` for consume).
- **Per-tile sync runtime.**
  `producer_dependency_per_consumer` is parsed and resolved into
  `CommPlan.deps`; the v1 ring backend ignores it (every fold uses
  the full producer set).  A future smarter ring or
  `PointToPointReduceBackend` would honour `plan.producers_for(c)`
  to short-circuit unnecessary folds.
- **Disambiguate the two `num_cores` facts.**  There are two
  `num_cores`-style facts in the simulator and the names collide:
  - **`GridExecutor.num_cores`** — live workgroup size from the
    IR's `grid` attribute.  Always available.  Reachable via
    `ctx.num_cores` (injected by `attach_scheduler`).  Used by
    `RingReduceBackend.run` to size the ring.
  - **`HardwareConfig.num_cores`** — chip-spec constant
    (default 32).  Only present with latency tracking on.
    Parameterises
    `hbm_bytes_per_cycle_per_core = total_hbm_bw / num_cores`.

  These are different facts: an IR `grid = [4]` running on a
  32-core chip is still bandwidth-shared across 32 cores' worth of
  HBM ports.  The current code is correct
  (`RingReduceBackend.run` reads `ctx.num_cores`, latency
  `_estimate` reads `HardwareConfig.num_cores`), but the name
  overlap is easy to mix up at call sites.  Rename
  `HardwareConfig.num_cores` to something hardware-flavoured
  (`hw_num_cores`, `chip_cores`, …) so the workgroup-vs-chip
  distinction is visible.

### DIFF
--- a/docs/cross_core_scheduling.md
+++ b/docs/cross_core_scheduling.md
@@ -56,33 +56,10 @@ KTIRInterpreter
 
 | Layer | Responsibility | Notes |
 |---|---|---|
-| **`CommOps`** | High-level **per-core** comm primitives. Stable surface for dialect handlers. Today: `reduce` (passthrough to a `ReduceBackend`). | Dialect handlers call into here, not into scheduler primitives directly. |
 | **`CoreContext`** | Per-core SSA scope, LX scratchpad, and the public `send_to` / `get_lx` methods. Per-run scheduler bindings (`_send_fn`, `_transfer_fn`) are installed via `attach_scheduler` and cleared via `detach_scheduler`. | |
 | **`CoreExecutionStack`** | Wraps one core's op list as a generator. The only place that calls `.send()` / `next()` on client generators. Bubbles `RecvRequest` to the scheduler via `yield from`. | Internal. |
 | **`GridExecutor`** | Owns cores, the message queue, and the scheduler loop. Attaches scheduler bindings on every core before stepping; drives every generator to completion or raises on deadlock. | Internal. |
 | **`ReduceBackend`** / **`TransferBackend`** | Pluggable algorithms / transports. `ReduceBackend.run` is per-core, may yield `RecvRequest`. `TransferBackend.run` resolves remote LX access. Siblings under the scheduler protocol; distinct purposes. | See "Pluggable …" sections. |
-
----
-
-## Per-core semantics
-
-`CommOps` primitives describe **one core's behavior**. The function
-runs once per participating core; the scheduler runs N copies
-concurrently; they cooperate via the message queue.
-
-### `CommOps.reduce` — passthrough to a `ReduceBackend`
-
-```python
-@staticmethod
-def reduce(ctx, tile, core_group, backend: ReduceBackend):
-    return backend.run(ctx, tile, core_group)
-```
-
-The algorithm — ring rounds, LX-scratchpad accumulation, etc. — lives
-in the backend. `CommOps.reduce` exists so dialect handlers and tests
-have a single stable entry point regardless of which algorithm is in
-play. See `RingReduceBackend` for the canonical ring algorithm and a
-worked example.
 
 ---
 
@@ -100,9 +77,9 @@ A handler that calls a generator-shaped backend must **propagate**
 the generator — don't call it and discard the result:
 
 ```python
-return CommOps.reduce(ctx, tile, group, backend)              # OK — scheduler drives it
-return (yield from CommOps.reduce(ctx, tile, group, backend)) # OK — handler is itself a generator
-CommOps.reduce(ctx, tile, group, backend)                     # BUG — generator created, body never runs
+return backend.run(ctx, tile, group, reduce_fn)              # OK — scheduler drives it
+return (yield from backend.run(ctx, tile, group, reduce_fn)) # OK — handler is itself a generator
+backend.run(ctx, tile, group, reduce_fn)                     # BUG — generator created, body never runs
 ```
 
 **Key invariant:** `send_to` is fire-and-forget — the sender never
@@ -114,14 +91,13 @@ sender-side deadlock in symmetric patterns (all-to-all, ring).
 ## Control flow: blocking recv (end-to-end)
 
 ```
-dialect handler / test handler
-  └── CommOps.reduce(ctx, tile, group, backend)   ← passthrough
-        backend.run(ctx, tile, group)             ← generator object
-          for each round:
-            ctx.send_to(next_core, tile)          ← enqueues immediately
-            received = yield RecvRequest(src=prev) ← suspends here
-            ...
-          return result
+dialect handler
+  └── backend.run(ctx, tile, group, reduce_fn)    ← generator object
+        for each round:
+          ctx.send_to(next_core, tile)            ← enqueues immediately
+          received = yield RecvRequest(src=prev)  ← suspends here
+          ...
+        return result
 
 CoreExecutionStack._execute_until_block
   result = execute_op(op, core)                   ← gets the generator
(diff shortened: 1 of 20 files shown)


nwang-ibm:
Implementation looks correct and well-structured to me (w/ CC). 

A couple of non-blocking notes. 
---

### 1. would you consider Bidirectional Ring (CW + CCW)

Real Spyre hardware has a bidirectional ring (128 B/cycle per direction). Current `RingReduceBackend` is CW-only. A bidirectional backend could halve ring latency for all-reduce patterns. The scheduler already supports this — only a new backend subclass needed:

```python
class BidirectionalRingBackend(ReduceBackend):
    def run(self, context, tile, plan, reduce_fn, identity):
        n = context.num_cores
        my_id = context.core_id
        cw_next, cw_prev = (my_id + 1) % n, (my_id - 1) % n
        ccw_next, ccw_prev = (my_id - 1) % n, (my_id + 1) % n

        seed = tile if plan.is_producer(my_id) else identity
        left, right = split_tile(seed)
        left_fwd, right_fwd = left.copy(), right.copy()

        for k in range(1, (n + 1) // 2 + 1):
            self.bytes_moved += left_fwd.size_bytes() + right_fwd.size_bytes()
            context.send_to(cw_next, left_fwd)
            context.send_to(ccw_next, right_fwd)
            left_recv = yield RecvRequest(src=cw_prev)
            right_recv = yield RecvRequest(src=ccw_prev)
            if plan.is_producer((my_id - k) % n):
                left = reduce_fn(left, left_recv)
            if plan.is_producer((my_id + k) % n):
                right = reduce_fn(right, right_recv)
            left_fwd, right_fwd = left_recv, right_recv

        if not plan.is_consumer(my_id):
            return None
        result = merge_tiles(left, right, seed.shape)
        result.comm_bytes = self.bytes_moved
        return result
```
---

### 2. HardwareConfig vs. Real Hardware

Cross-checked defaults against Spyre MICRO 2026 paper:

| Parameter | Real Hardware | `HardwareConfig` | Status |
|---|---|---|---|
| Clock | **1.2 GHz** | 1.0 GHz | Off by 20% |
| Ring BW per link (one direction) | **128 B/cycle × 1.2 GHz = ~154 GB/s** | — | Not directly modeled |
| Ring BW (segmented, size=4) | **~4–6 TB/s** | 4.0 TB/s | OK (lower bound) |
| Off-chip mem (1p0 LPDDR5) | **204.8 GB/s** | 1.0 TB/s | Off |
| Ring topology | **Bidirectional** | CW only | Missing CCW |

fabianlim:
@kiszk @tnakaike @Prasanth-Chatarasi ... we have a review with @nwang-ibm, if you have time please look through as well

nwang-ibm (dismissed):

nwang-ibm (approved):

Issue #73, opened 2026-06-02
[Feature] Wire $symbol_operands through parse_construct_access_tile and resolve symbolic access_tile_set


WarningRan:
## Background

PR #69 closed the in-scope half of #51 (symbolic partition extents on `construct_memory_view`). The other half — symbolic `access_tile_set` on `construct_access_tile` — was scope-cut in #69 and the handler currently raises `NotImplementedError` to fail-fast at the boundary:

> `construct_access_tile: symbolic access_tile_set is not yet supported — the parser does not surface the op's $symbol_operands operand list (per ODS) needed to resolve the symbols. Tracked as a follow-up to issue #51.`

This issue is that follow-up.

## Why this isn't a 5-line handler change

ODS spec defines `Variadic<Index>:$symbol_operands` on `ktdp.construct_access_tile` as the binding source for symbols on `access_tile_set` — distinct from `construct_memory_view`'s `sizes:` operands. The handler change itself would be small (mirroring the memory-view eager-specialise pattern), but the parser does not currently surface the symbol-operand list separately from index operands.

Concretely [`ktdp_ops.py:491`](ktir_cpu/dialects/ktdp_ops.py#L491):

```python
operands = find_ssa_names(after_eq)
```

`find_ssa_names` is shared by sibling parsers under `ktir_cpu/dialects/`. There is no current way to tell "this SSA name is an index operand" from "this SSA name is a `$symbol_operands` entry" at the surface-syntax level — so a surface-syntax decision has to come first, and may require updates to other parsers that share the same SSA-collection helpers.

## Scope

1. **Decide on a surface-syntax convention** for separating `$indices` from `$symbol_operands` in the printed form of `ktdp.construct_access_tile`. Options:
   - Mirror `construct_memory_view`: an explicit `symbols:[...]` keyword for the symbol operand list.
   - Inline brackets after the access-tile type, e.g. `!ktdp.access_tile<...>[%s0, %s1]`.
   - A trailing `symbols=...` attribute (less natural since these are SSA values, not constants).
2. **Update `parse_construct_access_tile`** to extract `$symbol_operands` separately from `$indices` per the chosen convention.
3. **Audit shared helpers** (`find_ssa_names` and friends in `ktir_cpu/parser_utils.py`) — anywhere else that conflates index and symbol operands needs the same treatment, or the helper needs to grow a "skip symbols" mode.
[...]

PR #79, opened 2026-06-03
[Refactor] Route ktdp.load/store through _subtile_ref; drop is_full→None sentinel

### LINKED ISSUE #52: [Feature] BoxSet rectangular fast paths in direct ktdp.load/store and indirect_load

WarningRan:
Closes #52.

> Updated 2026-06-09 — round-2 reframed as caller-side cleanup; see PR comment for analysis. Title retitled per reviewer suggestion.

## Summary

Three additions and one forced fixture fix, landed together as one logical change ("BoxSet fast path end-to-end on single-allocation routes"):

- **Parser default-fill.** `AccessTile.coordinate_set = None` retired; parser default-fills `BoxSet([0, shape))` at parse time. Concrete-int `TypeError` on `access_shape` (not `assert` -- must survive `python -O`).
- **Direct path.** `MemoryOps.boxed_load` / `boxed_store` reuse `_subtile_ref` to read/write rectangular slices without enumerating the AffineSet point list. Dispatched from `ktdp.load` / `ktdp.store` on `isinstance(css, BoxSet) and (cso is None or cso.is_identity())`. Empty-box returns zero sticks.
- **Indirect path.** `_build_indirect_coords` dispatches to a vectorised coord build (`np.indices` column-by-column) when `variables_space_set` is `BoxSet` and `variables_space_order` is identity. New `_eval_subscript_vectorized` mirrors the slow-path AST grammar tag-for-tag. Per-pt HBM lookups stay on the slow path.
- **Forced.** `_patch_tile_size` / `_patch_tile_dim0` in `tests/test_latency_modeling.py` were writing `coordinate_space_set`, a non-existent field. The typo was masked by the old `None` sentinel; the default-fill exposes it. Both helpers now write the real `coordinate_set`.

## Architecture notes

A few non-obvious design moves worth surfacing:

- **`partition_origin or zero_tuple` quiet polymorphism.** `_subtile_ref` is the distributed-path helper; on a non-distributed parent its fallback collapses cleanly to `box.lo * strides`. The direct path reuses it instead of writing a parallel non-distributed helper. Trade-off accepted: ties the direct path to a distributed-path implementation detail; no-trample tests pin the contract.

- **`cso.is_identity()` guard upgrade.** The parser invariant "identity normalised to `None`" was an implicit contract spread across three files. Promoting it to a local call-site self-check converts cross-file maintenance to local maintenance for cold-path cost only.

- **Indirect vectorisation scope.** `_resolve_idx_reads` is already per-view batched -- rewriting it has zero collinear win. Only the structural coord build moves into NumPy; per-pt indirect HBM lookups stay because stick locality is unpredictable and outside the BoxSet structure.

- **Symbolic shapes ruled out at access-tile layer.** Confirmed on PR #69 (dynamic shapes resolved at memory view layer; access tiles concrete by construction). #73 closed not-planned. The new `TypeError` is a defence against future parser changes leaking symbols past that boundary.

## Performance framing

Constant-factor savings on the Python-side coord build for the indirect path; no measured benchmarks gathered. Reasonable benchmark target is the paged-tensor indirect-access fixture if reviewers ask.

The deeper `O(N -> 1)` collapse the issue alludes to would need a `MemoryOps.load(coords: ndarray)` API change which is out of scope here -- `_flat_memory_offsets` still iterates per coord and consumes `List[Tuple]`, so the closing `tolist()` / `tuple()` conversion is where the entire Python-side overhead of the indirect fast path lives.

## Tests added

`tests/test_boxset_fastpath.py` -- ten functions / 13 collected items, picked by non-overlapping failure modes:

- **Unit `boxed_load` / `boxed_store`** on offset, full, and empty boxes. Pins the helpers' contract independent of dispatch.
- **Dispatch verification** in both directions: `ktdp.load` / `ktdp.store` reach the boxed path for axis-aligned `BoxSet`s, AND fall back to the AffineSet path under non-identity `coordinate_order` (previously every test only verified the fast path was *taken*, never that it was *rejected* when it should be).
- **End-to-end no-trample** on column-packed and row-major fixtures. Expectations are derived from `_flat_memory_offsets` rather than hand-counted addresses, so a future stride-rule change does not silently invalidate them.
- **Indirect parity** parametrised over `floordiv` / `mod` / `neg` / `max`, comparing the `BoxSet` fast path against the AffineSet slow path on the same coords. Subscript-grammar regressions surface as test failures rather than as silent address divergence.

`tests/test_latency_modeling.py::_patch_tile_size` and `_patch_tile_dim0` are fixed in passing -- see "Forced" above.

## Follow-ups

- **`MemoryOps.load(coords=...)` accepting `ndarray`** would let `_build_indirect_coords_box` skip its closing `tolist()` / `tuple()` conversion entirely. The entire Python-side overhead of the indirect fast path lives in that conversion. File as follow-up if performance proves to matter.
- **Forced-AffineSet-vss fixture** for indirect parity tests. The current `test_mixed_direct_expr_and_indirect_matches_slow_path` re-implements slow-path coord build manually; a manually-constructed `AffineSet` bypassing `try_from_affine_set`'s axis-aligned lowering would be cleaner. Judged not-worth-it today; revisit if a slow-path order change actually breaks parity.

### DIFF
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -193,23 +193,23 @@ def ktdp__load(op, context, env):
         return MemoryOps.distributed_load(
             context, access_tile.parent_ref, result_shape=result_shape
         )
-    css = access_tile.coordinate_set    # BoxSet | AffineSet  (parser default-fills BoxSet)
+    css = access_tile.coordinate_set    # BoxSet | AffineSet (always present post-parse)
     cso = access_tile.coordinate_order  # AffineMap | None
     # BoxSet fast path: rectangular access with identity coordinate order →
-    # delegate to MemoryOps.boxed_load, which builds a sub-TileRef and lets
-    # MemoryOps.load take its own contiguous/strided fast path on the sub-ref.
-    # Box extent IS the iteration shape (the data tile shape contract is 1:1
-    # for BoxSet); non-rectangular AffineSet still routes through the
+    # build a sub-TileRef on the box and let MemoryOps.load take its own
+    # contiguous/strided fast path on the sub-ref.  Mirrors the BoxSet branch
+    # in MemoryOps.distributed_load — same _subtile_ref + MemoryOps.load
+    # composition, keeping a single strategy across distributed and single-
+    # allocation paths.  Non-rectangular AffineSet still routes through the
     # coord-list path below where ``_result_shape`` reshapes the gather.
     if isinstance(css, BoxSet) and (cso is None or cso.is_identity()):
-        return MemoryOps.boxed_load(context, access_tile.parent_ref, css)
-    if css is not None:
-        coords = css.enumerate(access_tile.shape)
-        if cso is not None:
-            coords = [cso.eval(pt) for pt in coords]
-        result_shape = op.attributes.get("_result_shape", access_tile.shape)
-        return MemoryOps.load(context, access_tile.parent_ref, coords=coords, result_shape=result_shape)
-    return MemoryOps.load(context, access_tile.parent_ref)
+        sub_ref = MemoryOps._subtile_ref(access_tile.parent_ref, css)
+        return MemoryOps.load(context, sub_ref)
+    coords = css.enumerate(access_tile.shape)
+    if cso is not None:
+        coords = [cso.eval(pt) for pt in coords]
+    result_shape = op.attributes.get("_result_shape", access_tile.shape)
+    return MemoryOps.load(context, access_tile.parent_ref, coords=coords, result_shape=result_shape)
 
 
 @register("ktdp.store", latency_category=LC.MEMORY)
@@ -231,18 +231,18 @@ def ktdp__store(op, context, env):
     tile_ref = access_tile.parent_ref
     css = access_tile.coordinate_set
     cso = access_tile.coordinate_order
-    # BoxSet fast path: symmetric to ktdp.load. Slice the source tile
-    # rectangularly (no-op when box covers the full access tile) and store
-    # through a sub-TileRef. Avoids the read-modify-write scatter that the
-    # coord-list path needs for general AffineSet stores.
+    # BoxSet fast path: symmetric to ktdp.load.  Build a sub-TileRef on the
+    # box and store through it.  Avoids the read-modify-write scatter that
+    # the coord-list path needs for general AffineSet stores.  Same shape as
+    # the BoxSet branch in MemoryOps.distributed_store — single strategy
+    # across distributed and single-allocation paths.
     if isinstance(css, BoxSet) and (cso is None or cso.is_identity()):
-        return MemoryOps.boxed_store(context, value, tile_ref, css)
-    if css is not None:
-        coords = css.enumerate(access_tile.shape)
-        if cso is not None:
-            coords = [cso.eval(pt) for pt in coords]
-        return MemoryOps.store(context, value, tile_ref, coords=coords)
-    return MemoryOps.store(context, value, tile_ref)
+        sub_ref = MemoryOps._subtile_ref(tile_ref, css)
+        return MemoryOps.store(context, value, sub_ref)
+    coords = css.enumerate(access_tile.shape)
+    if cso is not None:
+        coords = [cso.eval(pt) for pt in coords]
+    return MemoryOps.store(context, value, tile_ref, coords=coords)
 
 
 # ---------------------------------------------------------------------------
@@ -538,15 +538,19 @@ def parse_construct_access_tile(op_text, parse_ctx: ParseContext):
         base_map_str = f'affine_map<({dims}) -> ({dims})>'
     base_map = parse_affine_map(base_map_str)
 
+    # ``access_tile_set`` is required per ODS
+    # (Builtin_IntegerSetAttr:$access_tile_set, no OptionalAttr); absence
+    # is invalid IR.  ``parse_affine_set`` lowers axis-aligned sets to
+    # ``BoxSet`` automatically, so the full-rectangle case naturally
+    # comes out as ``BoxSet([0, shape))`` and routes through the BoxSet
+    # fast path in the dialect handler — no normalisation needed.
     coord_set_str = attrs.get('access_tile_set')
-    coordinate_set = parse_affine_set(coord_set_str) if isinstance(coord_set_str, str) else None
-    # Default-fill an axis-aligned BoxSet covering the full rectangle when
-    # the attribute is absent.  Routing every rectangular access through
-    # a single ``coordinate_set`` shape (BoxSet) lets the BoxSet fast path
-    # in ktdp.load/store handle translated slices uniformly via
-    # ``_subtile_ref``, and lets ``BoxSet.is_full`` keep its strict
-    # ``[0, shape)`` semantics without having to special-case the absent
-    # attribute.  ``access_shape`` is structurally ``int`` (parsed from
+    if not isinstance(coord_set_str, str):
+        raise ValueError(
+            "construct_access_tile: access_tile_set is required (per ODS)"
+        )
+    coordinate_set = parse_affine_set(coord_set_str)
+    # ``access_shape`` is structurally ``int`` (parsed from
     # ``<NxMxindex>``); ``raise`` (not ``assert``) so the guard survives
     # ``python -O``.
     if not all(isinstance(s, int) for s in access_shape):
@@ -554,8 +558,6 @@ def parse_construct_access_tile(op_text, parse_ctx: ParseContext):
             f"construct_access_tile: access_shape must be concrete ints, "
             f"got {access_shape!r}"
         )
-    if coordinate_set is None:
-        coordinate_set = BoxSet(lo=(0,) * len(access_shape), hi=tuple(access_shape))
 
     coord_order_str = attrs.get('access_tile_order')
     coordinate_order = parse_affine_map(coord_order_str) if isinstance(coord_order_str, str) else None
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -193,8 +193,16 @@ def ktdp__load(op, context, env):
         return MemoryOps.distributed_load(
             context, access_tile.parent_ref, result_shape=result_shape
         )
-    css = access_tile.coordinate_set    # AffineSet | None
+    css = access_tile.coordinate_set    # BoxSet | AffineSet  (parser default-fills BoxSet)
     cso = access_tile.coordinate_order  # AffineMap | None
+    # BoxSet fast path: rectangular access with identity coordinate order →
+    # delegate to MemoryOps.boxed_load, which builds a sub-TileRef and lets
+    # MemoryOps.load take its own contiguous/strided fast path on the sub-ref.
+    # Box extent IS the iteration shape (the data tile shape contract is 1:1
+    # for BoxSet); non-rectangular AffineSet still routes through the
+    # coord-list path below where ``_result_shape`` reshapes the gather.
+    if isinstance(css, BoxSet) and (cso is None or cso.is_identity()):
+        return MemoryOps.boxed_load(context, access_tile.parent_ref, css)
     if css is not None:
         coords = css.enumerate(access_tile.shape)
         if cso is not None:
@@ -223,6 +231,12 @@ def ktdp__store(op, context, env):
     tile_ref = access_tile.parent_ref
     css = access_tile.coordinate_set
     cso = access_tile.coordinate_order
+    # BoxSet fast path: symmetric to ktdp.load. Slice the source tile
+    # rectangularly (no-op when box covers the full access tile) and store
+    # through a sub-TileRef. Avoids the read-modify-write scatter that the
+    # coord-list path needs for general AffineSet stores.
+    if isinstance(css, BoxSet) and (cso is None or cso.is_identity()):
+        return MemoryOps.boxed_store(context, value, tile_ref, css)
     if css is not None:
         coords = css.enumerate(access_tile.shape)
         if cso is not None:
@@ -526,12 +540,22 @@ def parse_construct_access_tile(op_text, parse_ctx: ParseContext):
 
     coord_set_str = attrs.get('access_tile_set')
     coordinate_set = parse_affine_set(coord_set_str) if isinstance(coord_set_str, str) else None
-    # Normalise to None when the set covers the full rectangular tile in
-    # row-major order — it carries no information beyond "load/store everything".
-    # This lets ktdp.load/store take the contiguous fast path instead of
-    # enumerating all coords on every execution.
-    if coordinate_set is not None and coordinate_set.is_full(access_shape):
-        coordinate_set = None
+    # Default-fill an axis-aligned BoxSet covering the full rectangle when
+    # the attribute is absent.  Routing every rectangular access through
+    # a single ``coordinate_set`` shape (BoxSet) lets the BoxSet fast path
+    # in ktdp.load/store handle translated slices uniformly via
+    # ``_subtile_ref``, and lets ``BoxSet.is_full`` keep its strict
+    # ``[0, shape)`` semantics without having to special-case the absent
+    # attribute.  ``access_shape`` is structurally ``int`` (parsed from
+    # ``<NxMxindex>``); ``raise`` (not ``assert``) so the guard survives
+    # ``python -O``.
+    if not all(isinstance(s, int) for s in access_shape):
+        raise TypeError(
+            f"construct_access_tile: access_shape must be concrete ints, "
+            f"got {access_shape!r}"
+        )
+    if coordinate_set is None:
+        coordinate_set = BoxSet(lo=(0,) * len(access_shape), hi=tuple(access_shape))
 
     coord_order_str = attrs.get('access_tile_order')
     coordinate_order = parse_affine_map(coord_order_str) if isinstance(coord_order_str, str) else None
--- a/ktir_cpu/mlir_frontend/parser.py
+++ b/ktir_cpu/mlir_frontend/parser.py
@@ -238,12 +238,20 @@ def _adapt_construct_access_tile(mlir_op, attributes, result_type, operands):
     canonical = str(AffineMapAttr(mlir_op.attributes["base_map"]).value)
     attributes["base_map"] = parse_affine_map(f"affine_map<{canonical}>")
 
-    # access_tile_set → coordinate_set; normalize full sets to None
-    if "access_tile_set" in mlir_op.attributes:
-        # str(IntegerSetAttr) → "affine_set<(d0) : ...>"
-        cs = parse_affine_set(str(mlir_op.attributes["access_tile_set"]))
-        if not cs.is_full(shape):
-            attributes["coordinate_set"] = cs
+    # access_tile_set → coordinate_set.  Required attribute per ODS
+    # (Builtin_IntegerSetAttr:$access_tile_set, no OptionalAttr); absence
+    # is invalid IR.  ``parse_affine_set`` lowers axis-aligned sets to
+    # ``BoxSet`` automatically, so the full-rectangle case naturally
+    # comes out as ``BoxSet([0, shape))`` and routes through the
+    # BoxSet fast path in the dialect handler — no normalisation needed.
+    if "access_tile_set" not in mlir_op.attributes:
+        raise ValueError(
+            "construct_access_tile: access_tile_set is required (per ODS)"
+        )
+    # str(IntegerSetAttr) → "affine_set<(d0) : ...>"
+    attributes["coordinate_set"] = parse_affine_set(
+        str(mlir_op.attributes["access_tile_set"])
+    )
 
     # access_tile_order → coordinate_order; normalize identity maps to None
     if "access_tile_order" in mlir_op.attributes:
--- a/ktir_cpu/ops/memory_ops.py
+++ b/ktir_cpu/ops/memory_ops.py
@@ -261,133 +261,6 @@ def _resolve_idx_reads(
     return per_view_values, total_sticks
 
 
-def _eval_subscript_vectorized(expr: tuple, pts: np.ndarray) -> np.ndarray:
-    """NumPy-vectorised dual of :func:`eval_subscript_expr`.
-
-    *pts* is an ``(N, n_vars)`` int array; the return is an ``(N,)`` int64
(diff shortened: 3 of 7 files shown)


WarningRan:
cc @fabianlim

fabianlim (changes requested):
## PR #79 Review — BoxSet rectangular fast paths

So after the PR the full picture is:

Point enumeration:
- AffineSet.enumerate — evaluates affine constraints per point, returns List[Tuple]
- BoxSet.enumerate — itertools.product over ranges, returns List[Tuple]
- _enumerate_in_vso_order — wraps AffineSet.enumerate / BoxSet.enumerate with vso permutation
- _enumerate_box_pts (new) — np.indices, returns ndarray

Subscript evaluation:
- eval_subscript_expr — per-point scalar, in ktdp_helpers.py
- _eval_subscript_vectorized (new) — per-column NumPy, mirrors the above tag-for-tag

Coord building:
- _build_indirect_coords — slow path, uses _enumerate_in_vso_order + eval_subscript_expr per point
- _build_indirect_coords_box (new) — fast path, uses _enumerate_box_pts + _eval_subscript_vectorized

**Context: two qualitatively different BoxSet wins**

The distributed path BoxSet fast path is an **O(N→1) win**: instead of enumerating every coordinate in `C_i`, computing per-element offsets, and scattering element-by-element, it builds a sub-`TileRef` from `lo`/`hi` directly and does one `MemoryOps.load` + one NumPy slice assignment. Enumeration is skipped entirely.

The indirect path BoxSet fast path (this PR) is a **constant-factor win on enumeration only**: `np.indices` builds the variable-space point grid all at once in C, replacing a Python `itertools.product` loop. Specifically, for shape `(N, M)`, `np.indices((N, M))` returns a `(2, N, M)` array where `[0]` holds the row index and `[1]` the column index at every position; `_enumerate_box_pts` reshapes this to `(N*M, 2)` and adds `box.lo` to shift from `[0, extent)` to `[lo, hi)`. The result is identical to `BoxSet.enumerate` via `itertools.product`, but built in C. However, the downstream `MemoryOps.load(coords=List[Tuple])` still iterates per coord — so the gain on the enumeration side is partially eaten by the `tolist()`/`tuple()` conversion at the boundary.

Issue #52 says "the variable-space enumeration becomes O(ndim) construction instead of O(∏shape)" — but this does not mean enumerate all points with `np.indices`. The O(ndim) hint is about using the box structure to build the **index-view address arrays** in O(ndim) array ops rather than a per-point Python loop:

1. **Range pull on index views**: when `vss` is a `BoxSet`, each indirect dim's `idx_exprs` is an affine function of the box axes. The full address array into the index view can be built with `np.arange` + broadcasting (one operation per axis, O(ndim)), rather than evaluating `eval_subscript_expr(e, pt)` per point.
2. **Direct/direct_expr dims are free**: they're axis ranges from the box — `np.arange(lo[d], hi[d])` meshgridded, no lookup needed.
3. **`read_scattered` gets a vectorised input**: once the address array is built with range ops, `read_scattered` receives a fully-formed address array. The reads themselves are still O(∏shape) (each index tensor element may sit on a different HBM stick), but the address computation collapses to O(ndim) array ops — which is what the issue meant.

The PR misses this: `_build_indirect_coords_box` still calls `_enumerate_box_pts` to materialise all points first, then evaluates subscripts per-column. The address array for `_resolve_idx_reads` is still built point-by-point inside `_resolve_idx_reads` itself (the PR doesn't touch that function). So the actual index-view reads are unchanged — only the coord-build side got the NumPy treatment, not the address-build side.

**Problem**: This PR adds fast paths overfitted to BoxSet without unifying the underlying primitives. 

Step by step:

1. `variables_space_set` (vss) is the iteration domain of the intermediate variables — the set of points `(v0, v1, ...)` enumerated to drive indirect index lookups. It is not the data tile or the index tensor.
2. When `vss` is a `BoxSet`, `_build_indirect_coords_box` speeds up the enumeration using `np.indices` as described above.
3. But `_eval_subscript_vectorized` — which evaluates dim subscript expressions across all points simultaneously — is a parallel copy of `eval_subscript_expr`. The slow path still calls `eval_subscript_expr` per point. Neither replaces the other.
4. Similarly, `_enumerate_box_pts` is a parallel copy of `BoxSet.enumerate` returning `ndarray` instead of `List[Tuple]`.
5. The right shape: make `_eval_subscript_vectorized` the single evaluator for both paths, and `_enumerate_box_pts` the single enumerator — the slow path batches all points into a single `ndarray` call, the BoxSet path adds nothing extra. As written, two grammars must be kept in sync.



---

### Issue 1: default-fill assumption in `ktdp_ops.py` is wrong

The PR replaces the `is_full` → `None` normalisation in the text parser (`ktdp_ops.py`) with a `BoxSet` default-fill when `access_tile_set` is absent. But `access_tile_set` is a **required** attribute in the ODS definition (`Builtin_IntegerSetAttr:$access_tile_set`, no `OptionalAttr`). Absent means invalid IR — the parser should raise, not silently default-fill.

The MLIR frontend parser (`parser.py`) still normalises full sets to `None` and leaves absent as `None`, so the two parsers now produce different `coordinate_set` values for the same op. The `isinstance(css, BoxSet)` fast path never fires for ops parsed through `parser.py`.

The clean fix:
- Both parsers: raise if `access_tile_set` is absent; drop the `is_full` → `None` normalisation.
- `parse_affine_set` already calls `BoxSet.try_from_affine_set` internally, so axis-aligned sets (including full rectangles) come out as `BoxSet` automatically. No default-fill needed anywhere.
- Execution handler: drop the `css is None` branch — it can never happen once both parsers enforce presence.

The `isinstance(css, BoxSet)` fast path then works for free, including the full-rectangle case.

---

### Issue 2: `distributed_load`/`distributed_store` don't use the new helpers

`boxed_load`/`boxed_store` were introduced to avoid duplicating the `_subtile_ref` + `MemoryOps.load` delegation, but `distributed_load` and `distributed_store` still inline that same logic. The helpers don't save any lines relative to the distributed path — they just add a third copy. Either retrofit the distributed path to call them, or drop the helpers and inline the same pattern in the `ktdp__load`/`ktdp__store` dispatch directly.

---

### Issue 3: `_eval_subscript_vectorized` / `_build_indirect_coords_box` are ~100 lines of parallel code

`_eval_subscript_vectorized` mirrors `eval_subscript_expr` tag-for-tag, and `_build_indirect_coords_box` mirrors the slow path in `_build_indirect_coords`. Neither replaces the other — both exist side-by-side. The vectorised evaluator is only called from the BoxSet fast path; the slow path still calls `eval_subscript_expr` per point.

The ~100 lines are only justified if `_eval_subscript_vectorized` becomes the single evaluator for both paths — i.e. migrate `_build_indirect_coords` to call it directly (passing a single-row `pts` array per point, or restructuring to batch all points). That way the BoxSet path just adds `_enumerate_box_pts` on top, and the slow path stops duplicating the expression grammar. As written, the vectorised evaluator is a constant-factor win on the coord-build side only, with the gain partially eaten by the `tolist()`/`tuple()` conversion before `MemoryOps.load` (which still consumes `List[Tuple]`). No benchmark numbers are provided to justify the maintenance cost.

### Issue 4: `vso` guard is overly conservative (but intentional per issue spec)

The fast path fires only when `vso is None or vso.is_identity()`. Issue #52 explicitly scoped non-identity `vso` out: "the box may not stay axis-aligned after permutation, so these stay on the slow path." However, a permuted box is still a box — `_build_indirect_coords_box` could handle non-identity `vso` by pre-permuting `lo`/`hi` before calling `np.indices`, without any loss of the rectangular structure. The issue's reasoning is incorrect on this point.

### Issue 5: no benchmark numbers

The PR description explicitly says "Constant-factor savings on the Python-side coord build for the indirect path; no measured benchmarks gathered." The speedup is asserted but unquantified. Given that the `tolist()`/`tuple()` conversion at the output boundary partially eats the NumPy gain (acknowledged in the PR description as the remaining bottleneck), benchmarks are needed to justify ~100 lines of parallel implementation.

---

### Minor

- `_eval_subscript_vectorized` has a `"sym"` branch that raises. But `"ssa"` nodes are pre-resolved to `("const", v)` by `_resolve_node` before any evaluator is called — there is no `"sym"` tag in the subscript grammar at evaluation time. The branch is dead code and can be removed.

---

WarningRan:
Round 2 update — addressing all review issues in this comment rather than per-thread replies, so the bench analysis (Issue 5) reads as one continuous narrative. Commit hash for round-2: `<commit-hash>`. The PR description has also been updated with a `Round 2 summary` banner at the top; the original description is preserved beneath it for review-thread context.

Quick disposition map: Issue 1 fixed, Issue 2 fixed, Issue 3 reverted, Issue 4 deferred per discussion, Issue 5 benched + analysed below, Minor resolved by the Issue 3 revert.

## Issue 1 — parser asymmetry

Acknowledged — this was a critical bug; the round-1 default-fill made the BoxSet branch invisible to the MLIR frontend path. Fixed following your suggested approach exactly: both the text parser (`ktdp_ops.py`) and the MLIR frontend parser (`mlir_frontend/parser.py`) raise if `access_tile_set` is absent (it's required per ODS); both drop the `is_full → None` normalisation and rely on `parse_affine_set` to lower axis-aligned sets to `BoxSet` upstream; the dialect handler drops the `css is None` branch — unreachable now that both parsers enforce presence. `AccessTile.coordinate_set` is always populated post-parse.

## Issue 2 — `boxed_load` / `boxed_store` add a third copy

Took option 2 (drop the helpers, inline the same pattern in the dispatch). `MemoryOps.boxed_load` / `boxed_store` are removed; the `ktdp.load` / `ktdp.store` BoxSet branches now do `MemoryOps._subtile_ref(parent, css)` + `MemoryOps.load(context, sub_ref)` inline — the same two lines `distributed_load` / `distributed_store` use for their per-partition BoxSet branch. No third copy. Tests in `test_boxset_fastpath.py::TestSubTileRefShortcut` were adjusted to call `_subtile_ref` + `MemoryOps.load` / `store` directly (same composition as the dispatch), preserving unit-level coverage of col-packed no-trample and empty-box cases without a wrapper layer.

## Issue 3 — ~100 lines of parallel grammar

Fully reverted: `_eval_subscript_vectorized`, `_build_indirect_coords_box`, `_enumerate_box_pts` are all gone; `_build_indirect_coords` is back to its pre-PR per-point implementation. The corresponding test class (`TestIndirectBoxFastPathConsistency`) is removed. The right shape you described — single evaluator servicing both slow and fast paths via array-shaped pts — is the correct fix and we agree it should land. Tracked as a follow-up issue (drafted, to be filed after this PR merges); see also the related O(ndim) address-array work captured at the same time. The indirect path is unchanged in this PR.

## Issue 4 — `vso` guard overly conservative

Agreed — issue #52's reasoning here is wrong; a permuted box is still a box, and pre-permuting `lo` / `hi` before constructing the variable-space iteration preserves the rectangular structure. Per our discussion this is deferred to a follow-up rather than folded into this PR. The follow-up issue is drafted and will be filed once this PR merges (so it doesn't dangle on an unmerged ref).

## Issue 5 — benchmark numbers (and what they mean)

Bench done. The result reframes what this PR delivers — 0% speedup on realistic fixtures, with the reason being structural rather than a missed implementation.

**Methodology.** `bench_issue52.py` runs `KTIRInterpreter.execute_function` with `time.perf_counter()`, N=10–20 iters after 3–5 warmup, median reported. Baseline = `e99f234` (tip of `main` pre-PR-79); after = current HEAD. Same machine, same Python (`.venv`).

| Fixture | main (median) | round-2 (median) | speedup |
|---|---|---|---|
| `matmul_small` (16×64 × 64×64) | 6.15 ms | 6.21 ms | 1.00x |
| `matmul_kernel` (64×8192×2048) | 20763.81 ms | 20763.79 ms | 1.00x |
| `softmax_kernel` (4096×1024) | 3132.14 ms | 3161.75 ms | 0.99x |

**Why 0% speedup — structural analysis.** `MemoryOps.load` has one optimised internal path: `_is_contiguous(shape, strides)` triggers a one-shot `mgr.read(N)` + `reshape`; otherwise it falls through to per-coord `_flat_memory_offsets`. Both `main` and round-2 reach the same path on every realistic fixture.

On `main`, the `is_full → None` sentinel routes "full access tile set" requests to `MemoryOps.load(parent_ref)` with no coords; `construct_access_tile`'s base_ptr offset shift makes `parent_ref.shape == access_tile_shape`, `_is_contiguous` returns True, the contig fast path triggers naturally. On round-2, `_subtile_ref(parent, BoxSet([0, shape)))` produces a `sub_ref` mathematically identical to `main`'s `parent_ref` — same shape, strides, base_ptr — reaching the same contig fast path. Spy verification on `softmax_kernel`: 8192 contig hits on both sides, 0 fall-throughs to `_flat_memory_offsets`.

For the rare "non-full BoxSet" case (e.g. `[0, 4) × [0, 8)` inside an 8×16 access tile), `_subtile_ref` produces a strided sub-tile (parent strides ≠ box-extent strides), `_is_contiguous` fails, both sides fall through to `_flat_memory_offsets`. Across all 6 (`coordinate_set` × `coordinate_order`) combinations on the direct path, `main` and round-2 reach equivalent code paths — the 0% delta is structural.

The actual fast path that would help — strided sub-rectangle reads via `np.lib.stride_tricks.as_strided` + `.copy()`, skipping per-coord `_flat_memory_offsets` — needs to live inside `MemoryOps.load` itself. That's a callee-side change, structurally distinct from this PR's caller-side cleanup. Tracked as a follow-up.

**Strategy validation — distributed cross-check.** To confirm the BoxSet strategy delivers when its target bottleneck exists, ran a `distributed_tile_access` micro-bench on the same fixture pattern as `tests/test_distributed_view.py::test_distributed_tile_access_*` (256×512 distributed view, 4 row-band partitions of 64×512, access shape 32×128, indices `(50, 64)` — the cross-partition-boundary case from the existing test fixture; N=10 iterations after 3 warmup, median reported):

```
BoxSet  (parse_affine_set):       median 0.020 ms
AffineSet (parse_affine_set_raw): median 803.170 ms
Speedup: 40,453x
```

Same order of magnitude as the "~58000×" cited in PR #69. The win on distributed comes from O(ndim) intersect (`B_i.intersect(xA_box)`) vs O(∏shape) enumerate during partition selection, plus `out[slc] = tile.data` slice-assign vs per-coord output-buffer writes during load-side assembly. Single-allocation `ktdp.load` doesn't have either step — no partition selection, no output buffer to assemble — so the strategy can't carry that win over architecturally.

**Reframing.** Round-2 is architectural cleanup (parser symmetry, single strategy across distributed and direct paths, removal of duplicate code) rather than a runtime speedup. The PR title and description still reflect round-1's "fast path" framing — happy to update them to match this cleanup-only positioning in a follow-up commit, but left them in round-1 form for now since the reframing is significant and we wanted your read on this analysis first. The `MemoryOps.load` strided-sub-rectangle fast path is the natural follow-up for the actual perf work.

## Minor — `"sym"` dead branch

Resolved by the Issue 3 revert — the whole `_eval_subscript_vectorized` function (containing the dead `"sym"` branch) is gone.

cc @fabianlim

fabianlim:
**Round-2 review — cleanup verified, follow-ups filed**

Reviewed the round-2 changes against the code. The architectural reframing is accurate and the cleanup is sound.

**Completed / verified:**
- **Issue 1 (parser asymmetry)** — fixed. Both `ktdp_ops.py` and `mlir_frontend/parser.py` now raise on absent `access_tile_set` (required per ODS) and rely on `parse_affine_set` to lower axis-aligned sets to `BoxSet`. The `css is None` branch is gone from the handler. This was a real frontend correctness fix, not just cosmetics.
- **Issue 2 (third copy)** — resolved. Helpers dropped; `ktdp.load`/`store` inline `_subtile_ref` + `MemoryOps.load`/`store`, matching the distributed path.
- **Issue 3 (parallel grammar)** — fully reverted; confirmed `_eval_subscript_vectorized`, `_build_indirect_coords_box`, `_enumerate_box_pts` are gone and `_build_indirect_coords` is back to the per-point form. Indirect path untouched.
- **Issue 5 (0% speedup)** — analysis confirmed against the code. Both `main` and round-2 reach the contiguous fast path in `MemoryOps.load` for full-rectangle access; `_subtile_ref(box[0, shape))` produces a sub_ref identical to `main`'s `parent_ref`. The 0% is structural, as you describe.

**On the "fast path" framing:** agreed this PR is caller-side unification + a parser fix, runtime-neutral by design. Suggest retitling away from "fast path" before merge so the history reflects what landed.

**#52 disposition:** since the actual fast-path scope of #52 was reverted/deferred, we're closing #52 as superseded-by-reframing and tracking the real work as fresh issues:
- #95 — strided sub-rectangle fast path in `MemoryOps.load`/`store` (the genuine perf follow-up; callee-side, lights up both direct + distributed once landed).
- #96 — preserve BoxSet structure under permuted vso (Issue 4; depends on the indirect BoxSet path being reinstated first).
- #97 — `MemoryOps.load`/`store` accept ndarray coords (the `tolist()`/`tuple()` overhead).

All three assigned to @WarningRan.

One note on #95: this PR doesn't build toward the strided fast path, but it does cleanly unblock it — the callee-side change in `MemoryOps.load` will benefit both direct and distributed callers at once thanks to the unified `_subtile_ref` routing. That's the concrete payoff of the unification.

WarningRan:
Restored title to your edit — overwrote inadvertently.

fabianlim (approved):
approving this issue for now. Please move on to #98 which addresses the vectorized indirect access routine that was reverted. after that move on to #95 , #96 #97

WarningRan:
Anything blocks this PR now? Thanks! @fabianlim

fabianlim:
@WarningRan sorry over looked this. Just triggered the CI and will merge if that passes

PR #80, opened 2026-06-03
[Test] e2e MLIR coverage for symbolic-BoxSet distributed memory views

### LINKED ISSUE #74: [Test] End-to-end MLIR test gap for symbolic distributed views
## Background

PR #69 added symbolic-`BoxSet` support so distributed memory views with `?`-shape can stay on the `distributed_tile_access` fast path. The PR's tests cover the change at three layers:

- AST helpers and `BoxSet` round-trips (`tests/test_ast.py`, `tests/test_affine.py`).
- Distributed-view geometry on parameterised symbolic partitions (`tests/test_distributed_view.py`).
- Dialect handler boundary in isolation, including the trailing-`?` regression guard for `memref<64x?xf16>` (`tests/test_dialects_exec.py::TestKtdp::test_construct_memory_view_specializes_symbolic_coord_set_*`).

What's missing is a single test that drives the **full path**: parse a `?`-shape distributed view from MLIR text, execute through the dialect handlers all the way to `ktdp.load`, and check the loaded values against a reference. Today's tests stop either at the ops layer (with `access_tile_set=None`) or at the handler boundary in isolation — nothing exercises the dynamic-shape lowering across the full execution stack.

## Why it matters

- The existing distributed-view example, [`examples/rfc/distributed-view-copy.mlir`](examples/rfc/distributed-view-copy.mlir), uses fully concrete partition extents — the dynamic-shape lowering path it was designed to test isn't actually exercised by it.
- Without an e2e fixture, a regression in the regex parser → handler → `distributed_tile_access` → `ktdp.load` chain on symbolic input could land silently. The unit-level tests in #69 catch a misbinding inside the handler in isolation, but cannot catch a wiring bug between the regex parser and the handler, or between the handler and the load path.

## Scope

1. **Add a `?`-shape distributed-view example** under `examples/rfc/` (e.g. `distributed-view-copy-dynamic.mlir`), modeled on `distributed-view-copy.mlir` but with at least one partition's coordinate set carrying a symbolic dim sourced from the view's `sizes:` operands.
2. **Wire it into `tests/test_examples.py`** alongside `TestVectorAddDynamicExecution` — execute via the regex parser path and check results against a NumPy reference.
3. **Cover the `memref<64x?xf16>` trailing-dynamic shape** so the symbol-binding regression — that #69's handler test guards in isolation — is also caught at the e2e layer if the parser ever feeds the wrong operand list down.

## Out of scope

- Adapt path (`tests/mlir_frontend/test_examples_adapt.py`) coverage. `MLIRFrontendParser` has no handler for `ktdp.construct_distributed_memory_view` (raises `NotImplementedError`) and its `construct_memory_view` handler doesn't resolve dynamic dims from the `sizes:` operand list — same structural reason `TestVectorAddDynamicExecution` has no adapt sibling. Extending the frontend parser to cover distributed/dynamic is separate capability work, not this test-coverage PR.
- Performance benchmarking. The `~58,000×` / `~315×` numbers from #51 are already covered conceptually; this issue is about correctness coverage on the symbolic e2e path.

## Acceptance criteria

- The new dynamic-shape example runs end-to-end through the regex parser via `tests/test_examples.py`.
- Output matches a NumPy reference computed with the same logical partitioning at concrete symbol values.
- The `memref<64x?xf16>` trailing-dynamic shape is covered, so the e2e test would fail if the symbol-source bug from #69 (binding `s_0` to a static dim) regressed.
- Existing `examples/rfc/distributed-view-copy.mlir` continues to pass unchanged.

## References

- Parent: #51
- Predecessor PR: #69 (in-scope half of #51)
- Related: #55 (open) tracks wiring the concrete `distributed-view-copy.mlir` into `test_examples.py` + `test_examples_adapt.py`.

---

WarningRan:
## Summary

Covers the `BoxSet` path of #74 with an end-to-end test that drives a symbolic-`BoxSet` distributed view through the full execution stack:

```
regex parser -> construct_memory_view (symbol binding)
             -> construct_distributed_memory_view (composition)
             -> construct_access_tile
             -> distributed_tile_access (per-partition routing)
             -> ktdp.load / ktdp.store
```

The `AffineSet` path remains under #74 until a downstream caller surfaces.

PR #69 added the symbolic-`BoxSet` capability; existing tests cover it at the ops layer (`tests/test_distributed_view.py::test_distributed_tile_access_dynamic_shape_emits_box_set`, `access_tile_set=None`) and at the dialect handler in isolation (`tests/test_dialects_exec.py::TestKtdp::test_construct_memory_view_specializes_symbolic_coord_set_*`). No test exercised the full path on symbolic input — a regression in any link of that chain could land silently.

## What changes

- `examples/rfc/distributed-view-copy-dynamic.mlir` — two HBM partitions split a logical 64×26 tensor at column `s0`. Each partition is `memref<64x?xf16>` with `sizes: [64, %s0]`; coordinate sets reference symbolic `s0` and `2*s0`.
- `TestDistributedViewCopyDynamicExecution(InterpreterTestMixin)` in `tests/test_examples.py` — drives the fixture end-to-end and compares the result bit-exact against the NumPy source.
- `distributed_view_copy_dynamic` entry in `tests/conftest.py` (`s0_in: 13`).

## Coverage

- **Trailing-`?` on the partition memref** pins the regression class fixed in #69: a static leading dim must not steal symbol index 0 from the dynamic tail dim.
- **Cross-partition access tile** ensures `distributed_tile_access` per-partition routing is actually walked (single-partition coverage would short-circuit it).
- **Bit-exact comparison** via `assert_array_equal(result, a_full)` — `a_full` is the source the partitions are sliced from, so partition routing or symbol-binding errors all surface as element-wise mismatch.

## Design notes

The global view and output `B` are concrete `memref<64x26xf16>` because the regex parser does not accept `?` in `construct_distributed_memory_view`'s result type, and `construct_access_tile` requires a concrete `access_tile_set`. Symbolic binding is still exercised on every `construct_memory_view` call for the partitions — exactly where #69's bug class lives.

Two intentional differences from the existing concrete `examples/rfc/distributed-view-copy.mlir` template, anchored here for reviewers using it as a comparison point:

1. **Per-buffer pointers via `func.func` arguments**, not `arith.constant <addr>`. Matches the existing dynamic precedent `vector_add_dynamic_ktir.mlir` and keeps the orthogonal `base_ptr` stick-index vs element-index concern out of scope.
2. **Random-seeded source compared bit-exact against itself**: `a_full = randn(64, 2*s0); a0 = ascontiguousarray(a_full[:, :s0]); a1 = ascontiguousarray(a_full[:, s0:]); assert_array_equal(result, a_full)`. Each partition lives in its own NumPy buffer, so partition routing errors surface as either out-of-bounds reads or full-array mismatch — disjoint per-partition fingerprints would only have been load-bearing under a shared HBM buffer.

Supersedes the approach floated in #74's earlier comment; mirrors `TestVectorAddDynamicExecution` instead.

## Out of scope

- Adapter coverage in `tests/mlir_frontend/test_examples_adapt.py`. `MLIRFrontendParser` registers no handler for `ktdp.construct_distributed_memory_view`, and `_adapt_construct_memory_view` does not resolve dynamic dims from the `sizes:` operand list. Same structural reason `TestVectorAddDynamicExecution` has no adapt sibling. Extending the frontend parser is independent capability work.
- Performance benchmarking (already covered conceptually by #51).

## Test plan

- Targeted bug injection at `ktir_cpu/dialects/ktdp_ops.py:94` (flipping the binding filter `isinstance(s, str)` → `isinstance(s, int)` so static dims would steal the symbol slot) is caught by both the new test (50% mismatch on B) and `TestVectorAddDynamicExecution` (4 failures total). Reverting brings both back to green.


WarningRan:
cc @fabianlim

fabianlim:
## Review

Verified the test against current `main` (with #86 and #89 merged): it's additive, the rebase is clean (`conftest.py`/`test_examples.py` are untouched by intervening PRs), and it passes. It also genuinely catches the #69 binding regression — injecting the `isinstance(s, str)→int` flip at `ktdp_ops.py:94` fails the test. No conflict with #86/#89: the op stays on the regex path, and #89's guard already allowlists `construct_distributed_memory_view` (issue #87). 

One substantive gap worth addressing, though — the coverage is narrower than the description claims.

### The `s0 = 13` design is degenerate

The fixture pins `s0 = 13` so that a single full-domain `64x26` access tile (`#full`, which carries no `s0`) coincides exactly with the global domain (`2*s0 = 26`). Because the tile equals the whole domain, `distributed_tile_access` returns **both** partitions on the one and only offset — so the per-partition routing/intersection (`C_i = B_i ∩ (x+A)`) is only ever walked in the trivial "tile covers everything, split exactly in half" configuration. The "Cross-partition access tile ensures routing is actually walked" claim doesn't really hold: routing never has to *exclude* a partition.

It's also conflating two unrelated constraints. The comment says `s0` is fixed *because* `construct_access_tile` requires a concrete `access_tile_set` — but `#full` is already concrete (no symbol), so that requirement doesn't constrain `s0` at all. The real reason is just data coverage: the tile is hard-coded to `64x26`, which only matches the partitions when `2*s0 = 26`. Nothing in the fixture scales with `s0`, which is the tell that it's testing one frozen instance of the dynamic path rather than the dynamic path.

### Suggested fix: fixed tile + iterated offsets, let `s0` vary

A concrete access tile is fine — the trick is to keep the **tile** fixed and let the **offset count** scale with `s0`, so the fixture works for a range of `s0` (and drop the odd `26` for powers of two):

- tile width = **32** (fixed, concrete), rows 64
- parametrize `s0 ∈ {16, 32}`, global width `2*s0`, offset loop trip count `2*s0 / 32`

This exercises both routing modes in one fixture:
- `s0 = 16` (global 32): one offset `[0,0]`, the 32-wide tile **straddles** both 16-wide partitions → non-trivial `C_i = B_i ∩ (x+A)` split on each side.
- `s0 = 32` (global 64): two offsets `[0,0]`,`[0,32]`, each tile lands **inside one partition** → single-partition routing, the other partition correctly excluded.

`scf.for` is supported on the regex path (`scf_ops.py`), and a single tile spanning multiple partitions in one `load` is already the pattern in the concrete `distributed-view-copy.mlir` (`192x64` tile over 3 partitions), so both shapes are valid against the current engine. Constructing a fresh `construct_access_tile` per iteration with `%off` as the index keeps `access_tile_set` concrete.

Happy to pair on the loop form if useful. 

### Adapter coverage is a tracked follow-up, not a blocker

This test runs on the **regex** parser, where `construct_distributed_memory_view` is hooked up (`ktdp_ops.py:408`); the **MLIR frontend adapter** has no handler for it yet, so there is intentionally no `...Adapt` sibling here. That gap is already tracked by #89's guard (allowlisted with a reason, pointing at #87). So this PR doesn't need to block on #87 — once #87 lands the frontend handler, it removes the allowlist entry and adds the `TestDistributedViewCopyDynamicAdapt` subclass reusing this exact fixture, and we'll have distributed-memory-view coverage on the MLIR-parser path as well. #89's `test_unsupported_allowlist_has_no_stale_entries` will force that to happen, so it can't be silently forgotten.

fabianlim (changes requested):
the symolic tests have to accomodate for a variety of values for s0

WarningRan:
Thank you for the reviews! @fabianlim 
The fixture now keeps the access tile concrete at 64x32 and varies the offset count via `scf.for %off = %c0 to %ub step %c32`, so the per-partition routing/intersection actually has to do work:

- **s0 = 16** (global = 32): one offset, the 32-wide tile straddles both 16-wide partitions — non-trivial `C_i = B_i ∩ (x+A)` split on each side.
- **s0 = 32** (global = 64): two offsets, each tile lands inside one partition — single-partition routing with the other partition correctly excluded.

Verified locally that all three branches are walked (added a temporary print in `distributed_tile_access`, removed before commit):

- s16 offset 0 → `survivors=2 excluded=[]`
- s32 offset 0 → `survivors=1 excluded=[1]`
- s32 offset 32 → `survivors=1 excluded=[0]`

One deviation from your suggestion to mention: `parametrize s0 ∈ {16, 32}` ended up as **two .mlir files** rather than one, because the regex parser (`ktdp_ops.py:471`, `int(p)`) requires the `construct_distributed_memory_view` result type to be a concrete shape — `2*s0` cannot be `?` there. Two fixtures keep the parser path unchanged. Happy to revisit if you'd prefer a single fixture once the parser accepts dynamic distributed-view result types.

Adapter sibling stays out of scope of this PR — `construct_distributed_memory_view` is on #89's allowlist pointing at #87, so when #87 lands the `TestDistributedViewCopyDynamicAdapt` subclass goes in there reusing this fixture.

Both fixtures fail the `ktdp_ops.py:94` `isinstance(s, str)→int` injection (`KeyError: "Value '64' not found"` on `construct_memory_view`); reverting brings them back to green.

fabianlim:
@WarningRan the workaround you noted in the design notes (regex parser rejecting `?` in the `construct_distributed_memory_view` result type) — is that gap tracked anywhere? Note that this limitation is specific to the regex parser in `ktir_cpu/dialects/ktdp_ops.py`; the MLIR frontend ([ktir-mlir-frontend](https://github.com/torch-spyre/ktir-mlir-frontend)) has no such restriction — its TableGen definition uses `AnyMemRef` and the verifier only checks rank and element type. Issue #87 is about the `MLIRFrontendParser` adapter layer, which is a separate gap. Worth opening a dedicated issue for the regex parser fix, or are you planning to address it here?

WarningRan:
@fabianlim Good catch — you're right that I conflated the two. The `int(p)` limitation is on ktir-cpu's regex parser specifically (`ktdp_ops.py:471`), not on the dialect — `AnyMemRef` in the upstream TableGen makes the dynamic-shape distributed-view-result form fully legal at the spec level. And #87 is the `MLIRFrontendParser` adapter gap, separate from this.

Filing a dedicated issue for the regex parser fix and keeping it out of #80 — this PR stays a test-only change. The fix itself looks low-risk: `construct_memory_view`'s parser already handles `?` (`ktdp_ops.py:339-382` — parses `?` as `None`, then resolves from `sizes:`); the same pattern carries over to `construct_distributed_memory_view`, just resolving from each operand memref's shape instead of a `sizes:` attribute. Once that lands, the two-fixture split here collapses to one parametrized fixture, which would be a clean regression test that the parser fix actually works.

I'll link the new issue back here once filed and update the design-notes section accordingly.

WarningRan:
Filed as #99.

fabianlim:
> Filing a dedicated issue for the regex parser fix and keeping it out of https://github.com/torch-spyre/ktir-cpu/pull/80 — this PR stays a test-only change. The fix itself looks low-risk: construct_memory_view's parser already handles ? (ktdp_ops.py:339-382 — parses ? as None, then resolves from sizes:); the same pattern carries over to construct_distributed_memory_view, just resolving from each operand memref's shape instead of a sizes: attribute. Once that lands, the two-fixture split here collapses to one parametrized fixture, which would be a clean regression test that the parser fix actually works.

ok then we need to hold off this PR until #99 is addressed

WarningRan:
Implementation resolves dynamic axes from `partition.coordinate_set.hi` rather than operand memref shapes — see #111 for the s0/2·s0 counterexample. Scope step 2 (resolve from operand shapes) and step 3 (ambiguous-across-operands raise) are superseded.

lasch (approved):
Consider it approved unless @fabianlim or others find holes in the summary below:
(I can provide the long version of this if requested)

## Short Summary

- **Verification performed**
  - Ran the touched suite (8 passed) and full suite (1122 passed, no regressions) against the PR head.
  - Reproduced the PR's own bug-injection claim; both new tests failed exactly as described, then reverted.
  - Confirmed the two-fixture split was a genuine regex-parser limitation, not an authoring shortcut, at merge time.

- **Code assessment**
  - Test correctly detects the #69 binding-filter regression via bit-exact output comparison across partitions.
  - Original reviewer's routing-triviality concern was substantively fixed three weeks ago by commit `5061685`.
  - PR scope is honestly bounded, excluding frontend-adapter and benchmarking coverage for structurally sound reasons.
  - Pure test/fixture addition — no interpreter code touched, so no aliasing, lifetime, or bookkeeping side effects.

- **Spec / RFC grounding (RFC 0682)**
  - Test-only change; adds no ops or attributes, so no KTIR-spec semantics are touched or altered.
  - Fixture's two-partition layout faithfully models RFC 0682's distributed-view composition across multi-core hardware.
  - Separation of concerns (memory view, access tile, load/store) stays intact throughout the added fixture.
  - Concrete-shape workaround is correctly identified as a parser limitation, not a dialect-level spec constraint.

- **Testing gaps**
  - Two-fixture split is now removable since dependency PR #111 merged today, enabling dynamic result shapes.
  - No negative/error-path coverage for symbolic binding, but this matches existing sibling test conventions.

- **Verdict**
  - Approve pending fresh review; sole blocker is a stale `CHANGES_REQUESTED` predating the fix that resolved it.
  - Fixture collapse is a worthwhile follow-up now that #111 landed, but it should not block merging this PR.

WarningRan:
Still has leftover to do in this PR

examples/rfc/distributed-view-copy-dynamic-s16.mlir
```diff
@@ -0,0 +1,85 @@
...
+
+    %A1_view = ktdp.construct_memory_view %a1_ptr, sizes: [64, %s0], strides: [%s0, 1] {
+      coordinate_set = #A1_set,
+      memory_space = #ktdp.spyre_memory_space<HBM>
+    } : memref<64x?xf16>
+
+    // (1) Compose into a single logical 64x32 distributed view.
+    %A_view = ktdp.construct_distributed_memory_view
+        (%A0_view, %A1_view : memref<64x?xf16>, memref<64x?xf16>)
+        : memref<64x32xf16>
```
fabianlim: @WarningRan @lasch the main problem i have now is that this output memref cannot be static if the inputs are dynamic.

fabianlim: it can be resolved to static only after the access tile is created, but not at the memory view. We should review some of the other examples (i.e., the non-distributed ones) in how they handle dynamic shapes at the mem view + access tile level

WarningRan: Addressed - the composed distributed view and the output B are now dynamic memref<64x?xf16> with the concrete 64x32 appearing only at the access-tile level (mirroring the non-distributed dynamic examples), which let the two static fixtures collapse into a single ?-shape fixture parametrized over s0 ∈ {16, 32}.

fabianlim:
@WarningRan Optional: this example merges only on the symbolic dimension (the dim=1). Do we also want to have another example, just for thoroughness, to merge on the non-symbolic dimension. THat is the tile sizes are say [64, S0] and [64, S0], and we get something that is [128, S0]?

WarningRan:
Added another example, thank you!

Issue #82, opened 2026-06-04
[Bug] parse_tensor_type loses element types whose name contains 'x' (index, complex)


kcirred:
## Summary

  The regex parser's `parse_tensor_type` helper splits the inner text of a
  tensor type on the literal character `'x'`. Element type names that
  themselves contain `'x'` — `index` and `complex<...>` — are split mid-word,
  so the dtype field of the returned dict comes back as `""` instead of
  the real element type. Any KTIR that mentions `tensor<NxMx...xindex>` —
  notably the shape operand of `tensor.reshape` produced by lowering
  `tt.reshape` — silently parses with a wrong (empty) dtype.

## Bug — element-type names that contain `'x'` are mis-tokenised

  `parse_tensor_type` matches `tensor<...>`, takes the inner text, and
  splits it on `'x'`. It then takes everything except the last segment
  as the shape (filtering for `isdigit()`) and the last segment as the
  dtype. For a tensor whose element type name does not contain `'x'`
  (`f16`, `f32`, `i32`, `bf16`, …) this works — splitting `"256xf16"`
  gives `["256", "f16"]` and the last segment is the right dtype.

  For `index` it does not. Splitting `"2xindex"` on `'x'` yields three
  segments — `["2", "inde", ""]` — because `index` itself contains `x`:

  "2xindex".split("x")  ==  ["2", "inde", ""]

  The shape filter keeps only the all-digit segments, so it correctly
  recovers `(2,)`. But the dtype is `parts[-1] == ""`. The function
  returns:

  ```python
  {"shape": (2,), "dtype": ""}

fabianlim:
closed with merging of #83

PR #83, opened 2026-06-04
Fix parse_tensor_type for element types containing 'x'

kcirred:
The regex parser's `parse_tensor_type` helper splits the inner text of
  a tensor type on the literal character `'x'`, which silently corrupts
  element types whose name itself contains `'x'` — `index` and
  `complex<...>`. For `tensor<2xindex>`, `"2xindex".split("x")` yields
  `["2", "inde", ""]`, so the helper returned `{"shape": (2,), "dtype":
  ""}` with an empty dtype. Downstream `arith.constant` and
  `tensor.reshape` parsers consume that wrong dtype either as `""` or as
  the silent `f16` fallback in `arith_ops.py`, so any KTIR carrying an
  `index`-typed shape operand parsed wrong.

The fix walks the dimension prefix one `(\d+)x` match at a time and
  takes the remainder as the dtype — element type names cannot start
  with a digit followed by `x`, so the loop terminates at the right
  boundary regardless of which letters appear inside the dtype. Rank-0
  behaviour is preserved (`tensor<f32>` still returns `None`), keeping
  existing call sites unchanged.                                                               

  Tests in `tests/test_parser_utils.py` pin three groups: ordinary
  float/integer dtypes across rank 1–4 (9 cases), the `index` dtype
  regression at rank 1, 2, and 3 (4 cases), and rejection of non-tensor
  / malformed / rank-0 inputs (6 cases). Full ktir-cpu suite passes
  (914 / 7 skipped / 7 xfailed) with no regressions.

Addresses Issue #82 

cc: @fabianlim

### DIFF
--- a/ktir_cpu/parser_utils.py
+++ b/ktir_cpu/parser_utils.py
@@ -60,14 +60,18 @@ def parse_tensor_type(type_str: str) -> Optional[Dict]:
         {"shape": tuple, "dtype": str} if tensor type, else None
     """
     tensor_match = re.match(r'tensor<([^>]+)>', type_str)
-    if tensor_match:
-        inner = tensor_match.group(1)
-        parts = inner.split('x')
-        if len(parts) >= 2:
-            shape = tuple(int(p) for p in parts[:-1] if p.isdigit())
-            dtype = parts[-1]
-            return {"shape": shape, "dtype": dtype}
-    return None
+    if not tensor_match:
+        return None
+    inner = tensor_match.group(1)
+    shape: list = []
+    pos = 0
+    while (m := re.match(r'(\d+)x', inner[pos:])):
+        shape.append(int(m.group(1)))
+        pos += m.end()
+    dtype = inner[pos:]
+    if not shape or not dtype:
+        return None
+    return {"shape": tuple(shape), "dtype": dtype}
 
 def parse_numeric(s: str, dtype: Optional[str] = None) -> Any:
     """Parse a numeric string to a Python int or float.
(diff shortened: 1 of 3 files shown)


fabianlim:
The while loop exists solely to handle element types whose names contain "x" (like index, complex<...>). The old "2xindex".split("x") produces ["2", "inde", ""] — it shreds the dtype. The loop peels off dimension tokens one \d+x match at a time and takes the remainder as dtype.

One note: the same could be done without a loop. A single regex like r"(\d+(?:x\d+)*)x([^x]+)" captures the full shape dims (minus the trailing x) and the dtype in one match. This is preferred against having a while loop, which creates nervousness that it does not properly terminate

fabianlim (changes requested):
try to remove the while loop in the regex

fabianlim:
Added a commit on top of this branch with two changes:

1. Simplified the regex: instead of a combined dim+dtype pattern with a module-level compiled RE, walk the leading `(\d+)x` tokens left-to-right and take the remainder as dtype. Avoids the known limitation with `>` in dtypes and drops the `_TENSOR_TYPE_RE` constant.
2. Dynamic dims (`?`) are silently dropped rather than mapped to `-1`, preserving prior behaviour — no callers handle `-1` in the shape tuple today.
3. Tests moved into `test_dialects_parse.py` alongside the existing `parser_utils` tests, rather than a new file.

fabianlim (approved):
@kcirred are you ok with the fixes I pushed? we can merge. ~I might have removed some of your original tests~ Nope: should be good

kcirred:
> @kcirred are you ok with the fixes I pushed? we can merge. ~I might have removed some of your original tests~ Nope: should be good

lgtm

Issue #76, opened 2026-06-02
[Bug] linalg.reduce latency: charges output element count instead of deferring to combiner region


fabianlim:
## Summary

`linalg.reduce` is registered with `latency_category=LC.COMPUTE_FLOAT` in `ktir_cpu/dialects/linalg_ops.py`. This causes `LatencyEstimator._estimate` to charge `_num_elements(result, operands)` — the **output** element count — on top of the combiner region ops that already fire through `record_op` individually.

For a `128×64 → 128` reduction this means:
- `linalg.reduce` charges 128 cycles (output shape)
- Each `arith.addf` in the combiner fires separately, charging 128×64 / simd_width cycles

The correct model mirrors how `linalg.matmul` works: the op itself is a pure compute orchestrator; the actual work is accounted for by the ops inside the combiner region. `linalg.reduce` should contribute **zero** cycles directly and let the combiner region ops (e.g. `arith.addf`) do all the accounting — they naturally accumulate the correct input element count.

## Fix

Change the registration in `linalg_ops.py`:

```python
# before
@register("linalg.reduce", latency_category=LC.COMPUTE_FLOAT)

# after
@register("linalg.reduce", latency_category=LC.ZERO)
```

## Acceptance criteria

- [ ] A `128×64 → 128` reduce with `arith.addf` combiner charges `128×64 / simd_elements_per_cycle` compute cycles total (from the `arith.addf` ops), not `128 + 128×64/simd`.
- [ ] Existing latency tests pass unchanged (or are updated to reflect the corrected cycle counts).

fabianlim:
**Update: `LC.ZERO` is the right direction, but the handler also needs to run the combiner region**

Heads-up from tracing PR #84: the acceptance criteria assume each `arith.addf` in the combiner *"fires separately"*, but it doesn't. The `linalg.reduce` handler only reads the combiner's **name** and does the reduction with one NumPy call — the combiner op is never dispatched, so it never reaches `record_op`. As-is, only `linalg.reduce` itself charges (softmax core 0: `linalg.reduce` 4.0 cyc, no `maximumf`/`addf` entries).

So `COMPUTE_FLOAT → ZERO` alone would make the reduction cost **vanish** (compute drops 16 → 12), not move it to the combiner. The fix needs `linalg.reduce` to actually execute its combiner region first; then `LC.ZERO` is correct.

[...]

PR #84, opened 2026-06-04
First issue: issue 76, latency for linalg.reduce

yuhaohaoyu:
Op notes on the first PR: 

1. Mostly claude directly to address issue #76 . 
2. Went through a mindful learning process with Bob-the-AI over the repo, test, with goal to build some fidelity of the vide'd test serves the purpose. 
3. Added tests with vector_reduce and vector_reduce for roofline, with the reduce mlir code under examples. 
4. Added a latency category: COMPUTE_REDUCE for 2 reasons:
   1.  REDUCE op can be non float strictly speaking (e.g. max). 
   2. Latency_of_reduce ~ O( size_of_inputs_to_reduce ),. slightly different from the existing ones for COMPUTE_FLOAT, which uses latency ~ O( size_inputs + size_outputs ). Yet the former estimation is specified in issue 76.

### DIFF
--- a/examples/ktir/reduce_multiop.mlir
+++ b/examples/ktir/reduce_multiop.mlir
@@ -1,0 +1,36 @@
+// linalg.reduce with a MULTI-OP combiner region: max expressed as
+// cmpf(ogt) + select rather than a single arith.maximumf.  This exercises the
+// general tree fold — every op in the region is executed and charged, and the
+// result does not depend on recognising a single combiner op by name.
+module {
+  func.func @reduce_multiop(%arg0: index) attributes {grid = [1, 1]} {
+    %c0 = arith.constant 0 : index
+    %cst = arith.constant 0xFF80 : f16  // -inf (identity for max)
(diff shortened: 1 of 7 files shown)


fabianlim (changes requested):
thanks for the first issue, I did a thorough review and found some things that is worth digging further into. I made it concrete with https://github.com/yuhaohaoyu/ktir-cpu/pull/1. Hope to discuss further

fabianlim (approved):
Approving. The reduce latency now comes from executing the combiner region (zero-cost `linalg.reduce`, like `linalg.generic`), which generalises to multi-op combiners and drops the NumPy name→reduction mapping.

Two pre-existing semantic gaps are left for follow-up (issue #85), each guarded by a `strict=True` xfail in `tests/test_dialects_exec.py::TestLinalg` so we're alerted if either starts passing:

- `test_reduce_multi_axis` (L891) — `dimensions = [0,1]` silently reduces only `dims[0]`.
- `test_reduce_folds_outs_init` (L908) — the `outs` init value isn't folded into the reduction.

Both are harmless under the current frontend (single-axis reduces, identity `outs` init).

yuhaohaoyu:
@fabianlim Fabian, your comment on 'Approving, ', I do not need to do anything now for this PR, the suggested in that comment is for next steps, right ?

fabianlim:
@yuhaohaoyu yes if the tests pass and if you are ok with the changes, we can merge this. I created a followup #85 that you also may be interested in

PR #81, opened 2026-06-04
[Fix] Handle linalg.batch_matmul in interpreter

fabianlim:
## Summary

- Adds a dedicated `linalg__batch_matmul` handler that executes batched matmul (`B×M×K @ B×K×N → B×M×N`) with optional accumulator.

## Note

`linalg.batch_matmul` is currently also registered on the `linalg__matmul` handler (line 128). The dedicated handler takes precedence at dispatch time. A follow-up can clean up the duplicate registration on `linalg__matmul`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #86, opened 2026-06-05
test: expose MLIRFrontendParser dropping dynamic SSA sizes

fabianlim:
## What

Adds tests that expose a bug in `MLIRFrontendParser`: for `ktdp.construct_memory_view` with dynamic memref dims (`memref<?x...>`), the parser leaves the `ShapedType` dynamic sentinel (`INT64_MIN`) in the `shape` attribute instead of the SSA-name string the executor needs. At execution time the handler recovers `coordinate_set` symbols from the string entries in `shape`, so the sentinel yields an empty symbol list and `specialize()` raises `IndexError`.

## Why this slipped through

- The dynamic execution test (`TestVectorAddDynamicExecution`) was never wired into the frontend adapter suite — it was added in #42, after the adapter class list was fixed in #10, and #42 only updated the parse-layer adapter.
- The parse-layer adapter assertion tolerated the sentinel by design.

## Changes (tests only)

- Wire `TestVectorAddDynamicExecution` into the frontend adapter suite (`TestVectorAddDynamicAdapt`).
- Tighten the mixed static/dynamic parse test to require the dynamic slot be an SSA-name string.

## Expected CI result

**These tests are intended to fail** on the first commit — the parser fix is deliberately not included here. The frontend path fails (`IndexError` in `specialize` / sentinel int in `shape`) while the regex path passes, isolating the defect to `MLIRFrontendParser`.

In the following commit, will push to fix and reinstate green status

### DIFF
--- a/ktir_cpu/mlir_frontend/parser.py
+++ b/ktir_cpu/mlir_frontend/parser.py
@@ -39,6 +39,7 @@ try:
         IntegerAttr,
         IntegerSetAttr,
         Module,
+        ShapedType,
     )
     from mlir_ktdp.passmanager import PassManager
     from tools_ktdp.ir_utils import ktdp_context, walk_module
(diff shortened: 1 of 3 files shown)


fabianlim:
@nwang-ibm @lchu6 @WarningRan I found a bug in the dynamic symbols handling when using mlir-frontend parser. I push this quickfix to restore functionality. Will appreciate a quick read as its blocking some other efforts now. If the main functionality is restored here, we can open follow up issues to handle other edge cases

nwang-ibm:
LGTM

fabianlim:
Follow-up: while auditing other handlers for the same SSA/dynamic-slot issue, found that the MLIR frontend path also can't parse distributed memory views (missing handler + per-core LX memory space). Tracked in #87 — out of scope for this PR.

cc @kiszk @lchu6 @nwang-ibm

nwang-ibm (approved):

fabianlim:
Ran a systematic audit for the same executor↔frontend test inconsistency that hid the dynamic-shape bug. Root cause: the regex parser has a catch-all fallback so new ops "just work", but the MLIR frontend has none and rejects any uninstalled op — and frontend coverage is purely derived from whichever regex test classes get subclassed into the adapter suites. So a new executor op + a direct handler-level test passes CI while silently breaking the frontend.

Found and fixed in this branch:
- `linalg.batch_matmul` — executor handler from #81, never installed in the frontend, no test on any path. Added handler + tests (regex parse, frontend parse, exec).
- `arith.index_castui`, `arith.ceildivui` — missing frontend install. Added handler + portable tests.
- `scf.if` — added a frontend handler + frontend-only test (note: the regex parser also silently drops `scf.if`).

Added a guard test (`tests/mlir_frontend/test_registry_consistency.py`) that fails if any `@register`'d executor op is neither frontend-installed nor explicitly allowlisted-with-reason — converting this whole class of bug from silent to loud.

Deeper gaps that can't be fixed in the Python adapter (ops unknown to the `mlir_ktdp` C++ dialect) filed separately:
- #88 — `ktdp.reduce`, `ktdp.coreid`
- #87 — distributed memory view (+ per-core LX memory space)

cc @kiszk @lchu6 @nwang-ibm

Issue #88, opened 2026-06-05
ktdp.reduce / ktdp.coreid are non-spec ops in the regex executor (reconcile or remove)


fabianlim:
## Summary

`ktir_cpu` registers executor handlers for `ktdp.reduce` and `ktdp.coreid` (`@register` in `ktir_cpu/dialects/ktdp_ops.py`), but **neither op exists in the authoritative `ktdp` dialect**. The dialect (`ktir-mlir-frontend`, `include/Ktdp/KtdpOps.td`) defines only:

`region_terminator`, `get_compute_tile_id`, `construct_memory_view`, `construct_distributed_memory_view`, `construct_access_tile`, `construct_indirect_access_tile`, `load`, `store`, `runtime_arg_extract`.

So `ktdp.reduce` / `ktdp.coreid` are **not** a frontend gap to be filled — they are non-spec op names that only survive on the regex parser path because that parser validates against no dialect (its `_parse_general_operation` fallback accepts any `op.name`). The MLIR frontend correctly rejects them at parse time (`custom op 'ktdp.reduce' is unknown`) because MLIR validates against the loaded dialect.

This is a **ktir-cpu conformance bug**, not a missing feature: per `CLAUDE.md` (do not invent `ktdp` ops not in the spec), these handlers should be reconciled to the real op forms or removed — **not** mirrored into the C++ dialect to match the invented spelling.

## What to do

- Determine the real, spec'd form of these operations (cross-core reduce; core identity). The actual op surface is being settled upstream in the dialect repo — see open PRs at https://github.com/torch-spyre/ktir-mlir-frontend/pulls.
- Once the real form lands in `KtdpOps.td`, update `ktir_cpu` to the spec'd op name/operands/attributes and add a frontend adapter handler + test.
- Until then, treat the current `ktdp.reduce` / `ktdp.coreid` handlers as non-conformant. `examples/ktir/ring_reduce.mlir` uses them and therefore cannot be parsed by the MLIR frontend.

## Guard

`tests/mlir_frontend/test_registry_consistency.py` lists these in `FRONTEND_UNSUPPORTED` pointing here, so the divergence stays visible until reconciled.

## Related

- Sibling of #87 (distributed memory view: a real spec op missing only a frontend adapter handler — different from this).
- Follow-up to #86.

Issue #87, opened 2026-06-05
[DistributedMemView] MLIRFrontendParser cannot parse distributed memory views (missing handler + per-core LX memory space)


fabianlim:
## Summary

`MLIRFrontendParser` (the native `mlir_ktdp` path) cannot parse programs that use distributed memory views. Two distinct gaps block `examples/rfc/distributed-view-copy.mlir`:

### Gap 1 — `construct_memory_view`: per-core LX memory space not parsed

The frontend handler's memory-space regex only matches the bare form:

```python
# ktir_cpu/mlir_frontend/parser.py (_adapt_construct_memory_view)
ms = re.search(r'#ktdp\.spyre_memory_space<(\w+)>', str(mlir_op.attributes["memory_space"]))
```

It rejects the RFC per-core form `#ktdp.spyre_memory_space<LX, core = 0>`:

```
ValueError: ktdp.construct_memory_view: cannot parse memory_space from
            Attribute(#ktdp.spyre_memory_space<LX, core = 0>)
```

The regex parser already handles this form (`ktir_cpu/dialects/ktdp_ops.py`, `<LX, core = N>` → `lx_core_id`). The frontend should match it and emit `lx_core_id`.

### Gap 2 — `construct_distributed_memory_view`: no frontend handler

There is no `@MLIRTypeAdapter.install("ktdp.construct_distributed_memory_view")` handler, so after bypassing Gap 1:

```
NotImplementedError: No MLIRTypeAdapter handler registered for op
                     'ktdp.construct_distributed_memory_view'.
```

The executor (`ktdp__construct_distributed_memory_view`) and regex parser both support it; only the frontend adapter is missing.

## Why this hasn't surfaced

No frontend (`tests/mlir_frontend/`) test loads `examples/rfc/distributed-view-copy.mlir`, so the only coverage for distributed views runs through the regex parser. Same structural gap as the dynamic-shape bug in #86 — the frontend adapter suite doesn't mirror the full set of distributed/dynamic execution tests.

## Suggested fix

- Extend the frontend `_adapt_construct_memory_view` memory-space parsing to accept `<LX, core = N>` and set `lx_core_id`, matching the regex parser.
- Add a frontend handler for `ktdp.construct_distributed_memory_view`.
- Add a frontend adapter test that loads `examples/rfc/distributed-view-copy.mlir` end-to-end.

Follow-up to #86.

yuhaohaoyu:
[...]

PR #89, opened 2026-06-05
fix: close executor↔frontend op coverage gaps + add guard test

fabianlim:
## What

Closes a class of bug where ops handled by the executor are silently unreachable through `MLIRFrontendParser`.

**Root cause:** the regex parser tolerates any `@register`'d executor op via its catch-all fallback (`_parse_general_operation`), but the MLIR frontend rejects any op lacking an explicit `@MLIRTypeAdapter.install` handler. Frontend test coverage is purely derived from which regex test classes are subclassed into the adapter suites — so an executor op plus a direct handler-level test passes CI while leaving the frontend broken (exactly what happened to `linalg.batch_matmul` in #81, and to dynamic shapes in #86).

## Changes

Install the missing frontend handlers and add tests on both parser paths:
- `linalg.batch_matmul` — regex parse, frontend parse, and exec tests (none existed)
- `arith.index_castui`, `arith.ceildivui` — portable parse tests
- `scf.if` — frontend handler + frontend-only test (the regex parser also silently drops `scf.if` — noted for follow-up)

**Guard test** — `tests/mlir_frontend/test_registry_consistency.py` asserts every executor op is either frontend-installed or explicitly allowlisted with a reason, so this whole class of divergence fails loudly going forward. The allowlist distinguishes:
- synthetic/alias ops (`region.bb0_args`, `return`)
- **non-spec ops** to reconcile or remove (`arith.convertf`, `ktdp.reduce`, `ktdp.coreid` — these are invented op names not in the authoritative `ktdp` dialect; see #88)
- **real spec ops** still missing a handler (`construct_distributed_memory_view` — see #87)

## Testing

`uv run pytest tests/mlir_frontend/ tests/test_dialects_parse.py tests/test_dialects_exec.py` → 372 passed, 7 skipped.

## Related

- Follows #86 (now merged)
- #87 — distributed memory view frontend handler + per-core LX memory space
- #88 — non-spec `ktdp.reduce` / `ktdp.coreid` to reconcile or remove

### DIFF
--- a/ktir_cpu/ops/control_ops.py
+++ b/ktir_cpu/ops/control_ops.py
@@ -70,7 +70,8 @@ class ControlOps:
         context.push_scope()
         result = region_executor(context, region)
         context.pop_scope()
-        return result
+        from ..dialects._helpers import unwrap_yield
+        return unwrap_yield(result)
 
     @staticmethod
     def for_op(
--- a/tests/test_dialects_exec.py
+++ b/tests/test_dialects_exec.py
@@ -1085,6 +1085,26 @@ class TestScfFunc:
         dispatch("scf.if")(op, ctx, env)
         assert ran == ["else"]
 
+    def test_if_then_else_yield_result(self):
+        # scf.if with a yielding then-branch returns the unwrapped value, not a _YieldResult wrapper
+        ctx = _ctx_with(**{"%cond": True, "%val": 42})
+        op = Operation(op_type="scf.if", operands=["%cond"], attributes={},
+                       result="%res", result_type=None,
+                       regions=[[Operation(op_type="scf.yield", operands=["%val"],
+                                           attributes={}, result=None, result_type=None)],
+                                 []])
+        env = _make_env()
+
+        def real_execute_region(ctx, ops):
+            result = None
+            for o in ops:
+                result = dispatch(o.op_type)(o, ctx, env)
+            return result
+
+        env.execute_region = real_execute_region
+        result = dispatch("scf.if")(op, ctx, env)
+        assert result == 42, f"expected 42, got {result!r}"
+
 # ---------------------------------------------------------------------------
 # ktdp dialect parsers
 # ---------------------------------------------------------------------------
--- a/tests/test_dialects_exec.py
+++ b/tests/test_dialects_exec.py
@@ -866,6 +866,17 @@ class TestLinalg:
         result = _call("linalg.matmul", ctx, _make_env(), operands=["%a", "%b"])
         assert np.allclose(result.data, b.data, rtol=1e-2)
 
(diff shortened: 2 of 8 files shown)


fabianlim:
cc @kiszk @lchu6 @nwang-ibm

lchu6:
for the `test_if_then_else` test, can we add a test for the returned result?   the returned result from yield might be a wrapper around the result, instead of the result itself.  it is good to check the returned result (a true value or a wrapped value that needs to be unpacked)

fabianlim:
@lchu6 addressed your comment abut the yield result, that was a good catch. See [0bc2d9a](https://github.com/torch-spyre/ktir-cpu/pull/89/commits/0bc2d9aa1fac398abecd3311f85fb6ad9389ae1f)

kiszk (approved):
generally LGTM. will leave some suggestions

tests/test_dialects_exec.py
```diff
@@ -945,6 +945,17 @@ def test_matmul(self):
...
+    def test_batch_matmul(self):
+        # batched identity: for each batch, I @ B == B
+        eye = np.broadcast_to(np.eye(2, dtype=np.float16), (3, 2, 2)).copy()
+        bdata = np.arange(3 * 2 * 2, dtype=np.float16).reshape(3, 2, 2)
+        a = Tile(eye, "f16", (3, 2, 2))
+        b = Tile(bdata, "f16", (3, 2, 2))
+        ctx = _ctx_with(**{"%a": a, "%b": b})
+        result = _call("linalg.batch_matmul", ctx, _make_env(), operands=["%a", "%b"])
+        assert result.shape == (3, 2, 2)
+        assert np.allclose(result.data, bdata, rtol=1e-2)
```
kiszk, line 957: It would be good if we could use a bit of practical data (non-identity).

ktir_cpu/ops/control_ops.py
```diff
@@ -70,7 +70,8 @@ def if_op(
         context.push_scope()
         result = region_executor(context, region)
         context.pop_scope()
-        return result
+        from ..dialects._helpers import unwrap_yield
+        return unwrap_yield(result)
```
lchu6, line 74: unwrap is good, but it seems on the interpreter side:
```
     # Multi-result ops return a tuple and have a list of result names.
            if isinstance(op.result, list) and isinstance(result, tuple):
                for name, val in zip(op.result, result):
                    context.set_value(name, val)
            else:
                context.set_value(op.result, result)
                if isinstance(result, Tile):
                    context.track_lx(op.result, result.size_bytes())
```
we failed to context.track_lx multi-value result?

lchu6 (commented):
only one place to double check

Issue #39, opened 2026-05-07
Refactor ops/ layer: shared tile/scalar helpers at package level


fabianlim:
## Background

PR #38 introduced `dialects/_helpers.py` with `_float_binop`, `_int_binop`, `_unary` to eliminate repetitive Tile-vs-scalar branching in the dialect layer (~60 lines saved there).

The same duplication exists in `ops/arith_ops.py` (592 lines) and `ops/math_ops.py` (268 lines). Two different patterns appear:

**Pattern A — `ops/arith_ops.py`: every int binop has the same 4-way branch (~68 instances):**
\`\`\`python
def andi(val1, val2):
    if isinstance(val1, Tile) and isinstance(val2, Tile):
        return Tile(val1.data & val2.data, val1.dtype, val1.shape)
    if isinstance(val1, Tile):
        return Tile(val1.data & int(val2), val1.dtype, val1.shape)
    if isinstance(val2, Tile):
        return Tile(int(val1) & val2.data, val2.dtype, val2.shape)
    return int(val1) & int(val2)

def shrsi(val1, val2):
    if isinstance(val1, Tile) and isinstance(val2, Tile):  # same structure again
        ...
\`\`\`

**Pattern B — `ops/math_ops.py`: every op is split into a `tile` method and a `scalar` method (16 scalar/tile pairs out of 33 methods):**
\`\`\`python
def exp(tile: Tile) -> Tile:
    result_data = np.exp(tile.data.astype(np.float32)).astype(tile.data.dtype)
    return Tile(result_data, tile.dtype, tile.shape)

def exp_scalar(value) -> np.floating:
    return np.float16(np.exp(float(value)))
\`\`\`

## Location

The helpers should live at `ktir_cpu/ops/_helpers.py` — same level as `ops/arith_ops.py` and `ops/math_ops.py`, parallel to `dialects/_helpers.py`. They must remain pure compute (numpy in, result out) with no IR knowledge.

## Suggested approach

**For Pattern A** — a `tile_binop(fn, val1, val2, scalar_cast=int)` helper:
\`\`\`python
def tile_binop(fn, val1, val2, scalar_cast=int):
    if isinstance(val1, Tile) and isinstance(val2, Tile):
        return Tile(fn(val1.data, val2.data), val1.dtype, val1.shape)
    if isinstance(val1, Tile):
        return Tile(fn(val1.data, scalar_cast(val2)), val1.dtype, val1.shape)
[...]

PR #93, opened 2026-06-05
[Refactor] ops/ layer: shared tile/scalar helpers at package level (issue #39)

lasch:
Introduce `ktir_cpu/ops/_helpers.py` with two shared helpers:
- `tile_binop(fn, val1, val2, scalar_cast=int)`: replaces the 4-way Tile/scalar branching repeated across ~14 integer binops in `arith_ops.py`
- `tile_unary(fn, val)`: replaces the tile/scalar method pairs in `math_ops.py`

`ops/arith_ops.py`: integer binops (`addi`, `subi`, `muli`, `divsi`, `remsi`, `divui`, `remui`, `floordivsi`, `ceildivsi`, `ceildivui`, `andi`, `ori`, `xori`, `shli`, `shrsi`, `minsi`, `maxsi`, `minui`, `maxui`) reduced to one-liners via `tile_binop`.

`ops/math_ops.py`: 16 tile/scalar method pairs collapsed into unified methods via `tile_unary`; `powf`, `fma`, `absf`, `absi` kept as-is.

`dialects/math_ops.py`: `_scalar` second args removed from `_unary` calls. `dialects/_helpers.py`: `_unary` `tile_fn` param renamed to `fn`; docstring updated.

Also cleans up a pre-existing issue from main: `dialects/arith_ops.py` had a duplicate `arith.cmpf` registration (introduced in PR #38): a thin delegating handler was registered first, then immediately overridden by the full implementation registered second. The delegating handler was therefore never reachable. The dead handler is removed, and `ArithOps.cmpf` result dtype is fixed to return `"i1"` instead of propagating the float input dtype (consistent with `cmpi`). Regression test added (`TestArithOpsCmpf.test_result_dtype_is_i1`).

One particular item to review/comment on is the change in `ktir_cpu/dialects/_helpers.py`. Since the 2nd input arg becomes generic (tile or scalar fn), it's been renamed to just `fn`. Accordingly, the explicit provision of the scalar functions has been removed from `dialects/math_ops.py`.

Net: ~390 lines removed, no behaviour change (814 passed, 7 xfailed).


ktir_cpu/ops/_helpers.py
```diff
@@ -0,0 +1,51 @@
...
+
+Pure helpers (numpy in, result out) with no IR knowledge.
+"""
+
+import numpy as np
+
+from ..ir_types import Tile
+
+
+def tile_binop(fn, val1, val2, scalar_cast=int):
```
fabianlim: there is a mix between tile_binop which does implicit int cast and `tile_unary` which does implicit float cast, so I suggest to use a generic helper then specialize,


```python
  from functools import partial

  def _tile_binop(fn, val1, val2, *, scalar_cast):   # keyword-only: no accidental 4th positional
      if isinstance(val1, Tile) and isinstance(val2, Tile):
          r = fn(val1.data, val2.data);  return Tile(r, val1.dtype, r.shape)
      if isinstance(val1, Tile):
          r = fn(val1.data, scalar_cast(val2));  return Tile(r, val1.dtype, r.shape)
      if isinstance(val2, Tile):
          r = fn(scalar_cast(val1), val2.data);  return Tile(r, val2.dtype, r.shape)
      return fn(scalar_cast(val1), scalar_cast(val2))

  tile_binop_int   = partial(_tile_binop, scalar_cast=int)
  tile_binop_float = partial(_tile_binop, scalar_cast=float)
```
  Integer binops call tile_binop_int; the cast intent is now in the name. The same pattern applies to the unary side
  (tile_unary_int / tile_unary_float) — absi could then use tile_unary_int instead of its hand-rolled path, removing the
  special-case. tile_binop_float also makes the obvious follow-up (folding addf/subf/mulf/divf in) a one-liner, though
  that's optional scope.

ktir_cpu/dialects/_helpers.py
```diff
@@ -70,13 +70,13 @@ def _int_binop(op, context, fn):
...
-    *scalar_fn* is called for scalars; defaults to *tile_fn* if omitted.
+    *fn* is called for both Tiles and scalars unless *scalar_fn* is given,
+    in which case *scalar_fn* handles the scalar path.
     """
     val = context.get_value(op.operands[0])
     if isinstance(val, Tile):
-        return tile_fn(val)
-    return (scalar_fn or tile_fn)(val)
+        return fn(val)
+    return (scalar_fn or fn)(val)
```
fabianlim, line 88: PR text re _unary is slightly off. dialects/_helpers.py::_unary's scalar_fn is not retired — dialects/arith_ops.py cast ops still pass it (extf→np.float32; extui/trunci/fptosi/fptoui→int; uitofp→float). 

Will be good to document in _unary doc string those few examples, and why they need a different scalar_fn

ktir_cpu/ops/_helpers.py
fabianlim: Also Latent shape bug in `tile_binop.` It hardcodes `Tile(..., val1.shape)`; old code used `result_data.shape`, which I think is more correct.

ktir_cpu/dialects/_helpers.py
fabianlim: is there any reason why we do not do casting in `_unary` but want to do it in the tile helpers?

fabianlim (changes requested):
Couple of questions regarding the implicit casting, one point on shape.

ktir_cpu/ops/arith_ops.py
```diff
@@ -514,80 +428,40 @@ def fptoui(value):
...
-        if isinstance(val1, Tile) and isinstance(val2, Tile):
-            return Tile(np.ceil(val1.data / val2.data).astype(val1.data.dtype), val1.dtype, val1.shape)
-        if isinstance(val1, Tile):
-            return Tile(np.ceil(val1.data / int(val2)).astype(val1.data.dtype), val1.dtype, val1.shape)
-        if isinstance(val2, Tile):
-            return Tile(np.ceil(int(val1) / val2.data).astype(val2.data.dtype), val2.dtype, val2.shape)
-        return math.ceil(val1 / val2)
+        def _ceildiv(a, b):
+            result = np.ceil(a / b)
+            return result.astype(a.dtype) if isinstance(a, np.ndarray) else math.ceil(result)
```
lchu6: if we have a as scalar value, b as ndarray.

then we are returning math.ceil(result) while `result` is a ndarray?

ktir_cpu/ops/arith_ops.py
```diff
@@ -603,58 +477,30 @@ def shrui(val1, val2):
...
-        if isinstance(val1, Tile) and isinstance(val2, Tile):
-            return Tile(np.ceil(val1.data / val2.data).astype(val1.data.dtype), val1.dtype, val1.shape)
-        if isinstance(val1, Tile):
-            return Tile(np.ceil(val1.data / int(val2)).astype(val1.data.dtype), val1.dtype, val1.shape)
-        if isinstance(val2, Tile):
-            return Tile(np.ceil(int(val1) / val2.data).astype(val2.data.dtype), val2.dtype, val2.shape)
-        return math.ceil(int(val1) / int(val2))
+        def _ceildiv(a, b):
+            result = np.ceil(a / b)
+            return result.astype(a.dtype) if isinstance(a, np.ndarray) else math.ceil(result)
```
lchu6: same here

ktir_cpu/ops/math_ops.py
```diff
@@ -21,224 +21,92 @@
...
+            return Tile(np.abs(val.data), val.dtype, val.shape)
+        return type(val)(abs(float(val)))
 
     @staticmethod
-    def floor_scalar(value: np.floating) -> np.floating:
-        """Scalar floor."""
-        return type(value)(np.floor(float(value)))
+    def absi(val):
+        """Element-wise absolute value (integer)."""
+        return tile_unary_int(np.abs, val)
```
lchu6, line 97: in your `_tile_unary`, you casts tile data to float32 first, would that hurt big integer values?  any value bigger than the f32 range but within i32 range will produce wrong results I think.

lasch:
@lchu6 Thanks for your comments. I'm going to turn this into a draft PR - I pushed my update without noting that I myself had to review the changes before asking for other feedback. Update coming asap...

ktir_cpu/ops/math_ops.py
lasch: forced cast has been removed from `tile_unary_int`.

ktir_cpu/dialects/_helpers.py
fabianlim: The docstring still doesn't explain *why* some ops need `scalar_fn`. Suggest adding a note:

```python
    """Fetch one operand and apply a unary function.

    *fn* is called for both Tiles and scalars unless *scalar_fn* is given,
    in which case *scalar_fn* handles the scalar path.

    *scalar_fn* is needed for cast ops that must coerce the scalar to a
    specific Python type before passing to fn — e.g. extf passes
    scalar_fn=np.float32, extui/trunci/fptosi/fptoui pass int, uitofp passes
    float.
    """
```

ktir_cpu/ops/_helpers.py
fabianlim: `tile_binop_float = partial(_tile_binop, scalar_cast=lambda x: x)` is inconsistent with `tile_binop_int` — the identity lambda means scalars pass through uncoerced. Should be `scalar_cast=float` for symmetry and to ensure scalar inputs are explicitly coerced, matching the pattern the original comment suggested.

ktir_cpu/dialects/_helpers.py
lasch: updated

ktir_cpu/ops/_helpers.py
lasch: should be fixed now.

lasch: sorry, forgot to push. It should be fixed with commit ddf1f38

fabianlim (approved):
comments are addressed

PR #103, opened 2026-06-10
[Feature] Support tensor.reshape and dense list constants

kcirred:
Lowers tt.reshape to tensor.reshape, fed by either tensor.from_elements 
(when the target shape is built from SSA values) or a folded arith.constant 
of the form dense<[16, 32]> : tensor<2xindex>. The interpreter previously
supported none of these, so any reshape-bearing kernel failed at parse
or execution time.

- Add tensor.reshape executor + parsers (regex and MLIR-frontend).
  Target shape is read from the result-type annotation rather than the
  runtime shape operand, matching how tensor.expand_shape and
  tensor.collapse_shape already work — the result type always pins the
  target shape statically, even when the operand is dynamic.
- Add tensor.from_elements executor + parsers, since it is the typical
  shape-operand producer for tensor.reshape.
- Extend arith.constant to handle non-splat dense<[...]> payloads. The
  splat path was the only one implemented, which silently mis-parsed
  list payloads to a splat-of-zero tensor. A new dense_list attribute
  routes the executor to materialize the array element-by-element.
  Covers both the regex parser and the MLIR-frontend adapter.

Tests pin the regression cases and exercise rank-changing reshapes
(1D->2D, 2D->1D, 1D->3D), the smallest legal tensor.from_elements
(N=1), an N>2 case, dtype generality of the dense-list path (index
and f32), and the size-mismatch error path.

### DIFF
--- a/ktir_cpu/dialects/arith_ops.py
+++ b/ktir_cpu/dialects/arith_ops.py
@@ -331,6 +331,10 @@ def arith__constant(op, context, env):
         shape = op.attributes["shape"]
         dtype_str = op.attributes.get("dtype", "f16")
         np_dtype = to_np_dtype(dtype_str)
+        # dense<[v0, v1, ...]> list form: each element is distinct, so
+        # build the array element-by-element rather than splatting one value.
+        if op.attributes.get("dense_list"):
+            return Tile(np.array(value, dtype=np_dtype).reshape(shape), dtype_str, shape)
         return Tile(np.full(shape, value, dtype=np_dtype), dtype_str, shape)
     return value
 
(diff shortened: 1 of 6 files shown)


fabianlim:
@kcirred pushed a follow-up refactor commit (a211593) to your branch:
- `parse_dense_payload` moved from `arith_ops.py` to `parser_utils.py` as a public function
- `parse_arith_constant` dispatch unified via `_set_dense`/`_set_tensor_attrs` helpers
- `_adapt_arith_constant` flattened to if/elif/else, removing the `is_list` flag pattern

All tests pass.

fabianlim (dismissed):
we addressed some refactoring and looks good now

fabianlim (approved):
we did some refactoring, looks good now

PR #104, opened 2026-06-10
fix: track_lx for multi-result ops; unify result-binding; strengthen batch_matmul test

fabianlim:
## What

Addresses two review comments from #89.

## Changes

- **`interpreter.py`** — multi-result branch (`op.result` is a list) never called `context.track_lx` for `Tile` values, while the single-result path did. Unified both paths into one loop over zipped names/values so `track_lx` is called for any `Tile` regardless of result count.
- **`test_batch_matmul`** — replaced identity-matrix `A` with non-identity data; uses `np.einsum` as the reference so index transposition bugs would produce wrong values.

## Related

- Follows #89

### DIFF
--- a/ktir_cpu/interpreter.py
+++ b/ktir_cpu/interpreter.py
@@ -214,14 +214,12 @@ class KTIRInterpreter:
         # Other result types — TileRef (metadata), AccessTile (coordinates),
         # int/index (scalars) — are bookkeeping and use no scratchpad memory.
         if op.result and result is not None:
-            # Multi-result ops return a tuple and have a list of result names.
-            if isinstance(op.result, list) and isinstance(result, tuple):
-                for name, val in zip(op.result, result):
-                    context.set_value(name, val)
-            else:
-                context.set_value(op.result, result)
-                if isinstance(result, Tile):
-                    context.track_lx(op.result, result.size_bytes())
+            names = op.result if isinstance(op.result, list) else [op.result]
+            values = result if isinstance(result, tuple) else [result]
+            for name, val in zip(names, values):
+                context.set_value(name, val)
+                if isinstance(val, Tile):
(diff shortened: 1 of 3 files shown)


fabianlim:
cc @kiszk @lchu6 — this addresses your review comments on #89.

kiszk:
Thanks, lgtm regarding my comment.

ktir_cpu/interpreter.py
```diff
@@ -214,14 +214,12 @@ def _execute_op(self, op: Operation, context: CoreContext) -> Any:
...
-            if isinstance(op.result, list) and isinstance(result, tuple):
-                for name, val in zip(op.result, result):
-                    context.set_value(name, val)
-            else:
-                context.set_value(op.result, result)
-                if isinstance(result, Tile):
-                    context.track_lx(op.result, result.size_bytes())
+            names = op.result if isinstance(op.result, list) else [op.result]
+            values = result if isinstance(result, tuple) else [result]
+            for name, val in zip(names, values):
```
lchu6, line 223: do we want a 
```
en(names) == len(values)
```
check?

this impl silently accept multi-result arity mismatches, say result=["%a", "%b"] with handler result 30.

not saying this would happen in real world, but a buggy/new multi-result handler could return a scalar or a too-short tuple which silently not captured.  

so maybe one more line to add a guard here.

lchu6 (commented):
left one minor

lchu6:
looks like our assertion did catch a mismatch...

fabianlim:
@lchu6 it was a test mistake, not a code mistake

fabianlim:
@lchu6 i removed the assertion and changed it to an actual exception raise as requested.

lchu6 (approved):

Issue #85, opened 2026-06-05
linalg.reduce: multi-axis reduce and outs-init not handled per MLIR semantics


fabianlim:
## Summary

cc: @yuhaohaoyu 

Two pre-existing `linalg.reduce` semantic gaps, surfaced while reworking the reduce handler in #84. Both are currently harmless under the frontend's actual output, but diverge from MLIR `linalg.reduce` semantics. Each has a `strict=True` xfail test (`tests/test_dialects_exec.py::TestLinalg`) so we're alerted if/when they get fixed.

### 1. Multi-axis reduce silently reduces only one axis

`linalg.reduce ... dimensions = [0, 1]` should reduce **all** listed axes. The parser (`parse_linalg_reduce`) parses the full list but keeps only `dims[0]`, and the tree fold indexes a single axis — so a multi-axis reduce silently reduces just the first axis (and crashes if a list reaches `_tree_fold` directly).

- xfail: `test_reduce_multi_axis`
- All current examples are single-axis, so nothing hits this today.

### 2. `outs` init value is not folded into the reduction

MLIR semantics: the `outs` buffer is the initial accumulator. The tree fold seeds from the input only, so e.g. `sum([1,2,3,4])` with `outs` init = 100 returns 10, not 110.

- xfail: `test_reduce_folds_outs_init`
- Harmless while the frontend inits `outs` to the identity (`0` for sum, `-inf` for max), as all current examples do.

## Fix sketch

1. Parser: keep the full `dimensions` list; handler: fold over each axis (or reshape the reduced axes together) and squeeze all of them.
2. Seed the tree fold's accumulator with the `outs` tile, or fold it in as a final combine step.

yuhaohaoyu:
1. PR #106 addresses the multi-axis tensor reduce
2. PR #105 addresses the requirement that the LHS of a tensor.reduce needs to be updated upon, not simply an assignment into.

This issue should be ready for closing or extension.

PR #105, opened 2026-06-11
Reduce: do update to the output parameter instead of assignment, with…

yuhaohaoyu:
Issue #85, part 2

## Fix Description

**Targeted fix:**

1. support Update ops to the output parameter of a linalg.reduce op, instead of assignment. 
2. toggled the test checker in `test_reduce_folds_outs_init` to expect pass (from xfail)

**Voluntary fixes:**

3. test_latency's reference (expected) value for reduce ops got adjusted with "+1" (by Claude)
4. In a handful mlir example programs (under examples/ folder), the `neg_inf` (negative infinity, -inf) in fp16 data type got initialized as `0xFF80` (NaN, aka, _Not a Number_, in fp16 data format), it'd be `0xFC00` (-inf in fp16). (by Claude, for 1 hr)

## Code changes

The numbering of the code changes are associated with the numberings of the 4 items in the "fix description" section

1. Fix `linalg__reduce` op to add a combination op (mlir term: combiner) to perform: Output_Parameter `+=` Reducd_result_in_the_reduce_op. A simple fix was to "add one extra combiner call with inputs `reduced_tile` and `outs_tile`, and output `combined` and return it. 
2. Removed the xfail test-checker for `test_reduce_folds_outs_init`. This was specified in the issue.
3. Constant expected value adjustments, 13 lines in test_latency.py.
4. Constant adjustment, from 0xFF80 to 0xFC00, aka, NaN to _inf for fp16 data format.

## Conclusion
1. issue #85 is half way done, the easy half. 
2. part 1 of issue #85 is for multiple dimensional reduction ( number_reduce_dimension >= 2 ), working on it now.

PR #106, opened 2026-06-11
Multi dimension tensor reduction op (linalg.reduce) support 

yuhaohaoyu:
Issue #85 request 1

## Operational design requirements with Claude: 

1. augment linalg.reduce support with tree_fold for multiple-dimension reduction
2. the number of reduce dimensions will be from 0 to number_of_dimensions of the inputy tensor
3. the reduce dimension can be adjecent or not adjecent.
4. the reduce operator does not have assumed commutativity, only with associativity.

## Implementation planning (penned by Claude): 

1. Parse and preserve the full dimensions list (no truncation to dims[0]).
2. Sort dimensions descending — rightmost (fastest-moving) axis folds first, matching linalg.reduce's innermost-first iteration order.
3. Loop _tree_fold once per axis; 
4. After all folds complete, squeeze all reduced axes together and combine with outs.

## Tests involved: 

1. `test_reduce_multi_axis`  --> expected to pass
2. `test_reduce_multi_axis_3d_disjoint_2d`. --> expected to pass
3. `test_reduce_multi_axis_treefold_bug`. --> expected to fail
4. TestLinalgParsers::test_reduce --> to intake 'dims' instead of 'dim'

## Next openings: tree_fold is not MLIR complied

MLIR's upstream documentation **does not** explicitly require **commutativity** for `linalg.reduce combiners` — only **associativity**. This is corroborated by IREE's documentation which states: "The combiner must be associative. It does not need to be commutative." In practice, all standard combiners used today (arith.addf, arith.maximumf, arith.mulf) happen to be commutative, but the spec does not contract on it — meaning **an implementation that reorders elements (as our tree fold does across disjoint axes) technically violates the semantics for a non-commutative associative combiner.** (penned by Claude).

### DIFF
--- a/ktir_cpu/dialects/linalg_ops.py
+++ b/ktir_cpu/dialects/linalg_ops.py
@@ -194,7 +194,6 @@ def linalg__reduce(op, context, env):
         ]
 
     dims = op.attributes.get("dims")
-    # Back-compat: legacy single-axis attribute
     if dims is None and op.attributes.get("dim") is not None:
         dims = [op.attributes["dim"]]
 
--- a/ktir_cpu/dialects/linalg_ops.py
+++ b/ktir_cpu/dialects/linalg_ops.py
@@ -207,6 +207,8 @@ def linalg__reduce(op, context, env):
         else:
             # Fold each axis, fastest-moving (rightmost) first.
             folded = tile.data
+	    # Tree-folding each axis independently reorders element groupings vs.
+	    # MLIR's left-associative scalar loop — see test_reduce_multi_axis_treefold_bug.
             for d in sorted(dims, reverse=True):
                 folded = _tree_fold(
                     Tile(folded, tile.dtype, folded.shape),
--- a/ktir_cpu/dialects/linalg_ops.py
+++ b/ktir_cpu/dialects/linalg_ops.py
@@ -193,17 +193,35 @@ def linalg__reduce(op, context, env):
             ),
         ]
 
-    dim = op.attributes.get("dim")  # axis to reduce; None → collapse all
+    dims = op.attributes.get("dims")
+    # Back-compat: legacy single-axis attribute
+    if dims is None and op.attributes.get("dim") is not None:
+        dims = [op.attributes["dim"]]
 
     outs_var = op.attributes.get("outs_var")
     outs_tile = context.get_value(outs_var) if outs_var else None
 
     if isinstance(tile, Tile):
-        folded = _tree_fold(tile, dim, bb0_names, body_ops, context, env)
-        if dim is None:
+        if dims is None:
+            # Collapse all axes: flatten then fold.
+            folded = _tree_fold(tile, None, bb0_names, body_ops, context, env)
+        else:
+            # Fold each axis, fastest-moving (rightmost) first.
+            folded = tile.data
+            for d in sorted(dims, reverse=True):
+                folded = _tree_fold(
+                    Tile(folded, tile.dtype, folded.shape),
+                    d, bb0_names, body_ops, context, env,
+                )
+
+        # Squeeze all reduced axes.
+        if dims is None:
             reduced = folded.reshape(()).astype(tile.data.dtype)
         else:
-            reduced = np.squeeze(folded, axis=dim).astype(tile.data.dtype)
+            reduced = folded
+            for d in sorted(dims, reverse=True):
+                reduced = np.squeeze(reduced, axis=d)
+            reduced = reduced.astype(tile.data.dtype)
 
         # Combine with the outs initial accumulator.
         if isinstance(outs_tile, Tile):
@@ -451,12 +469,11 @@ def parse_linalg_reduce(op_text, parse_ctx):
     if combiner_match:
         reduce_fn = combiner_match.group(1)
 
-    # dimensions = [1]
-    dim = None
+    # dimensions = [1] or dimensions = [0, 1]
+    dims = None
     dims_match = re.search(r'dimensions\s*=\s*\[(\d+(?:\s*,\s*\d+)*)\]', op_text)
     if dims_match:
         dims = [int(d.strip()) for d in dims_match.group(1).split(',')]
-        dim = dims[0]
 
     # ins(%x : type) — first operand is the input
     operands = []
@@ -471,8 +488,8 @@ def parse_linalg_reduce(op_text, parse_ctx):
         outs_var = outs_match.group(1)
 
     attributes = {"reduce_fn": reduce_fn}
-    if dim is not None:
-        attributes["dim"] = dim
+    if dims is not None:
+        attributes["dims"] = dims
     if outs_var is not None:
         attributes["outs_var"] = outs_var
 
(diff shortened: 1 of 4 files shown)


ktir_cpu/dialects/linalg_ops.py
```diff
@@ -193,18 +193,45 @@ def linalg__reduce(op, context, env):
             ),
         ]
 
-    dim = op.attributes.get("dim")  # axis to reduce; None → collapse all
+    dims = op.attributes.get("dims")
+    # Back-compat: legacy single-axis attribute
```
fabianlim: The back-compat shim can be dropped: both `parse_linalg_reduce` and `_adapt_linalg_reduce` are updated in this same PR to write `"dims"`, so nothing produces `"dim"` for `linalg.reduce` anymore. No caller path reaches this branch.

ktir_cpu/dialects/linalg_ops.py
```diff
@@ -193,18 +193,45 @@ def linalg__reduce(op, context, env):
...
+        if dims is None:
+            reduced = folded.reshape(()).astype(tile.data.dtype)
+        else:
+            reduced = folded
+            for d in sorted(dims, reverse=True):
+                reduced = np.squeeze(reduced, axis=d)
+            reduced = reduced.astype(tile.data.dtype)
+
+        # Combine with the outs initial accumulator.
+        if isinstance(outs_tile, Tile):
```
lchu6, line 228: outs_tile seems to be referenced before it is ever assigned?

based on your intention here, I assume you might want to define outs_tile by resolving outs_var from the context, and handle the case where there is no outs_var.

ktir_cpu/dialects/linalg_ops.py
```diff
@@ -426,12 +453,11 @@ def parse_linalg_reduce(op_text, parse_ctx):
     if combiner_match:
         reduce_fn = combiner_match.group(1)
 
-    # dimensions = [1]
-    dim = None
+    # dimensions = [1] or dimensions = [0, 1]
+    dims = None
     dims_match = re.search(r'dimensions\s*=\s*\[(\d+(?:\s*,\s*\d+)*)\]', op_text)
```
lchu6: the regex parser cannot parse dimensions = []. That leaves dims unset, and the executor treats dims is None as “collapse all axes”. Since the PR’s stated goal includes zero reduce dimensions, regex-parsed dimensions = [] will silently become a full reduction instead of a no-op/shape-preserving reduction.

yuhaohaoyu: Neat. Good point

ktir_cpu/dialects/linalg_ops.py
yuhaohaoyu: Will fix this

ktir_cpu/dialects/linalg_ops.py
yuhaohaoyu: comment removed. Thanks

ktir_cpu/dialects/linalg_ops.py
yuhaohaoyu: outs_tile was defined in #105 , this is addressed.

ktir_cpu/dialects/linalg_ops.py
yuhaohaoyu: Addressed with commit https://github.com/torch-spyre/ktir-cpu/pull/106/commits/55f07ed8d07f595e916a28be07553ccaa5e53a29

yuhaohaoyu:
@lchu6 @fabianlim Should addressed the issue. the dims=[] comment was good. 

Give me a reminder if need attention, I am not sure about the pgp signing fix over the older commit.

tests/test_dialects_exec.py
```diff
@@ -882,8 +882,8 @@ def test_reduce_explicit_region_single_op(self):
 
     def test_reduce_multiop_combiner(self):
         # MULTI-OP combiner: max expressed as cmpf(ogt) + select. The general
-        # tree fold must run BOTH region ops — there is no single combiner name
-        # to map to a NumPy reduction — and still return the correct max.
+        # tree fold must run BOTH region ops - here is no single combiner name
```
fabianlim: you introduced typos here

fabianlim (commented):
Two comment suggestions to make the tree-fold ordering divergence concrete:

**In `test_reduce_multi_axis_treefold_bug`**, add before the test body:

```python
# Tree-fold vs. MLIR sequential order for dims=[0, 1] on shape (2, 3):
#
#   data = [[a, b, c],
#            [d, e, f]]
#
# MLIR (scalar loop, row-major left-associative):
#   ((((( a + b ) + c ) + d ) + e ) + f )
#
# Tree-fold (this impl): fold axis 1 first, then axis 0:
#   axis-1: (a+b)+c,  (d+e)+f
#   axis-0: ((a+b)+c) + ((d+e)+f)
#
# These groupings differ, so f16 rounding can diverge.
```

**At the multi-axis fold loop in `linalg__reduce`**, add above the `for` loop:

```python
# Tree-folding each axis independently reorders element groupings vs.
# MLIR's left-associative scalar loop — see test_reduce_multi_axis_treefold_bug.
for d in sorted(dims, reverse=True):
    folded = _tree_fold(...)
```

yuhaohaoyu:
@fabianlim the two comments addressed. thanks

fabianlim (approved):
my comments are addressed

yuhaohaoyu:
@fabianlim I had to a pgp signing of the last week commit, with a cherry-picking, now it seems addressed the 'out of date branch' and 'a commit without pgp signing'. 

May need to re-kick the auto-merging ? Thanks.

fabianlim:
@yuhaohaoyu ready to merge but you have a DCO issue

yuhaohaoyu:
had been fighting dco or pgp. @fabianlim will address it now.

yuhaohaoyu:
@fabianlim did not want to pollute the space, the DCO complain is pinned down, it's not reasonable. my bot suggests a github issue, and the finger points to : commit `f5c1496`. 


Reference to keep here, need to study this. :) 

---
**_My Ques_**
please check this: https://github.com/torch-spyre/ktir-cpu/pull/106/changes/fc6a806ce6fb218b62fb234d7df8755644f7a81d

the message has my signature, yet the DCO red cross is still there.

**_Response_**
The problem is commit `f5c1496` — not your PR,  the DCO bot can't match:

- Commit email: `fabianlim@users.noreply.github.com` (GitHub's noreply)
- Signed-off-by lines in the PR: `flim@sg.ibm.com`

The DCO bot is checking all commits in the PR's diff range, including commits from upstream/main that got pulled in when the branch was behind. Once you rebase onto the latest upstream/main, that commit becomes part of the base — not part of the PR — and the DCO bot will stop checking it.

**So "branch out of date" and "DCO failure" are the same fix:**

```
git rebase -S upstream/main
git push --force-with-lease origin multi-dim-reduce-issue-85-p1
```

Once your branch is based on current upstream/main, `f5c1496` is in the base branch and the DCO bot only checks your 4 signed commits. Both blockers resolved in one shot.

Issue #109, opened 2026-06-12
[Bug] LX scratchpad double-counts `scf.for` iter_arg init values


nwang-ibm:
## Problem

The interpreter's LX tracking counts the same physical buffer twice when `scf.for` binds an iter_arg from an existing SSA value.

Given this MLIR pattern:
```mlir
%accum_zero = arith.constant dense<0.0> : tensor<768x1024xf16>
%result = scf.for %k = %c0 to %K step %BK iter_args(%accum_itr = %accum_zero) {
  ...
}
```

**What happens:**
1. `arith.constant` creates a Tile → `track_lx("%accum_zero", 1572864)` → LX used = 1.5 MB
2. `scf.for` binds iter_arg → `track_lx("%accum_itr", 1572864)` → LX used = 3.0 MB → **MemoryError**

Both SSA names (`%accum_zero` and `%accum_itr`) reference the same Python Tile object (`id()` is identical), but are tracked as independent LX allocations.

**Expected:** LX used = 1.5 MB. The iter_arg initialization is a name binding (pointer assignment), not a data copy. Real hardware does a register rename or buffer handoff — the data stays at the same LX address.

## Impact

Any matmul kernel with BLOCK_M × BLOCK_N > 524,288 (e.g., 768×1024) triggers a spurious MemoryError even though the actual LX footprint is well within 2 MB. This prevents running large tile sizes needed for compute-bound matmul (AI > 384 requires BLOCK_M, BLOCK_N ≥ ~768).

**Reproducer:**
```python
from ktir_cpu.interpreter import KTIRInterpreter
from ktir_cpu.latency import HardwareConfig
import numpy as np

hw = HardwareConfig(num_cores=32, clock_ghz=1.2, hbm_bandwidth_tb_s=0.2048,
                    systolic_flops_per_cycle=2048)
interp = KTIRInterpreter(latency_config=hw)

# Any matmul kernel with BLOCK_M=768, BLOCK_N=1024, BLOCK_K=32
# Hardware LX = (768*32 + 32*1024 + 768*1024) * 2 = 1.6 MB < 2 MB
# But interpreter tracks 2 × 768×1024×2 = 3 MB for the accum alone
interp.load(mlir_text)  # matmul with above block sizes
interp.execute_function(...)  # → MemoryError: LX scratchpad overflow
```

## Root Cause

`ktir_cpu/ops/control_ops.py` line 118:
```python
for name, val in zip(iter_arg_names, current_values):
[...]

Issue #102, opened 2026-06-10
[Feature] HBM/LX memory bootstrap: explicit placement API and allocation-aware next_ptr


fabianlim:
Follows up on #101 and the comment in fabianlim/ktir-cpu#5 (comment 4579332740).

## Problem

Three related gaps in the memory simulator that make tests brittle and allow silent aliasing:

### 1. No way to place a tensor at a specific address in `execute_function`

`execute_function` only accepts `ndarray` kwargs — it always allocates wherever `next_ptr` lands. Tests that embed literal MLIR pointer constants (e.g. `arith.constant 32 : index`) must monkey-patch `_prepare_execution` to write data after allocation, manually keeping the constant and the write address in sync. There is no supported API for "place this tensor at element index N."

### 2. `next_ptr` is unaware of externally bootstrapped memory

`HBMSimulator.allocate` bumps `next_ptr` from `0x10000` with no knowledge of anything written directly via `hbm.write(stick, ...)`. If a test writes to a stick that falls inside the allocator's range and then `execute_function` allocates, it may return the same stick again — silent aliasing, no assertion.

### 3. Same gap in `LXScratchpad`

`LXScratchpad.next_ptr` starts at 0 with no knowledge of externally written regions. `pop_scope` rewinds `next_ptr` blindly, which could overwrite bootstrapped data on the next execution within the same test.

## Proposed fix

**`execute_function` tuple API** — accept `(array, elem_idx)` tuples in kwargs:

```python
outputs = interp.execute_function("kernel",
    x_ptr=(x_array, 0),      # place x at element index 0
    idx_ptr=(idx_array, 32),  # place idx at element index 32 (i32, stick 1)
    y_ptr=y_array,            # allocate wherever (current behaviour)
)
```

When `elem_idx` is given: assert stick-aligned (`elem_idx * bpe % STICK_BYTES == 0`), write to that address, pass `elem_idx` as the MLIR SSA arg. When `None` / bare ndarray: current behaviour unchanged.

**`HBMSimulator.bootstrap(elem_idx, array)`** — writes data *and* registers the region with the allocator so `next_ptr` skips over it if it would otherwise overlap. `allocate` must skip over any already-occupied stick range rather than blindly bumping.

**`LXScratchpad` bootstrap contract** — same `bootstrap` API; `pop_scope` rewinds only to the high-water mark of bootstrapped allocations, never below, so externally seeded data survives scope rewind.

## Acceptance criteria

[...]

PR #110, opened 2026-06-12
[Fix] MemRef.base_ptr is now an element index; fix byte_address formula

fabianlim:
cc: @Prasanth-Chatarasi 

## Summary

- `MemRef.base_ptr` was a stick index for HBM (`byte_address = base_ptr * 128`). MLIR pointer operands are element indices, so mixing element offsets with a stick-indexed base silently produced wrong memory accesses.
- `byte_address` is now `base_ptr * bytes_per_elem(dtype)` for both HBM and LX.
- `interpreter.execute_function` converts `hbm.allocate()` (stick) to an element index before passing it as the MLIR SSA argument.
- All stick conversion and latency accounting stays in `_MemAccessor` — unchanged from PR #32.

### Source

**`ktir_cpu/dtypes.py`**
- New helper `stick_to_elem_idx(stick, dtype)` — converts a stick index to an element index. Used by the interpreter and tests.

**`ktir_cpu/ir_types.py`**
- `MemRef.base_ptr` documented as element index (both HBM and LX).
- `MemRef.byte_address` changed from `base_ptr * STICK_BYTES` to `base_ptr * bytes_per_elem(dtype)`.

**`ktir_cpu/interpreter.py`**
- `execute_function` now passes `stick_to_elem_idx(stick, dtype)` as the MLIR SSA argument instead of the raw stick.
- Stores stick separately for `hbm.write/read` at allocation and read-back time.

### Tests

**`tests/test_tile.py`, `tests/test_dialects_exec.py`, `tests/test_latency.py`**
- `MemRef(base_ptr=ptr, ...)` where `ptr = hbm.allocate(...)` updated to `MemRef(base_ptr=stick_to_elem_idx(ptr, dtype), ...)`.

**`tests/test_indirect_access.py`**
- Inline MLIR constants converted from stick to element indices:
  - `arith.constant 1 : index` (i32) → `32` (stick 1 = byte 128 = element 32 for i32)
  - `arith.constant 2 : index` (i32) → `64`
  - `arith.constant 3 : index` (f16) → `192`
  - `arith.constant 64 : index` (i32) → `2048`

**`tests/test_distributed_view.py`**
- `_write_strided`: now converts element-index `base_ptr` to stick/byte address before calling `hbm.write`/`lx.write`.
- `_seed_and_run`: output pointer converted from element index to stick for `hbm.write/read`.
- `test_distributed_view_copy_rfc`: seeding updated to match RFC example element-index constants.
- `MemRef(base_ptr=P*_STICK, ...)` updated to `MemRef(base_ptr=stick_to_elem_idx(P*_STICK, DTYPE), ...)`.

**`tests/test_latency_modeling.py`**
- `_patch_seed_lx`: mirrors HBM writes to LX at byte address `stick * STICK_BYTES` instead of the raw stick index, matching what `MemRef.byte_address` now produces.

## Test plan

- [ ] `uv run pytest tests/ -v` — all 775 tests pass, 7 pre-existing xfail unchanged
- [ ] Verify: `add-with-control-flow.mlir` (hardcoded element-index constants) executes correctly once `linalg.add` is implemented (tracked separately)


ktir_cpu/dtypes.py
```diff
@@ -76,6 +76,15 @@ def bytes_per_elem(dtype: str) -> int:
     return int(to_np_dtype(dtype).itemsize)
 
 
+def stick_to_elem_idx(stick: int, dtype: str, stick_bytes: int = 128) -> int:
```
Prasanth-Chatarasi, line 79: Wouldn't you need elem_to_stick as well since the KTIR operates addresses at element granularity and if your memory files are based on sticks?

fabianlim: @Prasanth-Chatarasi thank you for looking at this PR. Our intention was to handle all *byte-level* addressing through the `MemAccessor` interface. I apologize there was no good documentation of this intention. I wrote something in the comment below to attempt to make this more clear. Also you did identify a gap, and I called that out. Im trying of getting rid of the `index_ptr` accounting in interpreter.py as it is difficult to manage, so maybe if we can come up with a better design to address all these frailities would be good. Let me make an issue on this, i discussed this in the below comment

fabianlim:
Reviewer comment on `ktir_cpu/dtypes.py:79`:
> "Wouldn't you need `elem_to_stick` as well since the KTIR operates addresses at
> element granularity and if your memory files are based on sticks?"

Reference: commit `dbdea96` on `Prasanth-Chatarasi/ktir-cpu:pc/enhancements`

---

## 1. Confirming the use cases

Looking at `ktir_cpu/testing/memory_maps.py` and `tests/test_memory_maps.py` in the
reference branch, we believe the two use cases driving the comment are:

**Use case A — seeding HBM from a SEN memory-map dump at a known byte address**
(`write_tensors_to_hbm` + `TestWriteTensorsToHbm`):

The tensor lives at a **specific, externally-specified byte address** (not
allocator-assigned). Example from the reference test:

```python
# seed two tensors at known byte addresses
write_tensors_to_hbm(hbm, {"A": a, "B": b}, descs)

# verify by reading back at the same byte addresses
got_a = hbm.read(128000, 64, "f16")
got_b = hbm.read(128128, 64, "f16")
```

**Use case B — host read-back of kernel outputs via a byte address**
(`interpreter.py` in commit `dbdea96`):

After the kernel runs, the host needs to read the result tensors back out of HBM.
In the reference commit `allocate()` is treated as returning a byte address, and
that byte address is saved and used directly for read-back:

```python
# commit dbdea96 interpreter.py
byte_ptr = self.memory.hbm.allocate(tensor.nbytes)   # treated as byte address
self.memory.hbm.write(byte_ptr, tensor)
input_byte_ptrs[arg_name] = byte_ptr                 # saved for read-back
input_ptrs[arg_name] = byte_ptr // itemsize          # element index → kernel
...
# read-back
output_data = self.memory.hbm.read(byte_ptr, n_elements, dtype)
```

---

## 2. How `_MemAccessor` handles byte-level addressing today

### The abstraction

`_MemAccessor` (`ktir_cpu/ops/memory_ops.py`) is already the single place in the
codebase that owns byte-level addressing for all memory spaces. Its interface is
**always byte-addressed** — callers hand it a byte address and call `.read(n,
dtype)` or `.write(data)`, and it handles the HBM-specific stick decomposition
internally:

```
caller supplies: byte_addr
                    │
             _MemAccessor
            ┌──────┴──────────────────────────────────┐
            │  HBM:  stick, intra = divmod(byte_addr, 128)
            │        → hbm.read(stick, n, dtype, intra_byte=intra)
            │
            │  LX:   → lx.read(byte_addr, n, dtype)
            └─────────────────────────────────────────┘
```

The HBM simulator itself remains stick-indexed; `_MemAccessor` is the byte→stick
translation layer. LX is already byte-addressed so it passes through unchanged.

> NOTE: the HBMSimulator has accepts `intra_byte` in the `read/write` methods for intra stick addressing.

### How it is used today

Every load and store in the interpreter goes through this chain:

```
ktdp.construct_memory_view   ktdp.construct_access_tile   ktdp.load / ktdp.store
        │                             │                           │
        ▼                             ▼                           ▼
  MemRef                         TileRef                   MemoryOps.load/store
  .base_ptr = element idx        .base_ptr = byte addr          │
  .byte_address property ────────────────────────────►  _MemAccessor(byte_addr)
  (= base_ptr * itemsize)                                        │
                                                          hbm.read(stick, ...)
                                                          lx.read(byte_addr, ...)
```

`MemRef.base_ptr` is the element-indexed value the MLIR program carries.
`MemRef.byte_address` converts it once (`base_ptr * bytes_per_elem(dtype)`).
`TileRef.base_ptr` (produced by `MemRef.to_tile_ref()`) carries that byte address.
`_MemAccessor` is instantiated with `TileRef.base_ptr` at every `load`/`store`.

### How use case A maps onto `_MemAccessor`

Instead of calling `hbm.write/read` directly with a byte address (which requires
changing the `HBMSimulator` API), `write_tensors_to_hbm` can route through
`_MemAccessor`:

| Reference approach | `_MemAccessor` approach |
|---|---|
| `hbm.write(byte_addr, arr)` | `_MemAccessor(ctx, "HBM", byte_addr).write(arr)` |
| `hbm.read(byte_addr, n, dtype)` | `_MemAccessor(ctx, "HBM", byte_addr).read(n, dtype)` |

Both express "write/read at this byte address". The `_MemAccessor` path works with
the **current `HBMSimulator` API unchanged** — the stick decomposition happens
inside the accessor, not at the call site.

### How use case B maps onto `_MemAccessor`

For the interpreter's read-back path, our PR saves the stick before conversion:

```python
# our PR
stick = self.memory.hbm.allocate(tensor.nbytes)    # → stick
self.memory.hbm.write(stick, tensor)
input_sticks[arg_name] = stick                     # saved for read-back
input_ptrs[arg_name] = _stick_to_elem_idx(stick, dtype)   # element index → kernel

# read-back: works only because we saved the original stick
output_data = self.memory.hbm.read(input_sticks[arg_name], n, dtype)
```

> ⚠️ **Gap identified by reviewer:** `input_sticks` in `interpreter.py` is brittle
> — it only covers allocator-assigned addresses from the current `execute_function`
> call. Two concrete failure modes:
>
> 1. A caller that seeds HBM at an externally-specified element address (use case A)
>    has no read-back path — the element pointer is a one-way conversion and
>    `input_sticks` does not cover it.
> 2. The ptr accounting (`input_sticks`, `input_byte_ptrs`) lives in
>    `interpreter.py` rather than in the simulators themselves. `HBMSimulator` and
>    `LXScratchpad` are unaware of what has been written to them, so there is no
>    single source of truth for "what is at address X" — each caller has to keep its
>    own shadow map.
>
> These are the problems already tracked in **#102** (bootstrap placement API and
> allocation-aware `next_ptr`). The second point — ptr accounting inside the
> simulators — is broader than #102's scope and should be a separate issue.

The immediate fix within this PR's scope is to save `byte_ptr` instead of `stick`,
and route read-back through `_MemAccessor`. This closes failure mode 1 without
requiring simulator changes, and aligns with the reference design:

```python
# proposed — closes failure mode 1
stick = self.memory.hbm.allocate(tensor.nbytes)
byte_ptr = stick * HBMSimulator.STICK_BYTES
self.memory.hbm.write(stick, tensor)
input_byte_ptrs[arg_name] = byte_ptr               # byte addr saved for read-back
input_ptrs[arg_name] = byte_ptr // itemsize        # element index → kernel

# read-back via _MemAccessor — works for any stick-aligned byte address
_MemAccessor(ctx, "HBM", byte_ptr).read(n, dtype)
```

Failure mode 2 (ptr accounting in the simulators themselves) is left for the
follow-up issue.

---

## Motivation for keeping `HBMSimulator` stick-indexed

The design has a deliberate separation of concerns between two layers:

```
  Callers (ktdp_ops, interpreter, memory_maps)
       │  byte addresses only
       ▼
  ┌─────────────────────────────────────────────────────────────────┐
  │  _MemAccessor                                                   │
  │                                                                 │
  │  • Byte-addressed interface for all memory spaces               │
  │  • Sole producer of stick counts (count_sticks, read_scattered) │
  │  │    stick = addr // 128 ; intra = addr % 128                  │
  │  • Stamps unique_sticks onto Tile for latency accounting        │
  │  • Routes to correct simulator (HBM or LX) transparently       │
  └──────────────┬──────────────────────────┬───────────────────────┘
                 │ stick + intra_byte        │ byte addr
                 ▼                           ▼
       ┌──────────────────┐       ┌──────────────────┐
       │  HBMSimulator    │       │  LXScratchpad    │
       │                  │       │                  │
       │  Memory simulator│       │  Memory simulator│
       │  stick-indexed   │       │  byte-indexed    │
       │  models 128-byte │       │  models per-core │
       │  DMA granularity │       │  scratchpad      │
       └──────────────────┘       └──────────────────┘
                 │
                 ▼
       ┌──────────────────┐
       │  latency.py      │
       │                  │
       │  Reads           │
       │  Tile.unique_    │
       │  sticks from     │
       │  _MemAccessor    │
       │  charges traffic │
       │  at stick        │
       │  granularity     │
       └──────────────────┘
```

`_MemAccessor` is the interface layer: it presents a uniform byte-addressed API to
all callers and owns the byte→stick translation. `HBMSimulator` is purely a memory
simulator — it models HBM's 128-byte DMA granularity and exposes `intra_byte` to
represent sub-stick alignment accurately. Neither layer should take on the other's
responsibility.

Changing `HBMSimulator` to byte-addressed would collapse these two concerns into
one: the simulator would have to do the stick decomposition that `_MemAccessor`
currently owns, and callers wanting stick counts for latency would have no clean
seam to intercept them.

## Question for the reviewer

`HBMSimulator.read/write` currently takes a stick index; `memory.py` is not changed
in commit `dbdea96`. The reference tests call `hbm.read(128000, ...)` and
`hbm.write(128000, ...)` treating `128000` as a byte address — was the intent to
convert `HBMSimulator` to byte-addressed as a follow-up?

fabianlim (commented):
~@Prasanth-Chatarasi i posted #116 that would help to solve the read-back issue~

~@nwang-ibm the new issue may help you with #109~  Sorry i realized the new issue is not needed, Im proposing a much easier solution https://github.com/torch-spyre/ktir-cpu/pull/110/commits/384180f9709ecc37b12ca540a21e5fc8992015cb

fabianlim:
Following up on the earlier response — we've now landed the concrete changes in commit `1e3ba31`.

**The two helpers** (`hbm_read`, `hbm_write` in `ktir_cpu/ops/memory_ops.py`) are thin wrappers that expose a byte-addressed interface to HBM without changing `HBMSimulator`'s stick-indexed API. They perform the same `divmod(byte_addr, STICK_BYTES)` decomposition that `_MemAccessor` does internally:

\`\`\`python
def hbm_read(hbm, byte_addr, n_elements, dtype):
    stick, intra = divmod(byte_addr, HBMSimulator.STICK_BYTES)
    return hbm.read(stick, n_elements, dtype, intra_byte=intra)

def hbm_write(hbm, byte_addr, data):
    stick, intra = divmod(byte_addr, HBMSimulator.STICK_BYTES)
    hbm.write(stick, data, intra_byte=intra)
\`\`\`

**Use case A** (`write_tensors_to_hbm` with a known byte address) is now directly expressible: `hbm_write(hbm, desc.byte_addr, arr)` / `hbm_read(hbm, desc.byte_addr, n, dtype)` — no stick arithmetic at the call site.

**Use case B** (interpreter read-back) is fixed in `interpreter.py`: we now save `byte_addr = stick * STICK_BYTES` alongside the element-index `input_ptrs`, and read-back routes through `hbm_read`. The brittle `input_sticks`/`input_dtypes` shadow maps are gone.

`MemRef.base_ptr` stays element-indexed (the MLIR pointer operand contract); `_MemAccessor` remains the sole owner of byte→stick translation for load/store.

ktir_cpu/ops/memory_ops.py
```diff
@@ -167,6 +167,18 @@ def write(self, data: np.ndarray) -> None:
...
 
 
+def hbm_read(hbm: "HBMSimulator", byte_addr: int, n_elements: int, dtype: str) -> np.ndarray:
+    """Read n_elements of dtype from HBM at byte_addr (byte-addressed)."""
+    stick, intra = divmod(byte_addr, HBMSimulator.STICK_BYTES)
+    return hbm.read(stick, n_elements, dtype, intra_byte=intra)
+
+
+def hbm_write(hbm: "HBMSimulator", byte_addr: int, data: np.ndarray) -> None:
+    """Write data to HBM at byte_addr (byte-addressed)."""
```
nwang-ibm, line 177: The interpreter calls hbm_write(..., tensor.flatten()), should a note/assertion be added that tensor is 1d?

fabianlim: @nwang-ibm good catch, added an assertion in the latest push — thanks!

nwang-ibm (approved):

PR #111, opened 2026-06-12
Fix construct_distributed_memory_view dynamic result shape

### LINKED ISSUE #99: [Feature, DistributedMemView] Regex parser: accept dynamic dims in ktdp.construct_distributed_memory_view result type
## Background

Surfaced during #80 review. The regex parser for `ktdp.construct_distributed_memory_view` rejects `?` in the result type, even though the dialect spec permits it: upstream TableGen ([`ktir-mlir-frontend`](https://github.com/torch-spyre/ktir-mlir-frontend)) types the result as `AnyMemRef` and the verifier checks rank and element type only.

## Problem

`parse_construct_distributed_memory_view` (`ktir_cpu/dialects/ktdp_ops.py:471`) forces every dim to a literal integer:

```python
shape = tuple(int(p) for p in parts[:-1])
```

Writing `memref<64x?xf16>` as the result raises `ValueError` at parse time. The sibling op `construct_memory_view` already handles `?` correctly in the same file (`ktdp_ops.py:339-382`): parse `?` as `None`, validate `sizes:` count vs dim count, resolve each `None` from the corresponding `sizes:` SSA name. The same idea applies here, except resolution comes from the operand memref shapes rather than a separate `sizes:` attribute (the global shape is composed from partition shapes).

The current limitation forces fixtures with parametric global shapes to be split into one `.mlir` per shape value (e.g. `examples/rfc/distributed-view-copy-dynamic-s16.mlir` + `-s32.mlir`).

## Scope

In `parse_construct_distributed_memory_view` (`ktdp_ops.py:408`):

1. Parse each shape part as `None if p == "?" else int(p)`.
2. For each `None` slot, resolve from the operand memrefs' shapes along that axis (the global shape's dim is the union along that axis). Operand shapes are already SSA-resolved at the time `construct_memory_view` parsed each input.
3. If a `None` slot's resolution is ambiguous across operands, raise with a clear error.

In `ktdp__construct_distributed_memory_view` (`ktdp_ops.py:101`): re-use the existing partition shape validation against the now-possibly-symbolic `shape` attribute.

## Out of scope

- `MLIRFrontendParser` adapter coverage for distributed memory views — tracked in #87. Independent gap: even with this fix landed, the frontend adapter still has no handler for `ktdp.construct_distributed_memory_view`.
- Per-core LX routing (`<LX, core = N>`) — separate gap noted in `examples/rfc/distributed-view-copy.mlir`.

## Test plan

- Parametrize a fixture over multiple `s0` values that share one `.mlir` file with a `?` in the distributed-view result type — collapses #80's `-s16` / `-s32` pair to a single file, which doubles as a regression test that the parser fix actually works.
- Negative test: `?` in the result with no matching `?` in any operand still raises, with a message naming the unresolved axis.
- Existing concrete-shape fixtures (`distributed-view-copy.mlir`) continue to pass.

## References

- `ktir_cpu/dialects/ktdp_ops.py` — `parse_construct_distributed_memory_view:408` (the gap), `parse_construct_memory_view:278-382` (the pattern to mirror), `ktdp__construct_distributed_memory_view:101` (handler reuse).
- Upstream dialect: [`ktir-mlir-frontend`](https://github.com/torch-spyre/ktir-mlir-frontend) — `KTDPOps.td` `AnyMemRef` typing.
- #87 (frontend adapter, separate gap).

WarningRan:
Unblocks #80, which is hold-pending #99 per https://github.com/torch-spyre/ktir-cpu/pull/80#issuecomment-4655737273.

## Problem

`parse_construct_distributed_memory_view` rejects `?` in the result memref type. #80 had to ship `-s16.mlir` + `-s32.mlir` as two fixtures because of this; once `?` parses, the pair collapses to one `?`-bearing fixture in #80 itself. The dialect spec already permits `?` (upstream `KTDPOps.td` types the result as `AnyMemRef`); the restriction is purely a ktir-cpu parser limitation.

## Fix (parser entry + handler resolution)

Parser stores `None` for `?` in the shape attr. Handler resolves each `None` axis via `max(partition.coordinate_set.hi[axis])` — partitions are in global coords (see `MemoryOps.distributed_tile_access`), so this is the per-axis global upper bound.

### Why `coordinate_set.hi` instead of operand shapes (deviates from issue Scope step 2)

#80 splits axis-1 into two `memref<64x?xf16>` partitions (`sizes: [64, %s0]`) placed at `[0, s0)` and `[s0, 2·s0)`. Both partitions have local `shape[1] = s0`, so `max(operand.shape[1]) = s0` — wrong, the global is `2·s0`. Summing also fails in general: a row-split where both partitions span full axis-1 width `W` sums to `2W` vs global `W`. Neither `max` nor `sum` of local shapes is correct — only `coordinate_set.hi` carries placement, and `max(hi[1]) = max(s0, 2·s0) = 2·s0` is the right answer. This PR therefore resolves from `coordinate_set`, not operand shapes.

(The `s0` / `2·s0` notation describes the partition layout pre-specialization. `construct_memory_view` upstream specialises symbolic `BoxSet` partitions against runtime symbols before this handler runs, so `cs.hi` is concrete when `_resolve_dynamic_dist_shape` sees it — see the concrete-partition Constraint below.)

### Why `max(hi)` (and not `max(hi) - min(lo)`)

`max(hi)` is the smallest shape that satisfies `BoxSet.enumerate`'s `shape >= max(hi)` precondition for every partition; `max(hi) - min(lo)` would under-size the memref and break slow-path enumeration on tilings with `min(lo) > 0`. Coverage is not enforced at this layer — non-origin tilings (`min(lo) > 0`) and interior gaps produce a memref where uncovered global coordinates surface as `no partition covers` at access time via `MemoryOps.distributed_tile_access`. This function's job is shape resolution; coverage is the access-time contract.

### Constraints (each → `ValueError` naming partition / axis)

- partition is axis-aligned `BoxSet` (`AffineSet` rejected as an implementation cut — extracting per-axis upper bound from a general `AffineSet` is bounding-box derivation that this path does not do; tracked under #74)
- partition `coordinate_set` is concrete (non-concrete trips `ValueError` so the boundary check is loud under `python -O` too)
- partition rank matches result memref rank
- at least one partition (zero-partition dynamic-shape view rejected)
- per-axis `max(hi) > 0` on every dynamic axis (rejected only when every partition is non-positive on that axis; individual zero-extent partitions mixed with normal ones are tolerated — they behave like gaps and surface at access time)

## Out of scope (deliberately)

- `-s16/-s32` fixture collapse — ships in #80, not here. This PR only opens the parser door; #80 retains its bit-exact e2e regression.
- Frontend `MLIRFrontendParser` adapter for distributed views — #87.
- Per-core LX routing in distributed views.
- Overlap / gap detection across partitions — coverage is deferred to access time uniformly; uncovered global coords (whether from non-origin tilings, interior gaps, or overlap) surface as `no partition covers` from `distributed_tile_access`.

## Tests

- Parse: `memref<64x?xf16>` / `memref<?x?xf16>` round-trip to `None`-bearing shape; concrete-shape regression sanity.
- Handler positive: 2-partition `[0,16)+[16,32) → 32`; RFC's 3-partition 192×64 with leading axis dynamic; non-origin `[10,20)+[20,30) → 30` (resolves to `max(hi)`; coverage deferred to access time).
- Handler negative: AffineSet partition / rank mismatch / `max(hi)≤0` / non-concrete `coordinate_set` (each `ValueError` with specific match).

Full suite: 927 passed, 1 skipped (pre-existing collection-time skip, unrelated), 14 xfailed, 0 regressions.

### DIFF
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -15,6 +15,7 @@
 """KTDP dialect handlers — grid, memory view, access tile, load/store."""
 
 import re
+from typing import Optional, Sequence, Tuple
 
 from .ktdp_helpers import attach_reshape, parse_subscript_expr
 from ..affine import BoxSet
@@ -102,6 +103,18 @@ def ktdp__construct_distributed_memory_view(op, context, env):
     ``coordinate_set`` (= B_i in global coords).  The op does not allocate
     or move data — partition resolution at access time is performed by
     ``MemoryOps.distributed_tile_access``.
+
+    Dynamic dims (``None`` entries in ``shape``, parsed from ``?`` in the
+    result memref type) are resolved here by taking ``max`` of each
+    partition's upper bound along that axis.  Partition coordinate sets
+    are in global absolute coordinates (see
+    ``MemoryOps.distributed_tile_access`` which intersects them with
+    global ``xA_box`` directly), so ``max(B_i.hi)`` is the global extent.
+    Resolution requires axis-aligned ``BoxSet`` partitions; ``AffineSet``
+    is rejected because ``max(upper_bounds)`` recovers the global extent
+    only for orthogonal-bound boxes, not for general affine constraint
+    sets (where the bounding-box upper bound can over-approximate the
+    actual tensor extent).
     """
     partitions = [context.get_value(name) for name in op.operands]
     for i, p in enumerate(partitions):
@@ -114,11 +127,73 @@ def ktdp__construct_distributed_memory_view(op, context, env):
         raise ValueError(
             "construct_distributed_memory_view: missing required attributes 'shape'/'dtype'"
         )
+    raw_shape = tuple(op.attributes["shape"])
+    shape = _resolve_dynamic_dist_shape(raw_shape, partitions)
     return DistributedMemRef(
         partitions=partitions,
-        shape=tuple(op.attributes["shape"]),
+        shape=shape,
         dtype=op.attributes["dtype"],
     )
+
+
+def _resolve_dynamic_dist_shape(
+    raw_shape: Tuple[Optional[int], ...],
+    partitions: Sequence[MemRef],
+) -> Tuple[int, ...]:
+    """Resolve ``None`` entries in a distributed-view shape via partition union.
+
+    Concrete ``raw_shape`` returns unchanged.  For each ``None`` axis:
+    ``max(partition.coordinate_set.hi[axis])`` across all partitions, where
+    ``hi`` is in global coordinates.  Requires every partition's
+    ``coordinate_set`` to be a fully-resolved ``BoxSet``;
+    ``construct_memory_view`` specialises symbolic ``BoxSet`` partitions
+    against runtime symbols before they reach this point.
+    """
+    if not any(s is None for s in raw_shape):
+        return tuple(raw_shape)  # type: ignore[arg-type]
+    if not partitions:
+        raise ValueError(
+            "construct_distributed_memory_view: dynamic-shape result "
+            "requires at least one partition to derive the global extent"
+        )
+    # Partition pre-flight runs once even with multiple `None` axes.  AffineSet
+    # is rejected because non-orthogonal constraints can make the bounding-box
+    # upper bound an over-approximation rather than the tensor's true extent,
+    # so a single `max(hi)` doesn't recover the global shape.
+    ndim = len(raw_shape)
+    boxes = []
+    for i, part in enumerate(partitions):
+        cs = part.coordinate_set
+        if not isinstance(cs, BoxSet):
+            raise ValueError(
+                f"construct_distributed_memory_view: dynamic-shape result "
+                f"requires axis-aligned BoxSet partitions; partition {i} is "
+                f"{type(cs).__name__}, whose union extent is not recoverable "
+                f"by max(upper_bounds) on a non-axis-aligned constraint set"
+            )
+        assert cs.is_concrete, (
+            f"construct_distributed_memory_view: partition {i} coordinate_set "
+            f"is not concrete; construct_memory_view should have specialised "
+            f"it against runtime symbols before reaching this handler"
+        )
+        if cs.n_dims != ndim:
+            raise ValueError(
+                f"construct_distributed_memory_view: partition {i} has rank "
+                f"{cs.n_dims}, expected {ndim} to match result memref rank"
+            )
+        boxes.append(cs.hi)
+    resolved = list(raw_shape)
+    for axis, dim in enumerate(raw_shape):
+        if dim is not None:
+            continue
+        global_hi = max(hi[axis] for hi in boxes)
+        if global_hi <= 0:
+            raise ValueError(
+                f"construct_distributed_memory_view: could not resolve dynamic "
+                f"axis {axis}: all partitions have non-positive upper bound"
+            )
+        resolved[axis] = global_hi
+    return tuple(resolved)  # type: ignore[return-value]
 
 
 @register("ktdp.construct_access_tile")
@@ -464,14 +539,19 @@ def parse_construct_distributed_memory_view(op_text, parse_ctx: ParseContext):
             "has no dimensions"
         )
     dtype = parts[-1]
-    shape = tuple(int(p) for p in parts[:-1])
+    # '?' marks a dynamic dim resolved by the handler from the partitions'
+    # coordinate sets (max upper bound per axis); see
+    # ktdp__construct_distributed_memory_view.
+    shape = tuple(None if p == "?" else int(p) for p in parts[:-1])
 
     return Operation(
         result=result_name,
         op_type="ktdp.construct_distributed_memory_view",
         operands=operands,
         attributes={"shape": shape, "dtype": dtype},
-        result_type=f"memref<{'x'.join(str(s) for s in shape)}x{dtype}>"
+        result_type=(
+            f"memref<{'x'.join('?' if s is None else str(s) for s in shape)}x{dtype}>"
+        ),
     )
 
 
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -15,7 +15,7 @@
 """KTDP dialect handlers — grid, memory view, access tile, load/store."""
 
 import re
-from typing import Optional, Sequence, Tuple
+from typing import List, Optional, Sequence, Tuple
 
 from .ktdp_helpers import attach_reshape, parse_subscript_expr
 from ..affine import BoxSet
@@ -142,25 +142,35 @@ def _resolve_dynamic_dist_shape(
 ) -> Tuple[int, ...]:
     """Resolve ``None`` entries in a distributed-view shape via partition union.
 
-    Concrete ``raw_shape`` returns unchanged.  For each ``None`` axis:
-    ``max(partition.coordinate_set.hi[axis])`` across all partitions, where
-    ``hi`` is in global coordinates.  Requires every partition's
-    ``coordinate_set`` to be a fully-resolved ``BoxSet``;
-    ``construct_memory_view`` specialises symbolic ``BoxSet`` partitions
-    against runtime symbols before they reach this point.
+    Concrete ``raw_shape`` returns unchanged.  For each ``None`` axis the
+    resolved extent is ``max(hi[axis])`` across all partitions, where
+    ``hi`` is in global coordinates.  This is the smallest shape that
+    satisfies ``BoxSet.enumerate``'s ``shape >= max(hi)`` precondition
+    for every partition; ``max(hi) - min(lo)`` would under-size the
+    memref and break slow-path enumeration on tilings with
+    ``min(lo) > 0``.
+
+    Coverage is not enforced here.  Tilings with ``min(lo) > 0``
+    (non-origin) or interior gaps between partitions produce a memref
+    where uncovered global coordinates surface as ``no partition
+    covers`` at access time via ``MemoryOps.distributed_tile_access``.
+    This function's job is shape resolution; coverage is the
+    access-time contract.
+
+    Requires every partition's ``coordinate_set`` to be a fully-resolved
+    ``BoxSet``; ``construct_memory_view`` specialises symbolic ``BoxSet``
+    partitions against runtime symbols before they reach this point.
 
     Scope cut — ``AffineSet`` partitions are intentionally rejected on this
     path only.  ``construct_distributed_memory_view`` itself accepts
     ``AffineSet`` partitions in the concrete-shape path: they are stored
     in ``DistributedMemRef`` and dispatched at access time via
     ``find_partition`` / ``.contains()`` exactly like ``BoxSet``.  The
-    rejection here is specific to dynamic-shape *resolution*: the
-    ``max(upper_bounds)`` strategy only recovers the true extent for
-    axis-aligned ``BoxSet`` partitions; on a non-axis-aligned constraint
-    set the bounding-box upper bound over-approximates.  Concrete-shape
-    callers are unaffected.  Lifting this would require a different
-    resolution strategy (e.g. enumerating the set or computing
-    ``global_extent()`` per axis).
+    rejection here is an implementation cut, not a correctness one:
+    ``BoxSet`` exposes per-axis ``.hi`` directly, while extracting a
+    per-axis upper bound from a general ``AffineSet`` is bounding-box
+    derivation that this path does not do — tracked under #74.
+    Concrete-shape callers are unaffected.
     """
     if not any(s is None for s in raw_shape):
         return tuple(raw_shape)  # type: ignore[arg-type]
@@ -169,12 +179,9 @@ def _resolve_dynamic_dist_shape(
             "construct_distributed_memory_view: dynamic-shape result "
             "requires at least one partition to derive the global extent"
         )
-    # Partition pre-flight runs once even with multiple `None` axes.  AffineSet
-    # is rejected because non-orthogonal constraints can make the bounding-box
-    # upper bound an over-approximation rather than the tensor's true extent,
-    # so a single `max(hi)` doesn't recover the global shape.
+    # Single pass: validate each partition and track max(hi) per dynamic axis.
     ndim = len(raw_shape)
-    boxes = []
+    max_hi: List[Optional[int]] = [None] * ndim
     for i, part in enumerate(partitions):
         cs = part.coordinate_set
         if not isinstance(cs, BoxSet):
@@ -188,28 +195,35 @@ def _resolve_dynamic_dist_shape(
                 f"find_partition at access time).  Declare the result "
                 f"memref with concrete dims to use this partition."
             )
-        assert cs.is_concrete, (
-            f"construct_distributed_memory_view: partition {i} coordinate_set "
-            f"is not concrete; construct_memory_view should have specialised "
-            f"it against runtime symbols before reaching this handler"
-        )
+        if not cs.is_concrete:
+            # `ValueError` (not `assert`) so the failure is loud under `python -O`.
+            raise ValueError(
+                f"construct_distributed_memory_view: partition {i} coordinate_set "
+                f"is not concrete; construct_memory_view should have specialised "
+                f"it against runtime symbols before reaching this handler"
+            )
         if cs.n_dims != ndim:
             raise ValueError(
                 f"construct_distributed_memory_view: partition {i} has rank "
                 f"{cs.n_dims}, expected {ndim} to match result memref rank"
             )
-        boxes.append(cs.hi)
+        for axis in range(ndim):
+            if raw_shape[axis] is not None:
+                continue
+            hi_a = cs.hi[axis]
+            if max_hi[axis] is None or hi_a > max_hi[axis]:
+                max_hi[axis] = hi_a
     resolved = list(raw_shape)
     for axis, dim in enumerate(raw_shape):
         if dim is not None:
             continue
-        global_hi = max(hi[axis] for hi in boxes)
-        if global_hi <= 0:
+        assert max_hi[axis] is not None  # ≥1 partition × every dynamic axis traversed
+        if max_hi[axis] <= 0:
             raise ValueError(
-                f"construct_distributed_memory_view: could not resolve dynamic "
-                f"axis {axis}: all partitions have non-positive upper bound"
+                f"construct_distributed_memory_view: could not resolve "
+                f"dynamic axis {axis}: max(hi)={max_hi[axis]} is non-positive"
             )
-        resolved[axis] = global_hi
+        resolved[axis] = max_hi[axis]
     return tuple(resolved)  # type: ignore[return-value]
 
 
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -15,7 +15,6 @@
 """KTDP dialect handlers — grid, memory view, access tile, load/store."""
 
 import re
-from typing import List, Optional, Sequence, Tuple
 
 from .ktdp_helpers import attach_reshape, parse_subscript_expr
 from ..affine import BoxSet
@@ -104,17 +103,11 @@ def ktdp__construct_distributed_memory_view(op, context, env):
     or move data — partition resolution at access time is performed by
     ``MemoryOps.distributed_tile_access``.
 
-    Dynamic dims (``None`` entries in ``shape``, parsed from ``?`` in the
-    result memref type) are resolved here by taking ``max`` of each
-    partition's upper bound along that axis.  Partition coordinate sets
-    are in global absolute coordinates (see
-    ``MemoryOps.distributed_tile_access`` which intersects them with
-    global ``xA_box`` directly), so ``max(B_i.hi)`` is the global extent.
-    Resolution requires axis-aligned ``BoxSet`` partitions; ``AffineSet``
-    is rejected because ``max(upper_bounds)`` recovers the global extent
-    only for orthogonal-bound boxes, not for general affine constraint
-    sets (where the bounding-box upper bound can over-approximate the
-    actual tensor extent).
+    The global shape is assembled and validated against the partitions in
+    :meth:`DistributedMemRef.__post_init__`: dynamic dims (``None``, parsed
+    from ``?`` in the result type) are filled with the per-axis data span
+    ``max(hi) - min(lo)``, and concrete dims are validated against it (see
+    ``_assemble_dist_extent``).
     """
     partitions = [context.get_value(name) for name in op.operands]
     for i, p in enumerate(partitions):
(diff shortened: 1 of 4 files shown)


WarningRan:
cc @fabianlim — this unblocks #80 per [your hold-pending comment](https://github.com/torch-spyre/ktir-cpu/pull/80#issuecomment-4655737273).

fabianlim (changes requested):
The `_resolve_dynamic_dist_shape` function rejects `AffineSet` partitions, but this restriction is framed as an implementation detail of the resolution strategy rather than a deliberate, documented design choice. Since `construct_distributed_memory_view` fully supports `AffineSet` partitions in the concrete-shape path (they are stored in `DistributedMemRef` and dispatched via `find_partition` → `.contains()` just fine), the dynamic-shape path should explicitly acknowledge this as an intentional scope limitation and clarify what would be needed to lift it.

Specifically:

1. **`_resolve_dynamic_dist_shape` should have a docstring-level note** that this is a deliberate scope cut for dynamic-shape resolution only — `AffineSet` partitions are valid for `construct_distributed_memory_view` when the result shape is concrete, and the `AffineSet` rejection here is solely because `max(bounding_box.hi)` over-approximates the true extent for non-axis-aligned constraint sets. Lifting this would require a different resolution strategy (e.g., enumerating the set or computing `global_extent()` per axis).

2. **The `ValueError` message** currently says `whose union extent is not recoverable by max(upper_bounds) on a non-axis-aligned constraint set` — this reads as an internal justification. It should instead make clear that this is a known limitation of the dynamic-shape path specifically, and that the concrete-shape path handles `AffineSet` partitions without issue. A future PR can add `AffineSet`-aware dynamic resolution.

The distinction matters because a reader encountering this error needs to understand: (a) their `AffineSet` partition is not inherently invalid for `construct_distributed_memory_view`, and (b) the issue is specific to dynamic-shape resolution, not the op itself.

WarningRan:
Agreed — scoped the docstring + error message to make the dynamic-only nature explicit. `_resolve_dynamic_dist_shape` docstring now has a "Scope cut" paragraph spelling out that AffineSet rejection is specific to dynamic-shape resolution, that concrete-shape callers are unaffected, and that lifting the limitation would need a different resolution strategy (enumerating the set / per-axis `global_extent()`). The `ValueError` now points readers at the concrete-shape path as the workaround.

fabianlim:
@WarningRan undersatnd that you would like to cut the scope of #99  in this PR, but would the scope cut affect #80. If your e2e tests also involve affine sets wouldnt you need that part to be first resolved

WarningRan:
@fabianlim My understanding is that the scope cut shouldn't affect #80, but please correct me if I'm reading this wrong.

#80's partition `affine_set<>`s are axis-aligned with both bounds pinned per axis, so I believe `parse_affine_set` lowers them to symbolic `BoxSet` via `BoxSet.try_from_affine_set` at parse time, and `construct_memory_view`'s handler then specialises them to concrete `BoxSet` against the runtime `s0` binding. By the time `_resolve_dynamic_dist_shape` runs, the partition `coordinate_set`s should be concrete `BoxSet` and take the `max(hi)` happy path.

The `AffineSet` rejection in this PR fires only on genuinely non-axis-aligned partitions paired with `?` result type — defensive against a spec-allowed but currently unused combination, as I understand it.

fabianlim:
@WarningRan the confusion i had is #80 's title seems like it will cover both box and affine, but the description is limited to affine. and then #74 is actually general for affine / box

WarningRan:
Good point — re-reading #80's description, you're right that it reads ambiguous: the prose says "symbolic-`BoxSet`" but the code blocks all show `affine_set<>` syntax, so a skim could come away thinking the PR is `AffineSet`-scoped.

Honestly should have clarified this earlier - I'd been holding off because #80 has been pending #99/#111 with the planned post-merge fixture collapse looming over the body anyway, so a rewrite kept feeling premature.  Want me to push the description fix and a "symbolic-BoxSet" title tightening on #80 now, or fold them into the collapse rewrite once #111 lands?

fabianlim:
@WarningRan just scope down #80 to only the box-set case for now.

WarningRan:
Done - #80 title and description now scoped to the `BoxSet` path; the `AffineSet` path stays under #74 for now.

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -114,13 +127,92 @@ def ktdp__construct_distributed_memory_view(op, context, env):
...
+                f"construct_distributed_memory_view: partition {i} has rank "
+                f"{cs.n_dims}, expected {ndim} to match result memref rank"
+            )
+        boxes.append(cs.hi)
+    resolved = list(raw_shape)
+    for axis, dim in enumerate(raw_shape):
+        if dim is not None:
+            continue
+        global_hi = max(hi[axis] for hi in boxes)
+        if global_hi <= 0:
```
fabianlim: `max(hi[axis])` assumes all partitions are origin-anchored (`min(lo[axis]) == 0`), making `max(hi)` equal to the global extent. For a non-origin-anchored tiling the correct formula is `max(hi[axis]) - min(lo[axis])`.

Can you confirm the assumptions on the global axes of the partitions here? Specifically: is it guaranteed that every tiling starts from 0 on each axis? If not, this silently produces an oversized memref shape without raising an error.

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -114,13 +127,92 @@ def ktdp__construct_distributed_memory_view(op, context, env):
...
+                f"resolution does not support {type(cs).__name__} "
+                f"partitions (partition {i} on this op).  This is a "
+                f"limitation of the dynamic-shape path only — the "
+                f"concrete-shape path of construct_distributed_memory_view "
+                f"accepts AffineSet partitions normally (they dispatch via "
+                f"find_partition at access time).  Declare the result "
+                f"memref with concrete dims to use this partition."
+            )
+        assert cs.is_concrete, (
+            f"construct_distributed_memory_view: partition {i} coordinate_set "
```
fabianlim: Under `python -O` this assert is compiled out. A non-concrete `BoxSet` (symbolic string in `hi`) would then pass silently, reach `max(hi[axis] for hi in boxes)` below, and either produce a wrong resolved shape or crash with a `TypeError` deep inside rather than a clean error at the source.

If this invariant can be violated by malformed input (not just an internal logic bug), consider promoting it to a `ValueError` so the failure is clean under `-O` as well.

ktir_cpu/dialects/ktdp_ops.py
```diff
@@ -114,13 +127,92 @@ def ktdp__construct_distributed_memory_view(op, context, env):
...
+            f"is not concrete; construct_memory_view should have specialised "
+            f"it against runtime symbols before reaching this handler"
+        )
+        if cs.n_dims != ndim:
+            raise ValueError(
+                f"construct_distributed_memory_view: partition {i} has rank "
+                f"{cs.n_dims}, expected {ndim} to match result memref rank"
+            )
+        boxes.append(cs.hi)
+    resolved = list(raw_shape)
```
fabianlim: Only `cs.hi` is collected here, but the per-axis max is computed in a second loop below (`max(hi[axis] for hi in boxes)`). This iterates over all partitions once per dynamic axis — `O(N*D)`. A single pass could track both `max_hi` and `min_lo` per axis while validating each partition, avoiding the second loop and keeping the data needed if the extent formula is corrected to `max(hi) - min(lo)`.

ktir_cpu/dialects/ktdp_ops.py
WarningRan: After tracing the addressing path, I went with `max(hi)` rather than `max(hi) - min(lo)` or upfront origin validation. The reasoning, in short:

`distributed_tile_access` indexes by global coordinates (`base_map.eval(indices)`), and `BoxSet.enumerate` requires `shape >= max(hi)` — so `max(hi)` is the smallest shape that satisfies the addressing contract; `max(hi) - min(lo)` would under-size the memref and break slow-path enumeration on tilings with `min(lo) > 0`. Non-origin tilings under `max(hi)` produce a memref with an uncoverable `[0, min(lo))` prefix; accesses there surface as `no partition covers` at access time, which is the same failure mode interior gaps and overlap already take. Coverage is uniformly the access-time contract; the resolution layer doesn't validate it.

Spelled out under "Why `max(hi)` (and not `max(hi) - min(lo)`)" in the description.

ktir_cpu/dialects/ktdp_ops.py
WarningRan: Done — promoted to `ValueError` in front of the per-axis access. The defensive test had a `@pytest.mark.skipif(not __debug__, ...)` because of the `assert`; that's gone now and the test runs in all modes.

ktir_cpu/dialects/ktdp_ops.py
WarningRan: Done — single pass now. Validation (type / concrete / rank) and per-axis `max(hi)` tracking happen in one walk over `partitions`, and the `boxes` collector + second loop are gone. The version went with `max(hi)` rather than `max(hi) - min(lo)`, so `min_lo` ends up not being tracked — see comment 1 / the description for the addressing-path reasoning.

fabianlim:
A few suggestions on `_resolve_dynamic_dist_shape`:

**Background — the actual flow**

```
parse_construct_distributed_memory_view
  → shape = (64, None)  stored in op.attributes["shape"]

ktdp__construct_distributed_memory_view  (handler)
  → raw_shape = tuple(op.attributes["shape"])   # (64, None)
  → shape = _resolve_dynamic_dist_shape(raw_shape, partitions)
  → DistributedMemRef(partitions=..., shape=shape, ...)
```

**1. Rename**

`_resolve_dynamic_dist_shape` is misleading — the partitions coming in are already concrete `BoxSet`s (specialised upstream by `construct_memory_view`). The function is assembling the global extent from the partition layout, not resolving runtime dynamics. Suggest: `_assemble_dist_extent` or similar.

**2. Remove the early-exit**

```python
if not any(s is None for s in raw_shape):
    return tuple(raw_shape)   # ← remove this
```

This skips the concrete path entirely, so a declared `memref<20x64xf16>` with a partition whose `hi[0]=30` passes through unchecked.

**3. Move the call — validate before constructing `DistributedMemRef`**

Right now the function is only called when `None` is present. Instead, always run it and use it to validate (or fill) every axis — concrete axes check `declared >= max(hi)`, dynamic axes fill in `max(hi)`. The natural home for this is `DistributedMemRef.__post_init__` in `ir_types.py`, where dtype and partition-presence are already validated. The handler in `ktdp_ops.py` can pass `raw_shape` through; `__post_init__` resolves and checks it.

**4. Unit test that exposes the gap**

```python
def test_construct_distributed_memory_view_concrete_shape_undercount():
    # Partitions cover [0,16) and [16,32) on axis-1: global hi = 32.
    # Declared shape says 20 — smaller than max(hi)=32.
    # Without validation this silently constructs a DistributedMemRef(shape=(64,20))
    # whose enumerate() will crash the moment a partition with hi[1]=32 is passed.
    ctx = _make_ctx()
    ctx.set_value("%a0", self._make_box_partition(lo=(0,0), hi=(64,16), shape=(64,16)))
    ctx.set_value("%a1", self._make_box_partition(lo=(0,16), hi=(64,32), shape=(64,16)))
    with pytest.raises(ValueError, match=r"declared shape.*axis 1.*20.*max\(hi\).*32"):
        _call(
            "ktdp.construct_distributed_memory_view", ctx, _make_env(),
            operands=["%a0", "%a1"],
            attributes={"shape": (64, 20), "dtype": "f16"},
        )
```

We'd expect this test to pass after the fix. If it does not — i.e. the declared dim is accepted even though it's less than `max(hi)` — that confirms the concrete path is silently passing through an inconsistent shape, and we should surface a `ValueError` there too, forcing the MLIR author to write the correct extent explicitly.

fabianlim:
To add context on why suggestions 2–4 matter together:

In the concrete path today, `shape` comes straight from the MLIR text the author wrote — `_resolve_dynamic_dist_shape` early-exits and never touches the partitions. So if the author writes `memref<20x64xf16>` but the partitions actually span `hi[1]=32`, the inconsistency passes through silently into `DistributedMemRef`. It will only surface later (if at all) when `BoxSet.enumerate` is called with that shape and hits `hi[d] > shape[d]`.

The `?` path accidentally does better: it is forced to read the partitions to fill in the missing dims, so it can't be inconsistent in the same way.

Repurposing this function as a check for both paths closes that gap: concrete dims get validated against `max(hi)`, dynamic dims get filled by it. The author's declared shape is no longer trusted blindly — the partitions are the ground truth.

WarningRan:
Thanks — this makes sense to me. Reframing it as "partitions are the ground truth, validate against them on both paths" rather than just filling dynamic dims is clearly the stronger contract, and `DistributedMemRef.__post_init__` (alongside the dtype / partition-presence checks) feels like the natural home for it. Let me share how I'm currently thinking about it and where I'm unsure, and I'd like to hear your take.

The part I keep turning over is how this interacts with the `AffineSet` scope cut from your earlier review. The way I read it, the concrete path works for `AffineSet` partitions today precisely because it early-exits and never reads them — they're stored and dispatched at access time via `find_partition` / `.contains()`, never needing a per-axis `max(hi)`. Once the function runs on every construction, an axis backed by an `AffineSet` partition still has no per-axis upper bound to validate `declared >= max(hi)` against (`max(hi)` being well-defined only for axis-aligned boxes). So my current thinking is that those axes can't really be checked — they'd stay trust-the-author as today, with only `BoxSet`-backed axes getting the new check — which keeps concrete `AffineSet` views working and leaves general `AffineSet` extent derivation under #74. But I'm not sure that's how you're picturing it, so I'd rather check than assume.

Two more things I'm weighing and would value your view on:

- **Scope** — moving this into `__post_init__` means it validates *every* `DistributedMemRef` construction, not just the dynamic-result op, which is a bit broader than #111's original "open the parser door". I'm unsure whether it's cleaner to do it all here or keep #111 to the dynamic-shape fix and carry the concrete-path validation in a follow-up. How do you see the sizing?
- **Concrete check direction** — my instinct is `declared >= max(hi)` (reject only undersize; an oversized declared dim passing through as uncovered tail, consistent with the coverage-is-access-time stance already in the PR), but I could see an argument for strict equality given "partitions are ground truth". Curious which you'd prefer.

Once I understand how you're seeing the `AffineSet` interaction in particular, I'm happy to pick it up.

fabianlim:
>Two more things I'm weighing and would value your view on:

@WarningRan yes I know its deviating from scope of PR, but Im proposing this as a way to validate your `max(hi)` intuition, which I feel its not very correct. If you do what I proposed, it will be a way for you to validate this intuition.

WarningRan:
@fabianlim Worked through the exercise — short version: it doesn't surface an over-estimate, and tracing why points the other way (`max−min` would under-size). Details:

1. The exercise result — and why it isn't conclusive on its own. I added the `declared >= max(hi)` check and ran the suite: every distributed-view fixture passes with `declared == max(hi)`, none under-declare. I don't want to oversell that, though — all our fixtures are origin-anchored (min(lo)=0), where max(hi) == max−min, so they can't actually distinguish the two. The exercise rules out the easy case but doesn't settle it by itself.

2. Where it's actually decided: non-origin tilings. `max(hi)` and `max−min` only diverge when `min(lo) > 0`. Take partitions [10,20) and [20,30) on an axis (`max(hi)=30`, `max−min=20`). There the global shape is forced to `30` by existing code, not by my check:

the access origin is global — [global_base = base_map.eval(indices)](https://github.com/torch-spyre/ktir-cpu/blob/51e54f6fe407a35e8def2ff3443f436b16130949/ktir_cpu/ops/memory_ops.py#L722) — so the view's index space has to reach global coord 29;
the slow path enumerates each partition against the global shape — [B_i.enumerate(dist_ref.shape)](https://github.com/torch-spyre/ktir-cpu/blob/51e54f6fe407a35e8def2ff3443f436b16130949/ktir_cpu/ops/memory_ops.py#L757) — which [raises when hi > shape](https://github.com/torch-spyre/ktir-cpu/blob/51e54f6fe407a35e8def2ff3443f436b16130949/ktir_cpu/affine.py#L318-L321). A shape of `20` can't hold a partition whose `hi=30`.
So `max−min` doesn't just under-cover — it breaks enumeration. `max(hi)` is the smallest shape that works.

3. On consistency with the static case. The check you proposed (`declared >= max(hi)`) is itself the statement that `max(hi)` is the minimum a static author can legally declare. The dynamic path just fills that same minimum — so the two are consistent by construction, not in tension.

4. If the concern is rebasing — i.e. the `[0,10)` prefix looks wasted — that rebase already happens, at the partition-local layer: [p_i = min(B_i)](https://github.com/torch-spyre/ktir-cpu/blob/51e54f6fe407a35e8def2ff3443f436b16130949/ktir_cpu/ops/memory_ops.py#L753), and `C_i - p_i` maps global coords into each partition's local buffer. The global view stays in global coords by design (RFC 0682: "global view of all memory", "global coordinate domain is the union of the `coordinate_set` attributes"). Rebasing the global shape would desync it from the global coords `base_map` produces.

Net: `max(hi)` is the floor forced by global addressing + enumeration, not an over-estimate; `max−min` under-sizes on non-origin tilings. The non-origin case is pinned by `test_construct_distributed_memory_view_resolves_non_origin_to_max_hi` in this PR.

Does that line up with how you're seeing it — or is there a case where you'd expect `max(hi)` to genuinely over-shoot?

fabianlim:
@Prasanth-Chatarasi can I check something with you, in the below mlir for distributed memory view, 

```mlir
#P0_coord = affine_set<(d0, d1) : (d0 >= 10, -d0 + 19 >= 0, d1 >= 0, -d1 + 3 >= 0)>
#P1_coord = affine_set<(d0, d1) : (d0 >= 20, -d0 + 29 >= 0, d1 >= 0, -d1 + 3 >= 0)>
#ac_set   = affine_set<(d0, d1) : (d0 >= 0,  -d0 + 19 >= 0, d1 >= 0, -d1 + 3 >= 0)>  // 20x4
#ident    = affine_map<(d0, d1) -> (d0, d1)>

module {
  func.func @f() {
    %c0  = arith.constant 0  : index
    %c10 = arith.constant 10 : index
    %p0a = arith.constant 0    : index   // P0 byte/elem base
    %p1a = arith.constant 4096 : index   // P1 elem base
    %ba  = arith.constant 8192 : index   // B  elem base

    %P0 = ktdp.construct_memory_view %p0a, sizes: [10, 4], strides: [4, 1] {
        coordinate_set = #P0_coord, memory_space = #ktdp.spyre_memory_space<HBM>
    } : memref<10x4xf16>
    %P1 = ktdp.construct_memory_view %p1a, sizes: [10, 4], strides: [4, 1] {
        coordinate_set = #P1_coord, memory_space = #ktdp.spyre_memory_space<HBM>
    } : memref<10x4xf16>

    // <<< the global shape under question is here: memref<30x4xf16> >>>
    %A = ktdp.construct_distributed_memory_view
        (%P0, %P1 : memref<10x4xf16>, memref<10x4xf16>)
        : memref<30x4xf16>

   ...
```

My understanding is that the above MLIR is **wrong**. `%A` should be a memref of type `memref<20x4xf16>`, can you confirm what the correct mlir should be?

Prasanth-Chatarasi:
> @Prasanth-Chatarasi can I check something with you, in the below mlir for distributed memory view,
> 
> ```mlir
> #P0_coord = affine_set<(d0, d1) : (d0 >= 10, -d0 + 19 >= 0, d1 >= 0, -d1 + 3 >= 0)>
> #P1_coord = affine_set<(d0, d1) : (d0 >= 20, -d0 + 29 >= 0, d1 >= 0, -d1 + 3 >= 0)>
> #ac_set   = affine_set<(d0, d1) : (d0 >= 0,  -d0 + 19 >= 0, d1 >= 0, -d1 + 3 >= 0)>  // 20x4
> #ident    = affine_map<(d0, d1) -> (d0, d1)>
> 
> module {
>   func.func @f() {
>     %c0  = arith.constant 0  : index
>     %c10 = arith.constant 10 : index
>     %p0a = arith.constant 0    : index   // P0 byte/elem base
>     %p1a = arith.constant 4096 : index   // P1 elem base
>     %ba  = arith.constant 8192 : index   // B  elem base
> 
>     %P0 = ktdp.construct_memory_view %p0a, sizes: [10, 4], strides: [4, 1] {
>         coordinate_set = #P0_coord, memory_space = #ktdp.spyre_memory_space<HBM>
>     } : memref<10x4xf16>
>     %P1 = ktdp.construct_memory_view %p1a, sizes: [10, 4], strides: [4, 1] {
>         coordinate_set = #P1_coord, memory_space = #ktdp.spyre_memory_space<HBM>
>     } : memref<10x4xf16>
> 
>     // <<< the global shape under question is here: memref<30x4xf16> >>>
>     %A = ktdp.construct_distributed_memory_view
>         (%P0, %P1 : memref<10x4xf16>, memref<10x4xf16>)
>         : memref<30x4xf16>
> 
>    ...
> ```
> 
> My understanding is that the above MLIR is **wrong**. `%A` should be a memref of type `memref<20x4xf16>`, can you confirm what the correct mlir should be?

Looking at the P0_coord and P1_coord, it should be <30x4xfp16>, however, accessing anything from 0 to 9 of the first dimension has undefined semantics since those coordinates are not present in any of the input memory views passed to distributed memory view.

WarningRan:
@fabianlim 
Pushed the recover-shape change. Global shape is now the data span `max(hi) - min(lo)` — your `[10,30)` example comes out `20`, not `30`. Also unified the concrete and `?` paths so both check against the partitions (the concrete path used to trust the declared shape blindly), and added a test pinning the `20`.

fabianlim:
> Looking at the P0_coord and P1_coord, it should be <30x4xfp16>, however, accessing anything from 0 to 9 of the first dimension has undefined semantics since those coordinates are not present in any of the input memory views passed to distributed memory view.

@Prasanth-Chatarasi yes correct so I feel it we take it as `<30x4xfp16>,`, it introduces additional complexities since then in the `create_access_tile`, the `access_tile_set` will have its affine set range from 0 to <30,  with the first 10 row positions invalid.  Wouldnt it be more consistent if we enforced the memref from the construct_distributed view to be `<20x4xfp16>,`, then we simply consider `access_tile_set` going from 0 to < 20?


It also reads more clearly since P0 and P1 are both memrefs with only 10 rows each

```mlir
%A = ktdp.construct_distributed_memory_view
        (%P0, %P1 : memref<10x4xf16>, memref<10x4xf16>)
        : memref<30x4xf16>
```

fabianlim:
## Review — dynamic result shape

We've asked Prasanth to clarify the intended semantics as a follow-up; this review is my read of the PR as it stands, and these are the comments if it's reviewed as-is.

The shape math is correct: for partitions `[16,32) + [32,48)` the view genuinely has **32** elements, so `max(hi) − min(lo) = 32` is the right logical size. `max(hi) = 48` was never a real size — it only satisfied an enumeration bound. Two issues to focus on; a third (`__post_init__` layering) noted for follow-up.

### 1. Slow-path crash is a consumer bug, not a shape limitation — `ops/memory_ops.py:757`

```python
B_i_pts = B_i.enumerate(dist_ref.shape)   # wrong scaffold
```

`dist_ref.shape` is the *global logical size*, but `enumerate`'s argument is a *coordinate upper bound* — and `B_i` already carries its own `lo`/`hi`. For a non-origin tiling (`min(lo) > 0`) the resolved shape is `max(hi) − min(lo) < max(hi)`, so `B_i.hi > shape` and `BoxSet.enumerate` raises. The partition should enumerate against **its own extent**; the existing `p_i = min(B_i)` origin and the `C_i − p_i` / `cs.lo[d] − x[d]` translations already place each point into access-local coords, and none of that downstream reads `dist_ref.shape`. Fix is local:

```python
B_i_pts = B_i.enumerate()          # BoxSet is self-bounded; no external box needed
```

With that, `[16,32)+[32,48)` enumerates 16 + 16 points and slots them into local coords correctly. This is a one-line consumer fix, not an inherent limitation of the resolution rule — please either fix it here or re-label #74 as "slow-path `enumerate` uses global shape instead of partition bounds" so the docstring stops framing it as a shape-math limitation.

### 2. `_assemble_dist_extent` — collapse to the two real cases, and close the AffineSet gaps

Per axis this is really just: **dynamic → assign inferred extent; concrete → check declared against inferred extent**, where inferred = `max(hi) − min(lo)` from the partitions. Today that's buried behind an interleaved per-partition pass whose `has_dynamic` / `isinstance(BoxSet)` / `is_concrete` / `n_dims` branches make the exception set hard to audit — and that entanglement produces two real gaps:

- **Rank check skipped for AffineSet partitions (concrete shape).** The `has_dynamic`-gated `continue` on the `not isinstance(cs, BoxSet)` branch runs *before* the `n_dims != ndim` check, so a wrong-rank AffineSet partition against a concrete result isn't caught at construction.
- **Concrete validation is toothless when any partition is AffineSet.** `[0,16)` BoxSet + `[16,48)` AffineSet, declared `(48,)`: the AffineSet contributes no bound, so inferred = 16 and declaring `(16,)` passes despite real coverage `[0,48)`. So "do we validate the declared memref against the partitions?" is today **only against the BoxSet subset**.

Suggested shape — split *inference* from *decision*:

```python
def _infer_partition_extent(partitions) -> tuple[Optional[int], ...]:
    """Per-axis max(hi)-min(lo). Validates each partition (BoxSet, concrete,
    matching rank) up front. None on an AffineSet-only axis."""
    ...

def _assemble_dist_extent(shape, partitions):
    inferred = _infer_partition_extent(partitions)
    resolved = list(shape)
    for axis, declared in enumerate(shape):
        if declared is None:            # dynamic: assign
            # require inferred[axis] is not None and > 0, else raise
            resolved[axis] = inferred[axis]
        else:                           # concrete: check
            # if inferred[axis] is not None and declared < inferred[axis]: raise
            ...
    return tuple(resolved)
```

Moving the rank check into `_infer_partition_extent` (ahead of the kind dispatch) fixes the first gap. The second can't be fully closed without a bounding box from the AffineSet — that's genuinely #74 territory — so the honest move is to make the BoxSet-only scope explicit rather than silently skip.

### 3. (Follow-up, not blocking) `__post_init__` mutates `self.shape`

`DistributedMemRef(shape=(None,64))` silently returns `.shape == (128,64)`. Since resolution depends on partition coordinate sets (an execution concern) and the handler forwards the raw shape, resolving in the handler and keeping `__post_init__` validate-only would read more honestly. Let's address this separately after the two above.

### Not a bug
Over-sized declared dims don't leak into loaded tiles: `ktdp.load` defaults `result_shape` to `access_tile.shape`, not `dist_ref.shape`.

yuhaohaoyu:
@fabianlim @WarningRan @Prasanth-Chatarasi Here is my take (took me a while with Claude scanning code bases). Here is our take: 

## for **max(hi)**

### 1. Non-contiguous partitions break the min(lo)-rebasing anyway

- If [10,20) + [20,30) is valid MLIR, then [10,20) + [45,55) probably equally valid?  
- Under max(hi) - min(lo): shape = 55 - 10 = 45; with only 20 elements exist out of 45 slots. The rebase to min(lw) will have holes.  The "no holes" case works for the very special case, not for general case.

### 2. Holes native to **indirection**, not **distributed-view**

- Distributed-view (`coordinate_set`) is a regular, static partitioning — affine constraints, known at compile time. 
- Indirection (`indirect_access_tile`) is where irregular, sparse, data-dependent access patterns live.
- If force the distribution layer to "compact away" holes by rebasing, we're mixing the two responsibilities. 

### 3. Starting from 0 is simple and universal

- MLIR memrefs are 0-indexed. C arrays are 0-indexed. Everyone knows what `shape = 30` means: valid indices are [0, 30).
- With max(hi) - min(lo), `shape = 20` actually means "indices [0, 20) which map to global coordinates [10, 30)." Now every consumer needs to know the rebase offset, indirection mess. 

### 4. This discussion may belong to whether to compose distribution and indirection — not here

- The MLIR frontend allows feeding a **distributed view** into an **indirect access tile** (both typed `AnyMemRef`). If we want to reason about rebasing and compaction, that composition point is where it matters.
- ktir-cpu **doesn't compose** them today. At the emulator level, distributed views are always accessed through regular access tiles with affine sets, where max(hi) is enough.
- Introducing min/max rebasing now is solving a problem for a code path that doesn't exist yet. When indirect-into-distributed gets wired up, the rebasing/min decision will naturally live in the indirect logic.

---

## The question

When resolving a dynamic (`?`) dimension in `construct_distributed_memory_view`, should the shape be:

- **max(hi):** the global bounding box from 0
- **max(hi) - min(lo):** the data span, rebased to start at 0

yuhaohaoyu:
@fabianlim Maybe there is some needs for this corner cases and it's urgent ? surfacing some kernel lowering cases ? Otherwise, I did not like the 'min/max' pairing is just it smells like 'indirection'. 

There are many parallels of distributed/partitioned (address) space to indirection: virtual address space and physical memory space through paging in OS; reservation vs allocation, etc. 

So maybe we want to honestly bring in the 'composing of distributed vs indirection' is the natural way.

WarningRan:
Pushed two changes that are orthogonal to the max(hi) vs max(hi)−min(lo) question, so they don't pre-empt that discussion:

1. Slow-path `enumerate` decoupled from the global shape. A `BoxSet` partition is self-bounded, so the slow path now enumerates each one against its own `[lo, hi)` instead of `dist_ref.shape` — a non-origin partition no longer raises when its `hi` exceeds the resolved shape. `AffineSet` partitions still need an external search box; a tight per-partition box for them stays under #74.

2. Extent validation split into inference + decision, with the rank check moved ahead of the partition-kind dispatch (a wrong-rank `AffineSet` against a concrete result is now rejected instead of silently skipped). The concrete-shape check remains BoxSet-only.

Deliberately left for the ongoing discussion:
- The resolved-shape formula (max(hi) vs max(hi)−min(lo)) is untouched.
- No access-time guard was added for coordinates not covered by any partition — that region keeps its current behavior. Whether it should stay silently undefined or fault explicitly is still open; happy to add the guard once we settle that.

WarningRan:
Pushed the shape change on top of the earlier cleanup.

`cfda16b` resolves the global shape to max(hi) — the global bounding box from 0 — per RFC 0682's "global coordinate domain is the union of the coordinate_set attributes". So a non-origin tiling keeps its uncovered [0, min(lo)) prefix as space with undefined access semantics, rather than shrinking the shape. (`8e4b097` before it is the orthogonal cleanup from the earlier review: slow-path enumerate decoupled from the global shape, and extent validation split into inference + decision.)

Not touched yet:
- No access-time guard for the uncovered prefix — accessing a coordinate no partition covers keeps current behavior (silent for a partially-covered region; "no partition covers" only when the whole access region is uncovered).
- __post_init__ relayering left as a follow-up.
- AffineSet extent stays under #74.

Open for the reviewers to settle:
1. max(hi) vs max(hi) − min(lo) for the resolved shape — I've gone with max(hi); please confirm.
2. Accessing the undefined prefix: stay silent, or fault explicitly (a guard in create_access_tile / load)?
3. Should the "bounding box may contain uncovered coordinates; access is undefined" contract be written into the spec/docstring?

fabianlim:
**Stale comments (3x, leftover from the earlier `max(hi) - min(lo)` draft — the implemented/confirmed formula is `max(hi)` alone, per the discussion above):**

- `ktir_cpu/ir_types.py:266-267` (comment above `self.shape = _assemble_dist_extent(...)` in `DistributedMemRef.__post_init__`)
- `ktir_cpu/dialects/ktdp_ops.py:107-109` (docstring of `ktdp__construct_distributed_memory_view`)
- `tests/test_dialects_exec.py:1782` (docstring of `test_construct_distributed_memory_view_concrete_rejects_undersize`)

None of these affect behavior — the actual computation only uses `max(hi)` — but they should be updated to avoid confusing future readers.

**Correctness gap — concrete-shape check is one-sided (`ir_types.py:224`)**

```python
elif extent is not None and dim < extent:
```

This only rejects a declared axis smaller than the inferred `max(hi)` extent; a declared axis *larger* than the inferred extent is silently accepted. `construct_distributed_memory_view`'s own memref annotation is independently comparable against `_infer_partition_extent`'s result, so this should be an equality check:

```python
elif extent is not None and dim != extent:
    raise ValueError(
        f"construct_distributed_memory_view: declared shape axis {axis} "
        f"= {dim} does not match the partition extent (max(hi)) "
        f"= {extent}; the view's declared shape must equal the global "
        f"coordinate extent covered by its partitions"
    )
```

This also means the `_assemble_dist_extent` docstring (~lines 176-183, "concrete: validated `declared >= inferred`... an over-sized one is allowed") needs updating, and a test covering the over-size case should be added alongside the existing under-size test.

WarningRan:
Thank you so much @fabianlim ! Done in the latest commit:

- Concrete check is now equality (`declared == max(hi)`) — over-size rejected like under-size, message as suggested (+ over-size test).
- The three stale `max(hi) - min(lo)` comments now read `max(hi)`.
- Relayered: resolution moved to the handler, `__post_init__` only validates — so the extent check runs on every construction, not just the op path.

Will resolve the conflicts with main right before we want to move forward.

fabianlim:
@WarningRan there are conflicts in the dialect files that need to be resolved.

WarningRan:
Let's move on. I'll fold items 1–3 into a follow-up PR; item 4 stays under #74. Good to merge.

PR #112, opened 2026-06-12
[Feature] Add tensor.extract_slice and tensor.insert_slice ops

fabianlim:
Adds executor handlers, regex text parsers, and MLIRTypeAdapter handlers for `tensor.extract_slice` and `tensor.insert_slice`, along with parse and exec tests for both ops.

Also fixes the `tensor.extract` dispatch pattern (adds trailing space) to prevent it from shadowing `tensor.extract_slice` in the parser registry.

### DIFF
--- a/ktir_cpu/dialects/tensor_ops.py
+++ b/ktir_cpu/dialects/tensor_ops.py
@@ -238,7 +238,7 @@ def tensor__extract_slice(op, context, env):
     strides = _resolve_dynamic(static_strides, dyn_st_ops,  context)
 
     idx = tuple(
-        slice(off, off + sz, st)
+        slice(off, off + sz * st, st)
         for off, sz, st in zip(offsets, sizes, strides)
     )
     sliced = src.data[idx]
@@ -282,7 +282,7 @@ def tensor__insert_slice(op, context, env):
     strides = _resolve_dynamic(static_strides, dyn_st_ops,  context)
 
     idx = tuple(
-        slice(off, off + sz, st)
+        slice(off, off + sz * st, st)
         for off, sz, st in zip(offsets, sizes, strides)
     )
     result_data = dst.data.copy()
--- a/ktir_cpu/dialects/tensor_ops.py
+++ b/ktir_cpu/dialects/tensor_ops.py
@@ -184,6 +184,65 @@ def tensor__yield(op, context, env):
     return ControlOps.yield_op(values)
 
 
+@register("tensor.extract_slice")
+def tensor__extract_slice(op, context, env):
+    """Slice a sub-tensor and reshape to the result type's shape.
+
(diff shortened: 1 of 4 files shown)


fabianlim:
CC @kiszk @lchu6 @nwang-ibm

ktir_cpu/dialects/tensor_ops.py
```diff
@@ -184,6 +184,112 @@ def tensor__yield(op, context, env):
...
+    dyn_st_ops  = dyn_ops[n_dyn_off + n_dyn_sz:n_dyn_off + n_dyn_sz + n_dyn_st]
+
+    offsets = _resolve_dynamic(static_offsets, dyn_off_ops, context)
+    sizes   = _resolve_dynamic(static_sizes,   dyn_sz_ops,  context)
+    strides = _resolve_dynamic(static_strides, dyn_st_ops,  context)
+
+    idx = tuple(
+        slice(off, off + sz, st)
+        for off, sz, st in zip(offsets, sizes, strides)
+    )
```
lchu6, line 243: if I read the intention correctly here, we want "size" number of elements here.  so the range should be off -> off + sz * st?   do we miss the `*st` here in the slice range?

lchu6: if true, then same for `insert_slice`

ktir_cpu/dialects/ktdp_helpers.py
```diff
@@ -121,26 +125,11 @@ def parse_subscript_expr(token: str, var_names: list):
...
-    # Legacy fast-path: keyword forms not in the expression grammar.
-    m = _re.match(r'^(\w+)\s+floordiv\s+(\d+)$', bare)
-    if m:
-        return ("floordiv", var_names.index(m.group(1)), int(m.group(2)))
-    m = _re.match(r'^(\w+)\s+mod\s+(\d+)$', bare)
-    if m:
-        return ("mod", var_names.index(m.group(1)), int(m.group(2)))
-
-    # General case: recursive-descent parse, then classify named refs.
     node = _Parser(_tokenise(t)).parse_expr()
```
lchu6, line 133: does this handle well on things like `2 * %i floordiv 4`?

fabianlim: @lchu6 sorry I should not have pushed this commit, it was a mistake, the affine set parsing should be handled in #115

ktir_cpu/dialects/tensor_ops.py
fabianlim: thanks for the good catch!

lchu6 (dismissed):

nwang-ibm (approved):

PR #114, opened 2026-06-13
Fix redundant matmul computation in linalg.batch_matmul handler

### LINKED ISSUE #113: `linalg.batch_matmul` handler computes matmul twice
## Summary

In `ktir_cpu/dialects/linalg_ops.py`, the `linalg__batch_matmul` handler computes `tile_a.data @ tile_b.data` twice on line 313:

```python
result = Tile(tile_a.data @ tile_b.data, tile_a.dtype, (tile_a.data @ tile_b.data).shape)
```

The matmul is evaluated once for the `data` argument and again solely to derive `.shape`. This is redundant and wasteful — batched matmul is O(B·M·K·N).

## Fix

Store the result of the matmul once and reuse it:

```python
product = tile_a.data @ tile_b.data
result = Tile(product, tile_a.dtype, product.shape)
```

nwang-ibm:
## Summary

Fixes https://github.com/torch-spyre/ktir-cpu/issues/113

The `linalg.batch_matmul` handler computed `tile_a.data @ tile_b.data` twice — once for the data and once solely to derive `.shape`. This is wasteful since batched matmul is O(B·M·K·N). The fix stores the result once and reuses it.

## Changes

- `ktir_cpu/dialects/linalg_ops.py`: compute the matmul product into a local variable and reuse it for both the `Tile` data and shape arguments.

### DIFF
--- a/ktir_cpu/dialects/linalg_ops.py
+++ b/ktir_cpu/dialects/linalg_ops.py
@@ -310,7 +310,8 @@ def linalg__batch_matmul(op, context, env):
     """
     tile_a = context.get_value(op.operands[0])  # ins[0] = A  (B×M×K)
     tile_b = context.get_value(op.operands[1])  # ins[1] = B  (B×K×N)
-    result = Tile(tile_a.data @ tile_b.data, tile_a.dtype, (tile_a.data @ tile_b.data).shape)
+    product = tile_a.data @ tile_b.data
+    result = Tile(product, tile_a.dtype, product.shape)
     if len(op.operands) > 2:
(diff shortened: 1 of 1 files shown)


fabianlim (approved):
Verified this is the only instance of the duplicate-computation pattern. Checked every `Tile()` constructor call across all dialect and ops files — no other call has `Tile(expr, dtype, expr.shape)` where `expr` is an expensive duplicated computation. All others either use a pre-computed local variable, the input tile's shape directly, or a separate shape variable. LGTM.

PR #115, opened 2026-06-15
feat: floordiv/mod as first-class operators in affine expression parser

fabianlim:
## Problem

`floordiv` and `mod` are MLIR affine grammar keywords but the tokeniser was emitting them as plain identifiers. The parser intercepted them ad-hoc in `_term` via a `peek()` check, but this had two structural bugs:

1. The integer-leading branch (`5 floordiv 3`) had an early `return` that exited `_term` before reaching the operator loop, so `5 floordiv 3` would silently misparse.
2. `_atom` had no guard against keywords reaching it — it raised an opaque `ValueError: Unexpected token: 'floordiv'` instead of something actionable.

This blocked the gather stick kernel, which needs compound-LHS expressions like `(y_offset + d*64 + lane) floordiv 64`.

## Changes

**`ktir_cpu/parser_ast.py`**
- Add `_AFFINE_KEYWORDS` (reserved keyword set) and `_MULTIPLICATIVE_OPS` dispatch table (`token → ast_tag + coeff_first` flag).
- Unify `_term`: integer-leading `N*expr` keeps its prefix form; all postfix operators (`expr*N`, `expr floordiv N`, `expr mod N`) share a single `while` loop keyed off `_MULTIPLICATIVE_OPS` — no `if/elif` chain.
- Guard `_atom`: raises `ValueError` mentioning "Keyword" before the opaque catch-all.
- Add `("floordiv", node, int)` and `("mod", node, int)` AST tags; evaluate via `_eval_node` using Python `//` and `%` (floor semantics, non-negative remainder).

**`ktir_cpu/dialects/ktdp_helpers.py`**
- Recurse into `floordiv`/`mod` LHS in `_classify_refs` and `_reclassify_dims`.
- Remove legacy regex fast-path from `parse_subscript_expr`; simplify `eval_subscript_expr` to fully delegate to `_eval_node`.

**`tests/`**
- `TestFloorDivMod` in `test_ast.py`: AST structure, evaluation, keyword guard.
- `TestSubscriptExpr` in `test_indirect_access.py`: `parse_subscript_expr`/`eval_subscript_expr` with compound SSA expressions.

## Adding a new keyword in future

Touch exactly one place: add an entry to `_MULTIPLICATIVE_OPS` (e.g. `"ceildiv": ("ceildiv", False)`).

### DIFF
--- a/ktir_cpu/parser_ast.py
+++ b/ktir_cpu/parser_ast.py
@@ -232,6 +232,10 @@ class _Parser:
                 raise ValueError(f"Expected integer constant after {op!r}, got {num_tok!r}")
             n = int(self.consume())
             tag, int_both_sides = _MULTIPLICATIVE_OPS[op]
+            if op in ("floordiv", "mod") and n <= 0:
+                raise ValueError(
+                    f"{op!r} requires a positive integer divisor, got {n}"
+                )
             primary = (tag, n, primary) if int_both_sides else (tag, primary, n)
         return primary
 
--- a/ktir_cpu/parser_ast.py
+++ b/ktir_cpu/parser_ast.py
@@ -43,7 +43,8 @@ Linear affine expressions:
   - Addition (+), subtraction (-), negation (unary -)
   - Multiplication by a constant coefficient (N * dI or dI * N)
 
-Not supported: floordiv, ceildiv, mod.
+Supported division ops: floordiv (floor toward -∞), mod (non-negative remainder).
+Not supported: ceildiv.
 
 Affine set enumeration
 ----------------------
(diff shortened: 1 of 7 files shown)


fabianlim:
@nwang-ibm @WarningRan please review when you get a chance

lchu6:
wouldn't something like `((%y_offset + %d_stick * 64 + %d_lane) floordiv 64)` goes through `strip('()')` and becomes `%y_offset + %d_stick * 64 + %d_lane) floordiv 64` then eventually got wrong results (e.g. floordiv 64 got ignored)?

ktir_cpu/parser_ast.py
```diff
@@ -196,30 +211,40 @@ def _term(self) -> _Node:
...
+        else:
+            primary = self._atom()
 
-        # Atom that may be followed by a coefficient: expr * N
-        node = self._atom()
-        if self.peek() == "*":
-            self.consume("*")
+        # Postfix multiplicative operators: expr * N, expr floordiv N, expr mod N.
+        # All require an integer constant on the right.
+        while self.peek() in _MULTIPLICATIVE_OPS:
```
lchu6, line 228: should we add a check to make sure we reject n <= 0 for these two new ops?

fabianlim: @lchu6 thanks for the good suggestion, I pushed something to address this

lchu6 (approved):

PR #118, opened 2026-06-16
fix: remove LX double-counting in scf.for iter_arg bindings

### LINKED ISSUE #109: [Bug] LX scratchpad double-counts `scf.for` iter_arg init values

fabianlim:
## Summary

- `scf.for` iter_arg bindings are pure SSA name aliases — no new buffer is allocated. `for_op` was calling `track_lx` on both the init value (already tracked by `_execute_operation`) and on every re-bind after yield, causing `lx.used` to double-count and grow unboundedly across iterations.
- Removed all `track_lx` / `untrack_lx` calls from `for_op`. LX ownership stays entirely with `_execute_operation`, which tracks every Tile-producing op result exactly once.
- Added 9 regression tests in `TestForOpLXTracking` covering: double-count at init, leak after loop exit, iter_arg visibility in body, multiple iter_args, zero iterations, scalar iter_args, per-iteration stability, new tiles yielded from body, and nested loops.

Fixes #109.


fabianlim:
CC @kiszk @lchu6 @nwang-ibm (issue author)

lchu6:
should we re-track only when the yielded Tile is not already tracked/live under some existing SSA name? i.e., the accounting needs alias awareness or explicit ownership transfer, not blanket “never track iter_args.”

lchu6:
the current timeline looks like:

```
before loop:
  %init tracked                     lx.used = N

iteration 0:
  body creates %new0, tracked        lx.used = 2N
  pop body scope frees %new0         lx.used = N
  PR rebinds %itr = %new0            lx.used = N   <-- should account for carried %new0

iteration 1:
  body creates %new1                 lx.used = 2N according to PR
  actual live tiles are %init, %new0, %new1 = 3N
```

lchu6:
the new test didn't capture this because it pushes and pops an extra manual scope inside the fake body. The real for_op() already owns the loop-body scope, so the test is not faithfully simulating _execute_operation() creating a body-local result.

fabianlim:
The fix ended up broader than the original approach. The root cause and design are documented on #109.

In short: replaced `_lx_bytes: Dict[str, int]` (keyed by SSA name) with `_tile_refcount: Dict[int, int]` (keyed by `id(tile)`). `set_value` now auto-tracks LX — first binding charges once, aliases increment the refcount without double-charging, `pop_scope` frees via refcount. `track_lx`/`untrack_lx` removed entirely. This fixes the double-count, the multi-result tracking gap, and eliminates all manual bookkeeping in `for_op`. 20 tests updated.

Also fixed the test as @lchu6 flagged — body lambda was adding an extra `push_scope`/`pop_scope`, and the assertion now correctly expects `2N` after the loop.

Would it make sense to retitle this PR to `refactor: replace name-keyed LX tracking with object-identity refcount`?

fabianlim:
One thing worth verifying on `_tile_refcount`: it's keyed by `id(Tile)`, and Python can reuse `id()` values after GC. 

I believe as long as the refcount is in sync, we do not bypass the API and always use `set_value` then it should not go out of sync. 

On the other hand, If we are worried that the ref_count is stale that that prevents GC and memory leak, then we can change this to weak ref like #124 

CC @kiszk @lchu6 @nwang-ibm

nwang-ibm:
LGTM. Maybe naming "_tile_refcount" to "_lx_refcount" or "_lx_tile_refcount" would be a bit clearer, but not critical.

nwang-ibm:
The refcout is verified to be correct. Current PR can be approved. However, the lx's overflow issue remains due to the interpreter's SSA model, which will be track in issue  [#130](https://github.com/torch-spyre/ktir-cpu/issues/130).

nwang-ibm (approved):

PR #120, opened 2026-06-16
feat: chip-level utilization / throughput metrics on roofline()

WarningRan:
## Summary

**The one new signal is `chip_throughput`** — what fraction of the *whole chip* a kernel uses. The other fields support or decompose it.

| | Before | After |
| --- | --- | --- |
| Per-active-core efficiency | `efficiency = achieved / ceiling` (critical core) | unchanged |
| **▶ Chip-wide throughput — _the point_** | **missing — done by hand** | **`chip_throughput = Σ(core FLOPs) / elapsed / (peak × num_cores)`** · Nsight *Compute (SM) Throughput %* |
| Grid coverage — _decomposes why low_ | missing | `cores_active`, `num_cores`, `grid_coverage = cores_active / num_cores` · spatial dispatch coverage (not Nsight *SM Active %*) |
| HW peak reference — _feeds throughput_ | missing | `peak_gflops`, `chip_peak_gflops` |

`chip_throughput` sums the **actual per-core FLOPs** over elapsed time (idle / lighter cores pull it down), so it is exact under uneven tiling — the same way Nsight aggregates per-SM counters across all SMs.

## What you'll see

Memory-bound `add_kernel` — shows why the three angles differ:

```text
  Dominant unit   : simd
  Grid coverage   : 32/32  (grid_coverage 100.0%)
  Efficiency      : 92.5%  (per-active core, achieved/ceiling)
  Chip throughput : 7.5%   (chip-wide, achieved/peak with idle cores)
```

Every core busy (100% coverage) and 92.5% against the BW ceiling, yet only **7.5%** of compute peak — the number the report couldn't give before.

## Why it matters

A kernel on `grid=[4,1]` at 90% per-active-core efficiency still uses only 4/32 cores ≈ 11% of the chip — invisible before. `chip_throughput` surfaces it; `grid_coverage` + `efficiency` split the cause (under-dispatch vs. per-core inefficiency).

## Also included

`summary_dict()` had an overloaded `num_cores` key returning the grid product; split into `grid_cores` (counter entries) and `num_cores` (hardware total), matching the top-level field.

## Out of scope

- `efficiency` rename / denominator → #122
- `dominant_unit` alignment (max-%-of-peak, chip-wide) and a time-based SM-Active-% field → follow-ups


WarningRan:
cc @fabianlim  @nwang-ibm

ktir_cpu/latency.py
```diff
@@ -535,11 +567,17 @@ def roofline(self) -> Dict[str, Any]:
             achieved = flops / elapsed_s if elapsed_s > 0 else 0.0
             # Roofline ceiling at this kernel's AI for this unit.
             ceiling = min(peak, peak_bw * ai)
+            chip_peak = peak * num_cores
```
fabianlim, line 588: is it possible to have diferent CPUs have different peaks? in that case should be add the peaks across cores?

lchu6:
maybe clean a little bit on the notebook output?
image

ktir_cpu/latency.py
WarningRan: `peak` here is the accelerator's per-core compute peak and `num_cores` its cores (not host CPUs). Spyre cores are homogeneous — one clock + one systolic/SIMD rate in `HardwareConfig` — so `peak * num_cores` is exactly the sum of identical per-core peaks. Heterogeneity only appears per functional unit (modeled separately as `unit_ceilings`) and per precision/generation (set via config), never core-to-core. Added a comment making that explicit.

fabianlim:
On notebook outputs: agree with @lchu6 — would suggest just clearing all outputs before merging rather than re-running. The current diff contains output drift (e.g. `352 → 369` cycles, `3.10 → 2.95` AI for `add_kernel`) that looks like the notebook was run on a different code version; cleared outputs sidestep that entirely and make the diff easier to read.

Minor `__str__` nit: the implementation prints `Cores active` but the notebook output in the diff shows `Cores used`. One of them should be updated to match.

WarningRan:
@fabianlim good call — cleared all outputs rather than re-running, so the diff is source-only now. That removes both the number drift (stale cycle/AI values) and the `Cores active` / `Cores used` mismatch in one go: the old label came from outputs generated before the rename, and on re-execute it renders `Cores active`, matching the code. Landing in the next push.

ktir_cpu/latency.py
```diff
@@ -535,11 +585,23 @@ def roofline(self) -> Dict[str, Any]:
             achieved = flops / elapsed_s if elapsed_s > 0 else 0.0
             # Roofline ceiling at this kernel's AI for this unit.
             ceiling = min(peak, peak_bw * ai)
+            # Cores are homogeneous: every core shares the same clock and
+            # compute rates from HardwareConfig, so the chip peak is
+            # peak * num_cores, which equals summing identical per-core peaks.
+            # Heterogeneity lives on other axes (per functional unit, captured
+            # separately in unit_ceilings; per precision/generation, captured by
+            # the config values), never core-to-core.
+            chip_peak = peak * num_cores
```
fabianlim, line 588: **Upper-bound concern for uneven splits**

Tracing through the computation: `achieved` is `critical.flops / elapsed_s` where `elapsed_s` comes from the *slowest* core's cycle count. `chip_throughput` then extrapolates by multiplying `achieved * cores_active`, implicitly assuming every active core ran at the critical core's rate.

With uneven tiling (e.g. tiles split 100/100/50 elements across 3 cores), the lighter cores did fewer FLOPs in the same wall time — but they're counted as if they matched the heaviest core. So `chip_throughput` overstates actual chip utilization and is an upper bound.

The alternative would be `sum(core.flops for all active cores) / elapsed_s / chip_peak` — uses actual total chip FLOPs rather than extrapolating from the critical core. That would give a tighter, accurate number at the cost of summing across all counters instead of just `critical`.

Worth deciding whether `chip_throughput` should be: (a) the current upper bound (critical-core extrapolation, consistent with how `efficiency` already works), or (b) an accurate chip-wide aggregate (requires all-core flop sum). The docstring already notes the assumption — just flagging that for uneven splits the gap between the two can be significant.

WarningRan: Agreed it's an upper bound - went with (b), the chip-wide aggregate, since that's exactly what Nsight's SM Throughput does: it aggregates per-SM counters across all SMs over elapsed cycles (the `.sum`/`.avg` rollups of `sm__throughput...pct_of_peak_sustained_elapsed`), so idle/lighter SMs pull it down rather than being extrapolated up. `chip_throughput` now sums real per-core FLOPs over elapsed instead of `achieved * cores_active`. On a 300/200/100 uneven split the old form read 1.5× high; the two still coincide for even tiling (kept as a regression anchor).

Related but separate: the `dominant_unit` pick is still (a) chosen by raw FLOP volume rather than Nsight's max-%-of-peak (bottleneck pipeline), and (b) sampled from the critical core rather than aggregated chip-wide. It doesn't bite today — all cores run the same work mix, so the pick is stable — so I'd keep this PR scoped to chip_throughput and track the dominant_unit alignment as a follow-up, unless you'd rather fold it in.

ktir_cpu/latency.py
nwang-ibm: This is for open questions (3). maybe rename "num_cores" to "grid cores", keep num_cores as number of hardware cores and add "num_cores" with the hardware value for consistency.
```
def summary_dict(self) -> Dict[str, Any]:
    """Return summary as a dictionary."""
    return {
        "kernel_cycles": self.kernel_cycles,
        "kernel_time_us": self.kernel_time_us,
        "bottleneck": self.bottleneck,
        "grid_cores": len(self.counters),
        "num_cores": self.config.num_cores,
        "per_core": self.per_core_summary(),
    }
```

ktir_cpu/latency.py
fabianlim: >  dominant_unit alignment as a follow-up

@WarningRan yes sure might be a good followup, unless  @nwang-ibm thinks its not necessary

fabianlim:
Summarizing what this PR adds — please confirm my reading is right:

| | Before | After |
|---|---|---|
| Per-active-core efficiency | `efficiency = achieved / ceiling` for the critical core | Unchanged |
| Chip-wide throughput | Missing — had to compute by hand | `chip_throughput = achieved × cores_active / (peak × num_cores)` |
| Grid coverage | Missing | `cores_active`, `num_cores`, `compute_utilization = cores_active / num_cores` |
| HW peak reference | Missing | `peak_gflops`, `chip_peak_gflops` (plumbing for `chip_throughput`) |

So the core new signal is `chip_throughput`. The grid-coverage fields (`cores_active`, `compute_utilization`) are there to let you decompose *why* chip throughput is low — underdispatching vs. per-core inefficiency.

A few questions on `compute_utilization`:
- It is binary: a core counts as active if `total_cycles > 0`, even a single comm cycle. A full-grid kernel where all cores are barely active reads 100%.
- That makes the decomposition misleading for the per-core-inefficiency case — `compute_utilization` looks fine while `chip_throughput` is low, and the reader has to cross-reference `efficiency` to find the real cause.
- Is the intent purely "was this core dispatched at all"? If so, `grid_coverage` or `cores_dispatched_fraction` would be a more honest name than `compute_utilization`.

fabianlim (dismissed):
discussed with the author and feel its reasonable.

fabianlim (approved):
thanks for the fixes

Issue #95, opened 2026-06-09
[Feature] Strided sub-rectangle fast path in MemoryOps.load/store


fabianlim:
## Background

Split out from #52. PR #79 unified the direct and distributed access paths so that `ktdp.load` / `ktdp.store` route axis-aligned `BoxSet` accesses through `MemoryOps._subtile_ref` → `MemoryOps.load(sub_ref)` — the same composition `distributed_load` / `distributed_store` already use. The *full-rectangle* case lands on the existing contiguous fast path (`MemoryOps.load`, `_is_contiguous == True`) on both old and new code, so PR #79 is runtime-neutral there (benched at 0%, by design).

## Problem

The **non-full BoxSet** case (e.g. accessing `[0,4) × [0,8)` inside an `8×16` tile) is still slow. `_subtile_ref` produces a sub-tile with the box extent as `shape` but the parent's `strides` inherited verbatim:

- `shape=(4,8)`, `strides=[16,1]` → `_is_contiguous((4,8),[16,1])` is `False` (dim-0 stride 16 ≠ expected 8).
- Falls through to per-coord `_flat_memory_offsets` + fancy-index gather/scatter.

This is wasted work: a strided-but-rectangular region is exactly what NumPy `as_strided` + `.copy()` (or basic slicing on a reshaped view) does in one operation.

## Scope

Add a branch in `MemoryOps.load` / `MemoryOps.store` (after the `_is_contiguous` check, `ktir_cpu/ops/memory_ops.py:492` / `:566`) that detects a **strided rectangular** TileRef (no `coords`, strides describe a regular sub-rectangle of a parent allocation) and reads/writes it via `np.lib.stride_tricks.as_strided(...).copy()` instead of per-coord offsets.

Because this is **callee-side**, it lights up every caller at once — direct `ktdp.load/store` (via #79's `_subtile_ref` routing) and `distributed_load/store` both benefit with no further caller changes.

## Out of scope

- Caller-side dispatch (already landed in #79).
- Non-rectangular / coord-list accesses (preserve the existing slow path).

## Test plan

- Parity: strided sub-rectangle load/store returns identical result to the per-coord slow path.
- Stick-count accounting matches the slow path (latency model must not regress).
- No-trample on column-packed strides (mirror `test_distributed_store_col_packed_does_not_trample`).
- Spy `_flat_memory_offsets` is **not** called on the strided-rectangular case.

## References

[...]

PR #124, opened 2026-06-16
perf(interpreter): faster ktdp.load/store (vectorize offsets, ravel-not-flatten, O(log n) alloc lookup)

### LINKED ISSUE #1: [RFC conformance] Implement ktdp.construct_distributed_memory_view

starpit:
## Python reference-interpreter perf fixes

Three gratuitous-overhead fixes in `ktir_cpu` (the Python reference interpreter), all **correctness-preserving** — the load/store offsets are bit-identical (verified over 3,000 random shape/stride/coord cases) and the full `pytest` suite passes.

### What

1. **`ops/memory_ops._flat_memory_offsets`** — the `ktdp.load`/`store` address calculator linearized every loaded element's N-d coordinate to a flat offset with a pure-Python loop (`sum(c*s for c,s in zip(coord, strides))` per element, plus a `set` insert per element for HBM stick counting). Vectorized with numpy (`offsets = coords · strides`; distinct sticks via `np.unique`); returns the `int64` ndarray so the hot load/store paths fancy-index it directly.

2. **`memory._read_flat` / `_write_flat` / store path** — read the containing allocation via **`ravel`** (a view) instead of **`flatten`** (a full copy of the *entire* allocation — a whole weight — just to slice out one tile). The result is only sliced + `astype(copy=True)`, never mutated, so the copy was pure waste. On large allocations this is big: after fix #1, `flatten` was **57% of a llama-decode pass** (it is minor on smollm2, whose allocations are small). Also drops a redundant double-copy on the write path.

3. **`memory._find_allocation`** — the byte-address → allocation lookup scanned every live allocation linearly, **O(allocations) per access → O(n²)** for models whose function bodies hold thousands of live allocations (llama). Replaced with an **O(log n)** bisect over sorted base pointers (per-dict cache; allocations are non-overlapping and keys are only added or all-cleared, so a length change is a sound rebuild signal). Negligible for smollm2 (few live allocations); targeted at llama.

### Impact (Python per-node reference, ms per whole-model pass)

| config | before | after | speedup |
|---|---:|---:|---:|
| smollm2-135m decode (m=1)   |    36,757.6 |  2,536.3 | **14.5×** |
| smollm2-135m prefill (m=32) | 1,182,455.8 | 60,397.4 | **19.6×** |
| llama-3.2-1b decode (m=1)   |   395,961.2 | 31,777.6 | **12.5×** |

Where the wins come from (stage-by-stage, same branch/bench/inputs, only these fixes toggled):

- **offset vectorization** is the dominant fix for **smollm2** — it was ~94% of a smollm2 pass; with the bisect it takes smollm2 decode 36,757 → 3,004 (12.2×) and llama decode 395,961 → 78,810 (5.0×);
- **`ravel`-not-`flatten`** then adds only ~1.2× on smollm2 but a further **2.5×** on llama decode (and 2.8× on llama prefill) — large weight allocations make the per-read whole-allocation copy expensive;
- the **O(log n) `_find_allocation`** bisect is what makes the llama numbers possible at all (the old O(n²) scan).

(llama-3.2-1b prefill needs a raised LX cap to run — a separate change — so it has no pre-fix baseline here; the `ravel` fix alone cut it ~2.8×, from 2,778,093 → 977,688 ms.)

Measured on Apple M5, f16, per-node interpreter; one warm-up pass excluded. The measurement harness (a hermetic real-model bench) is a separate change, so the before/after isn't reproducible from this PR in isolation — but every row is the same branch/bench/inputs with only these three fixes toggled, so the deltas are attributable to them.


starpit:
cc @fabianlim @kiszk @lchu6 @nwang-ibm — standalone Python reference-interpreter perf fixes (vectorized `ktdp.load` offset calc + O(log n) allocation lookup), split out from the larger Rust-port work. Correctness-preserving (bit-identical offsets, full pytest passes); 5–14× on whole-model Python passes.

fabianlim:
The vectorization of `_flat_memory_offsets` is great. One coverage gap: the `distributed_load` / `distributed_store` **slow paths** (non-BoxSet `C_i`) still iterate element-by-element after calling `_flat_memory_offsets`, while the non-distributed `store` slow path already uses numpy fancy-indexing with the returned ndarray. Summary:

| Path | After this PR |
|---|---|
| `load` fast path | O(1) arithmetic — fine |
| `load` slow path | `flat[offsets]` — vectorized ✓ |
| `store` fast path | O(1) arithmetic — fine |
| `store` slow path | `flat[offsets] = tile.data.ravel()` — vectorized ✓ |
| `distributed_load` BoxSet | delegates → fast path — fine |
| `distributed_load` slow path | `for ac, off in zip(...)` — **still Python loop** |
| `distributed_store` BoxSet | delegates → `store` fast path — fine |
| `distributed_store` slow path | `for ac, off in zip(...)` — **still Python loop** |

The fix would be numpy fancy-indexing the `access_coords` → `out` / `flat` assignments in those two slow paths, symmetric with what `store` already does.

Would you be able to include that here, or should we open a follow-up issue to track it?

CC @kiszk @lchu6 @nwang-ibm

starpit:
thanks for the review! i can take a look today.

ktir_cpu/memory.py
```diff
@@ -94,6 +95,13 @@
 from .dtypes import to_np_dtype, bytes_per_elem
 
 
+# Per-dict cache for `_find_allocation`'s O(log n) bisect: maps a memory dict's
+# id() to (dict_ref, len, sorted_base_ptrs). The dict_ref pins the dict so its
+# id() can't be reused while cached; a len change (only adds, or an all-clear)
+# invalidates.
+_FIND_ALLOC_CACHE: Dict[int, Tuple[dict, int, list]] = {}
```
kiszk: Great perf impovements.

Since there is no chance to evict entries in `_FIND_ALLOC_CACHE`, memory consumption is always growing.
For example,
- LXScratchpad.clear() and lx.clear() do not evict entries
- multiple benchmark executions always add entries
- multiple test executions in pytest always add entires

Can we evict entries in the above cases?

starpit:
Good catch — addressed in the latest push (commit amended, still a single commit on the PR).

The `distributed_load` and `distributed_store` non-BoxSet slow paths now use NumPy fancy-indexing instead of the per-coordinate Python loop, symmetric with the non-distributed store slow path:

```python
# per-dimension index arrays built once from access_coords
out_idx = tuple(
    np.fromiter((c[d] for c in access_coords), dtype=np.intp, count=len(access_coords))
    for d in range(ndim)
)
out[out_idx] = flat[offsets]              # distributed_load
flat[offsets] = tile.data[src_idx]        # distributed_store
```

Order is preserved: `access_coords`, `local_coords`, and `offsets` all derive from `C_i` in the same order, so `offsets[i]` still pairs with `access_coords[i]`. Empty `C_i` degrades cleanly (empty index arrays).

Updated coverage table:

| Path | After this push |
| --- | --- |
| load slow path | `flat[offsets]` — vectorized ✓ |
| store slow path | `flat[offsets] = tile.data.ravel()` — vectorized ✓ |
| distributed_load slow path | `out[out_idx] = flat[offsets]` — vectorized ✓ |
| distributed_store slow path | `flat[offsets] = tile.data[src_idx]` — vectorized ✓ |

Added two cross-boundary slow-path tests (`test_distributed_load_slow_path_cross_boundary`, `test_distributed_store_slow_path_cross_boundary`) that build partitions via `parse_affine_set_raw` so the survivors carry a list `C_i` (not a BoxSet), asserting the vectorized slow path is genuinely exercised and not silently bypassed via the fast path. Full suite green.

starpit:
Good catch @kiszk — fixed in the latest push (commit amended, still one commit).

The root leak was that `_FIND_ALLOC_CACHE` was a module-global keyed by `id(memory)` whose value held a **strong** reference to the memory dict. That pinned every store ever created, so it could never be collected and `id()` reuse was masked — exactly the unbounded growth you describe across repeated benchmark/test runs.

Two changes:

1. **Weak, auto-evicting cache.** Memory now lives in a small `_AllocStore(dict)` subclass that is weak-referenceable and identity-hashed (plain `dict` is neither). The cache is a `weakref.WeakKeyDictionary` keyed by the store itself, with the value reduced to `(len, sorted_keys)` — no strong back-reference. When a store is garbage-collected (end of a benchmark run, end of a pytest test), its entry is dropped automatically.

2. **Eager eviction on clear.** `LXScratchpad.clear()` now also does `_FIND_ALLOC_CACHE.pop(self.memory, None)`. The store object survives `clear()` (same identity), so the weak dict would not drop it on its own; the `len`-change check would rebuild lazily, but we evict eagerly so a cleared scratchpad holds no stale sorted-key list.

Plain dicts passed directly to `_find_allocation`/`_read_flat` (e.g. the docstring examples) are detected via `isinstance(memory, _AllocStore)` and served without ever touching the cache.

Added `tests/test_alloc_cache.py` covering: `clear()` evicts, GC of a store auto-evicts (cache shrinks after `del` + `gc.collect()`), correctness after a rebuild on a new allocation, the plain-dict bypass, and the identity-keying of `_AllocStore`. Full suite green (931 passed).

fabianlim:
Thanks for the quick fix — distributed slow paths are now fully vectorized and the updated table looks complete. LGTM on that front.

Also noting: the strided sub-rectangle case (`coords is None`, non-contiguous strides) still goes through `_flat_memory_offsets` — that's tracked separately in #95 and out of scope here.

CC @kiszk @lchu6 @nwang-ibm

kiszk:
Thank you. LGTM on memory allocation side.

cc: @fabianlim @lchu6 @nwang-ibm

kiszk (approved):

Issue #122, opened 2026-06-16
[Bug] roofline() reports the wrong dominant compute unit for mixed systolic/SIMD kernels


WarningRan:
## Background

Reported by @nwang-ibm: the roofline `efficiency` does not distinguish whether work ran on the SIMD or the systolic unit. @fabianlim is aware. After investigating how profiling tools (e.g. Nsight Compute) model this, the scope widened from the original efficiency report to dominant-unit selection and per-unit arithmetic intensity.

## Phenomenon

`roofline()` picks a kernel's dominant compute unit by comparing accumulated FLOP counts across units. Systolic matmul FLOPs (`2·M·N·K`) and SIMD element-wise FLOPs (one per element) are not commensurable, so the comparison is not meaningful.

Concretely, `paged_attention` on the default `HardwareConfig()` reports `dominant_unit = systolic`, `efficiency ≈ 78%` — but on the critical-path core:

| unit | cycles spent | FLOPs |
| --- | ---: | ---: |
| systolic (matmul) | 1 | 524288 |
| SIMD (float + transcendental + int) | 494 | 28352 |

The systolic array runs 1 of ~2789 cycles, yet the kernel is labeled systolic and the headline `efficiency` / `arithmetic_intensity` are taken from that near-idle unit. Whether the mislabel surfaces depends on the configured peak ratio, so some `HardwareConfig` values hide it — a clean-looking report does not mean the logic is right.

## What we need

- Choose the dominant unit by **cycles actually spent** (`cycles_by_category`), not FLOP count.
- Give each compute unit its **own arithmetic intensity** (`unit_flops / total_bytes`, shared byte denominator) — as profiling tools do (Nsight Compute draws a separate roofline per pipeline) — with its own ceiling/efficiency; the top-level values follow the dominant unit.
- Update the notebook roofline plot to read the per-unit AI.
- Add a regression test: a mixed matmul + element-wise kernel must be classified by cycles, not FLOPs.
- Define each per-unit metric independently (`dominant_unit`, `arithmetic_intensity`, `achieved`, `ceiling`, `ridge`); `efficiency` keeps its existing `achieved / ceiling` formula and only receives the corrected per-unit inputs — the ceiling-vs-peak denominator question (#120) stays out of scope.

## Scope

**In:** dominant-unit selection + per-unit arithmetic intensity in `roofline()` (plus notebook plot + test).

[...]

PR #129, opened 2026-06-17
Fix: roofline dominant unit by cycles + per-unit arithmetic intensity

WarningRan:
Fix `roofline()`'s summary stage, which picked the dominant compute unit by accumulated FLOPs and fed both units one mixed arithmetic intensity. Step 1 of #122 — correctness only; the `efficiency` formula/denominator is unchanged.

| | before | after |
|---|---|---|
| dominant unit | most **FLOPs** | most **compute cycles** |
| arithmetic intensity | one mixed AI shared by both units | **per-unit** (`unit_flops / total_bytes`) |

- **Before**: dominant = `argmax(flops_by_category)`; ceiling = `min(peak, peak_bw × mixed_AI)`, shared by both units.
- **Why not enough**: matmul FLOPs (`2·M·N·K`) and SIMD per-element FLOPs aren't comparable, so a FLOP-heavy but near-idle unit was reported as dominant, and each unit's ceiling was inflated by the other unit's FLOPs.
- **What changed**: dominant is read from the per-category compute cycles already attributed upstream; each unit gets its own AI and ceiling; top-level AI/efficiency are sourced from the dominant unit.
- **How it solves it**: cycles identify the real bottleneck; per-unit AI gives each unit a correct ceiling. `efficiency` (`achieved/ceiling`) is untouched — it just receives the correct per-unit ceiling, so the reported value moves because the old one was computed from the wrong (mixed) ceiling.

Measured on default `HardwareConfig()`: dominant `systolic → simd`, top-level AI `7.71 → 0.40`, efficiency `0.78 → 0.82`. Adds a regression test pinning cycle-based dominant selection (previously unguarded).

Related to #120.

### DIFF
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -461,8 +461,8 @@ class LatencyReport:
         """Compute per-unit roofline metrics for the critical-path core.
 
         Two compute units are modelled (systolic for matmul, SIMD for everything
-        else).  The dominant unit — whichever accumulated the most FLOPs — sets
-        the compute ceiling.  The chart shows one roof::
+        else).  The dominant unit — whichever consumed the most compute cycles —
+        sets the headline.  The chart shows one roof::
 
             GFLOP/s
               ^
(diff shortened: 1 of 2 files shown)


WarningRan:
cc @nwang-ibm @fabianlim

fabianlim (approved):
Correctness fix is sound — cycles are the right proxy for dominant unit, and per-unit AI (NCU convention) gives each unit the correct ceiling. Test pins the exact pathological case. LGTM.

PR #132, opened 2026-06-19
Parse bundled multi-result LHS (%acc:2 = scf.for) in regex parser

### LINKED ISSUE #91: regex KTIRParser cannot resolve multi-result op references (%acc#0 / %acc#1)
The regex `KTIRParser` collapses multi-result ops (e.g. `scf.for` with `iter_args`) to a single name in its value map, so downstream uses of `%acc#0` / `%acc#1` KeyError at execution time. `MLIRFrontendParser` handles this correctly.

Repro: see below 

```mlir
%acc:2 = scf.for %i = %c0 to %c4 step %c1 iter_args(%a = %init0, %b = %init1) -> (f32, f32) : i32 { ... }
%sum = arith.addf %acc#0, %acc#1 : f32
```

The fix is to register both `%acc#0` and `%acc#1` in the value map when parsing a bundled multi-result LHS — `parse_multi_result_lhs` already produces the right names, but the interpreter's context binding needs to store them individually.

Surfaced by #83.

cc @kcirred

yuhaohaoyu:
## Issues Addressing

Fixes #91 - bundled multi-result LHS (`%acc:2 = scf.for ...`) causes KeyError at runtime because the regex parser does not recognize the `:N` suffix on result names.

## Design

The MLIR grammar defines result LHS as:

```
op-result-list ::= op-result (`,` op-result)* `=`
op-result      ::= value-id (`:` integer-literal)?
```

The parser needs to consume any conforming LHS and produce a flat list of individual SSA names mapping to the result (LHS for left hand sides of an expression) values. A single utility (`parse_multi_result_lhs`) implements this production rule; all call-sites delegate to it.

## Code Changes

| File | Function | Action |
|------|----------|--------|
| `parser_utils.py` | `parse_multi_result_lhs` | implements the production rule for parsing result LHS with mix of bundled and split forms. |
| `scf_ops.py` | `parse_scf_for` | LHS regex to accept `:N` suffix and delegates to `parse_multi_result_lhs` | 
| `parser.py` | `_parse_general_operation` | fallback LHS regex; delegates to `parse_multi_result_lhs` |
| `parser.py` | `_tokenize_operations`  | Regex updated to recognize `%name:N =` for symbol pickup in split-body mlir lines |
| `tests/test_parser_utils.py` | 7 tests | Unit tests for a variaty of "mixed result LHS", e.g. single, comma, bundled, mixed, malformed inputs |
| `tests/test_dialects_parse.py` | 2 tests | Parsing verification of bundled result from scf.for |
| `tests/test_interpreter.py` | 1 test | Execution test: MLIR code with `%acc:2` runs through |

### DIFF
--- a/ktir_cpu/dialects/arith_ops.py
+++ b/ktir_cpu/dialects/arith_ops.py
@@ -313,15 +313,15 @@ def arith__select(op, context, env):
 # ---------------------------------------------------------------------------
 
 @register_parser("arith.constant")
-def parse_arith_constant(op_text, parse_ctx):
+def parse_arith_constant(op_text, parse_ctx, result=None):
     from ..parser_utils import parse_numeric, parse_tensor_type, parse_dense_payload
 
-    result_match = re.match(r'(%\w+)\s*=\s*arith\.constant\s*(.*)', op_text)
+    result_match = re.match(r'arith\.constant\s*(.*)', op_text)
     if not result_match:
         return None
 
-    result_name = result_match.group(1)
-    rest = result_match.group(2).strip()
+    result_name = result
+    rest = result_match.group(1).strip()
 
     result_type = None
     attributes = {}
@@ -424,24 +424,23 @@ _CMP_PREDICATES = {
 
 
 @register_parser("arith.cmpi", "arith.cmpf")
-def parse_arith_cmp(op_text, parse_ctx):
-    m = re.match(r'(%\w+)\s*=\s*(arith\.cmp[if])\s+', op_text)
+def parse_arith_cmp(op_text, parse_ctx, result=None):
+    m = re.match(r'(arith\.cmp[if])\s+', op_text)
     if not m:
         return None
-    result_name = m.group(1)
-    op_type = m.group(2)
+    op_type = m.group(1)
     pred_pattern, default_type = _CMP_PREDICATES[op_type]
     pred_match = re.search(rf'{re.escape(op_type)}\s+({pred_pattern})', op_text)
     if not pred_match:
         raise ValueError(f"{op_type}: no valid predicate found in: {op_text!r}")
     predicate = pred_match.group(1)
-    operands = [o for o in find_ssa_names(op_text) if o != result_name]
+    operands = find_ssa_names(op_text)
     result_type = default_type
     type_match = re.search(r':\s*(.+)$', op_text)
     if type_match:
         result_type = type_match.group(1).strip()
     return Operation(
-        result=result_name,
+        result=result,
         op_type=op_type,
         operands=operands,
         attributes={"predicate": predicate},
@@ -450,13 +449,12 @@ def parse_arith_cmp(op_text, parse_ctx):
 
 
 @register_parser("arith.sitofp")
-def parse_arith_sitofp(op_text, parse_ctx):
-    result_match = re.match(r'(%\w+)\s*=\s*arith\.sitofp\s+(%\w+)', op_text)
+def parse_arith_sitofp(op_text, parse_ctx, result=None):
+    result_match = re.match(r'arith\.sitofp\s+(%\w+)', op_text)
     if not result_match:
         return None
 
-    result_name = result_match.group(1)
-    operand = result_match.group(2)
+    operand = result_match.group(1)
 
     result_type = "f16"
     to_match = re.search(r'to\s+(f\d+)', op_text)
@@ -464,7 +462,7 @@ def parse_arith_sitofp(op_text, parse_ctx):
         result_type = to_match.group(1)
 
     return Operation(
-        result=result_name,
+        result=result,
         op_type="arith.sitofp",
         operands=[operand],
         attributes={},
@@ -473,13 +471,12 @@ def parse_arith_sitofp(op_text, parse_ctx):
 
 
 @register_parser("arith.bitcast")
-def parse_arith_bitcast(op_text, parse_ctx):
-    result_match = re.match(r'(%\w+)\s*=\s*arith\.bitcast\s+(%\w+)', op_text)
+def parse_arith_bitcast(op_text, parse_ctx, result=None):
+    result_match = re.match(r'arith\.bitcast\s+(%\w+)', op_text)
     if not result_match:
         return None
 
-    result_name = result_match.group(1)
-    operand = result_match.group(2)
+    operand = result_match.group(1)
 
     dst_type = "f32"
     to_match = re.search(r'\bto\s+(\S+)\s*$', op_text)
@@ -487,7 +484,7 @@ def parse_arith_bitcast(op_text, parse_ctx):
         dst_type = to_match.group(1)
 
     return Operation(
-        result=result_name,
+        result=result,
         op_type="arith.bitcast",
         operands=[operand],
         attributes={"dst_type": dst_type},
--- a/ktir_cpu/dialects/arith_ops.py
+++ b/ktir_cpu/dialects/arith_ops.py
@@ -313,14 +313,13 @@ def arith__select(op, context, env):
 # ---------------------------------------------------------------------------
 
 @register_parser("arith.constant")
-def parse_arith_constant(op_text, parse_ctx, result=None):
+def parse_arith_constant(op_text, parse_ctx):
     from ..parser_utils import parse_numeric, parse_tensor_type, parse_dense_payload
 
     result_match = re.match(r'arith\.constant\s*(.*)', op_text)
     if not result_match:
         return None
 
-    result_name = result
     rest = result_match.group(1).strip()
 
     result_type = None
@@ -405,7 +404,7 @@ def parse_arith_constant(op_text, parse_ctx, result=None):
             attributes.setdefault("value", 0)
 
     return Operation(
-        result=result_name,
+        result=None,
         op_type="arith.constant",
         operands=[],
         attributes=attributes,
@@ -424,7 +423,7 @@ _CMP_PREDICATES = {
 
 
 @register_parser("arith.cmpi", "arith.cmpf")
-def parse_arith_cmp(op_text, parse_ctx, result=None):
+def parse_arith_cmp(op_text, parse_ctx):
     m = re.match(r'(arith\.cmp[if])\s+', op_text)
     if not m:
         return None
@@ -440,7 +439,7 @@ def parse_arith_cmp(op_text, parse_ctx, result=None):
     if type_match:
         result_type = type_match.group(1).strip()
     return Operation(
-        result=result,
+        result=None,
         op_type=op_type,
         operands=operands,
         attributes={"predicate": predicate},
@@ -449,7 +448,7 @@ def parse_arith_cmp(op_text, parse_ctx, result=None):
 
 
 @register_parser("arith.sitofp")
-def parse_arith_sitofp(op_text, parse_ctx, result=None):
+def parse_arith_sitofp(op_text, parse_ctx):
     result_match = re.match(r'arith\.sitofp\s+(%\w+)', op_text)
     if not result_match:
         return None
@@ -462,7 +461,7 @@ def parse_arith_sitofp(op_text, parse_ctx, result=None):
         result_type = to_match.group(1)
 
     return Operation(
-        result=result,
+        result=None,
         op_type="arith.sitofp",
         operands=[operand],
         attributes={},
@@ -471,7 +470,7 @@ def parse_arith_sitofp(op_text, parse_ctx, result=None):
 
 
 @register_parser("arith.bitcast")
-def parse_arith_bitcast(op_text, parse_ctx, result=None):
+def parse_arith_bitcast(op_text, parse_ctx):
     result_match = re.match(r'arith\.bitcast\s+(%\w+)', op_text)
     if not result_match:
         return None
@@ -484,7 +483,7 @@ def parse_arith_bitcast(op_text, parse_ctx, result=None):
         dst_type = to_match.group(1)
 
     return Operation(
-        result=result,
+        result=None,
         op_type="arith.bitcast",
         operands=[operand],
         attributes={"dst_type": dst_type},
--- a/ktir_cpu/dialects/ktdp_ops.py
+++ b/ktir_cpu/dialects/ktdp_ops.py
@@ -34,7 +34,7 @@ from ..ops.comm_ops import CommPlan, RingReduceBackend
 from ..ops.grid_ops import GridOps
 from ..ops.memory_ops import MemoryOps
 from ..parser_ast import enumerate_membership_keys, parse_affine_map, parse_affine_set
-from ..parser_utils import _extract_bracket_content, extract_named_attr, find_ssa_names, parse_attr_block, parse_multi_result_lhs, parse_tensor_type, split_top_level
+from ..parser_utils import _extract_bracket_content, extract_named_attr, find_ssa_names, parse_attr_block, parse_tensor_type, split_top_level
 from .registry import ParseContext, register, register_parser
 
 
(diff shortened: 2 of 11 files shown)


yuhaohaoyu:
This PR is ready for review, passing tests, and the signing are fine. Thanks. @fabianlim @lchu6

fabianlim:
Thanks @yuhaohaoyu — the fix is correct and the helper consolidation is the right move.

One structural observation worth considering: the LHS (`%result =`, `%a, %b =`, `%acc:2 =`) is not op-specific MLIR syntax, yet each dialect parser currently receives the full raw text and is responsible for stripping it independently. This PR brings all three sites to the same helper, which is good — but the underlying issue is that LHS stripping happens **three separate times** with three separate regexes.

The cleaner fix would be to strip the LHS **once** in `_parse_operation_text`, before dispatch, and pass the pre-computed result name(s) alongside the LHS-free body text to every parser. The **mechanical change** that entails:

1. `_parse_operation_text` strips the LHS and computes `result: str | list[str] | None` before calling `dispatch_parser`.
2. `dispatch_parser` (and `_parse_general_operation`) receive `body_text` (LHS-free) instead of `op_text`.
3. `ParserFn` signature gains a `result` keyword argument: `(op_text, parse_ctx, *, result)`.
4. Every registered parser (currently 11: `parse_scf_for`, `parse_get_compute_tile_id`, and the 9 others in `ktdp_ops.py`) drops its own LHS regex and uses the passed-in `result` directly.
5. `parse_scf_for` and `parse_get_compute_tile_id` can remove their `parse_multi_result_lhs` calls entirely.

That's a ~11-function signature change — mechanical but non-trivial.

**Question for you:** is this in scope for this PR, or should it be tracked as a follow-up? The current fix is self-contained and correct; the refactor is purely a cleanliness improvement. Either way is fine — just flagging so it doesn't get lost.

fabianlim:
To make point 4 concrete, here's what the before/after looks like across the three key sites:

---

**`_parse_operation_text` — before (current PR)**
```python
def _parse_operation_text(self, op_text, parse_ctx=None):
    op_text = op_text.strip()
    ...
    parser_fn = dispatch_parser(op_text)       # full text including LHS
    if parser_fn:
        return parser_fn(op_text, ctx)         # parser must strip LHS itself
    return self._parse_general_operation(op_text)  # same
```

**`_parse_operation_text` — after**
```python
def _parse_operation_text(self, op_text, parse_ctx=None):
    op_text = op_text.strip()
    ...
    # Strip LHS once, here, before any dialect sees the text
    lhs_match = re.match(
        r'((?:%\w+(?::\d+)?\s*,\s*)*%\w+(?::\d+)?)\s*=\s*(.*)', op_text, re.DOTALL
    )
    if lhs_match:
        names = parse_multi_result_lhs(lhs_match.group(1))
        result = names if len(names) > 1 else names[0]
        body_text = lhs_match.group(2).strip()
    else:
        result = None
        body_text = op_text

    parser_fn = dispatch_parser(body_text)     # dispatch on LHS-free text
    if parser_fn:
        return parser_fn(body_text, ctx, result=result)
    return self._parse_general_operation(body_text, result=result)
```

---

**`parse_scf_for` — before (current PR)**
```python
def parse_scf_for(op_text, parse_ctx):
    outer_result = None
    outer_match = re.match(r'((?:%\w+(?::\d+)?\s*,\s*)*%\w+(?::\d+)?)\s*=\s*scf\.for\s+', op_text)
    if outer_match:
        names = parse_multi_result_lhs(outer_match.group(1))
        outer_result = names if len(names) > 1 else names[0]

    scf_match = re.match(
        r'(?:(?:%\w+(?::\d+)?\s*,\s*)*%\w+(?::\d+)?\s*=\s*)?scf\.for\s+(%\w+)\s*=\s*(%\w+)\s+to\s+(%\w+)\s+step\s+(%\w+)',
        op_text
    )
    ...
    result_name = outer_result if outer_result else iter_var
```

**`parse_scf_for` — after**
```python
def parse_scf_for(op_text, parse_ctx, *, result=None):
    # op_text is now LHS-free: "scf.for %i = %c0 to %n step %c1 ..."
    # No optional leading %x = prefix to handle
    scf_match = re.match(
        r'scf\.for\s+(%\w+)\s*=\s*(%\w+)\s+to\s+(%\w+)\s+step\s+(%\w+)',
        op_text
    )
    ...
    result_name = result if result is not None else iter_var
```

The two regex patterns that handle the optional `%... =` prefix (one to extract it, one to skip over it) both disappear.

---

**`parse_get_compute_tile_id` — before (current PR)**
```python
def parse_get_compute_tile_id(op_text, parse_ctx):
    m = re.match(r"^(.*?)\s*=\s*ktdp\.get_compute_tile_id\s*:\s*([^{(]*)\s*$", op_text)
    if not m:
        return None
    names = parse_multi_result_lhs(m.group(1))
    ...
    result = names[0] if len(names) == 1 else names
    return _make_compute_tile_id_op(result)
```

**`parse_get_compute_tile_id` — after**
```python
def parse_get_compute_tile_id(op_text, parse_ctx, *, result=None):
    # op_text is now LHS-free: "ktdp.get_compute_tile_id : index, index"
    m = re.match(r"ktdp\.get_compute_tile_id\s*:\s*([^{(]*)\s*$", op_text)
    if not m:
        return None
    # result already computed and validated upstream — no parse_multi_result_lhs needed
    ...
    return _make_compute_tile_id_op(result)
```

---

The 9 other parsers in `ktdp_ops.py` that don't touch the LHS today only need the signature addition (`*, result=None`) — no logic change.

`_parse_general_operation` similarly drops its LHS regex and becomes a straight op-body parser accepting `result` as a parameter.

yuhaohaoyu:
@fabianlim Did the hoisting, 2 to 1 use of the parse LHS op. 

The commit message reposting here for ref: 
```
    Hoisting LHS stripping in _parse_operation_text before dialect dispatch
    
    1. Hoist parse_multi_result_lhs from parse_scf_for and parse_get_compute_tile_id into
       _parse_operation_text, for extensible multi-result LHS parsing at a single site.
    2. Side effect: redefine ParserFn API to pass result=None across all registered
       parsers and drop the LHS-stripping regex prefix from each of 28 result-producing
       parser ops (use result param) and 5 yield ops (signature-only, unused).
```

Likely we used similar prompt, spent some time to toast mine on if and why indeed it need update the signatures of all parser ops, not just ktdp_ops.py in your comments above. Here is his defense and I do not find a counter argument yet. Sharing it here: 

---

Yes, we must cover all of them. The reason is mechanical:

`_parse_operation_text` calls every dispatched parser uniformly:
```python
parser_fn(body_text, ctx, result=result)
```

The actual scope is every parser registered via `@register_parser` across all 5 dialect files:

| File | Result-producing | Yield/no-result |
|------|-----------------|-----------------|
| arith_ops.py | 4 | 0 |
| scf_ops.py | 1 | 2 |
| linalg_ops.py | 6 | 1 |
| tensor_ops.py | 10 | 1 |
| ktdp_ops.py | 7 | 2 |
| **Total** | **28** | **6** |

The 1006 passing tests confirm nothing was missed — any forgotten parser would crash on dispatch.

fabianlim:
Sorry, I steered you wrong on the `result=` parameter — after thinking it through more carefully, passing `result` into every parser is noisy because it's just a pass-through: the parser receives it only to hand it straight to `Operation(result=result, ...)` without doing anything with it.

A cleaner approach: strip the LHS in `_parse_operation_text` as you've done, but assign `op.result` *after* the parser returns rather than passing it in:

```python
# _parse_operation_text
op = parser_fn(body_text, ctx)         # no result= passed
if op is not None and result is not None:
    op.result = result
return op
```

This means:
- **All 34 parser signatures revert to `(op_text, parse_ctx)`** — no `result=None` anywhere in dialect code. The 6 yield/no-result signature-only changes disappear entirely, and the 28 result-producing parsers return `Operation(result=None, ...)` with the caller responsible for binding the name.
- **`parse_scf_for` simplifies the most** — the entire `outer_match`/`outer_result` block goes away, since `_parse_operation_text` now owns that. It just returns `Operation(result=iter_var, ...)` as the fallback for loops without an outer LHS, and the post-call assignment only fires when `result is not None`, preserving that fallback correctly.
- **`_parse_general_operation`** similarly drops its `result` parameter and the LHS regex — it receives clean body text and returns `result=None`.

The regex removals in the dialect parsers stay — that's the real work and it's correct. But the `result=None` signature churn across all 34 functions was unnecessary and this avoids it entirely.

yuhaohaoyu:
@fabianlim :) I did do a inspection on the bloating design point, there were a parser context surfaced. Glad the 35 signature changes raised alarm. Will update, kept the context around.

yuhaohaoyu:
@fabianlim It's interesting, The stage 1 of the remedy was clear. then it led to stage 2, please read below

**Stage 1: `result=None` cleanup**

Removed `result=None` from all 34 parser signatures and reverted `Operation(result=result, ...)` to `Operation(result=None, ...)`. 
The caller (`_parse_operation_text`) now assigns `op.result = result` after the parser returns - parsers never see or handle result names.

| File | Signatures reverted | Operation result reverted |
|------|---|---|
| `scf_ops.py` | 3 | 1 (scf.for uses `iter_var` as default) |
| `arith_ops.py` | 4 | 4 |
| `linalg_ops.py` | 7 | 6 |
| `tensor_ops.py` | 12 (incl. helper) | 10 |
| `ktdp_ops.py` | 9 | 7 |
| **Total** | **35** | **28** |

Core logic for result association for the parsing of `scf.for` and `compute_tile_id` in `parser.py` reduced to 2 lines:
```python
if op is not None and result is not None:
    op.result = result
```

---

**Stage 2: Remove `_declared_result_count` validation + 2 dead tests**

The above hoisting of the association of parse-op results introduced 5-line validation block in `_parse_operation_text`, solely to serve 2 tests (`test_get_compute_tile_id_bundled_count_mismatch_rejected`, `test_get_compute_tile_id_comma_count_mismatch_rejected`) that inject artificially malformed MLIR (LHS name count != type annotation count).

This error cannot occur in practice - all MLIR reaching this regex parser is produced by the ktir-mlir-frontend, which enforces the invariant. The PR #45 reviewer (@fabianlim) noted at the time that this validation [belongs in the frontend layer, not the regex parser](https://github.com/torch-spyre/ktir-cpu/pull/45#discussion_r3250141507).

**Removed/cleaned followings:**

1 `_declared_result_count` attribute protocol (1 line in `ktdp_ops.py`)
2 Redundant guards in `parse_get_compute_tile_id` (empty-type checks for regex-impossible cases)
3 5-line validation block in `_parse_operation_text`
4 `_COUNT_MISMATCH_MSG` constant + 2 test methods (22 lines in test file)

**Reference: added validation block for _declared_result_count vs len(results), that induced the investigation and changes in Stage 2**

```
# validation code for the potentially injected scenario
            declared = op.attributes.pop("_declared_result_count", None)
            actual = len(result) if isinstance(result, list) else 1
            if declared is not None and actual != declared:
                raise ValueError(
                    f"operation defines {declared} results but was provided {actual} to bind")
```

```
# above block was added in between these 2 statements
        if op is not None and result is not None:
            op.result = result
```

fabianlim:
## Review finding: `parse_get_compute_tile_id` drops result-count validation

The refactor moves LHS stripping into `_parse_operation_text`, which means individual parsers no longer see the result names and can't cross-check them against the RHS type list. `parse_get_compute_tile_id` was the one op that did this check — and it was silently dropped along with the two tests that covered it.

An input like `%x, %y = ktdp.get_compute_tile_id : index` (2 LHS names, 1 result type) now parses without error and produces `op.result = ["%x", "%y"]` on an op that only has one result value.

**Suggested fix** — use a `_result_count` sentinel attribute:

```python
# ktdp_ops.py — parse_get_compute_tile_id
@register_parser("ktdp.get_compute_tile_id")
def parse_get_compute_tile_id(op_text, parse_ctx: ParseContext):
    if not op_text.startswith("ktdp.get_compute_tile_id"):
        return None
    types_text = re.search(r':\s*(.+)$', op_text)
    if not types_text:
        raise ValueError("ktdp.get_compute_tile_id: no result types specified")
    type_list = [t.strip() for t in types_text.group(1).split(",") if t.strip()]
    return _make_compute_tile_id_op(None, expected_result_count=len(type_list))
```

```python
# _make_compute_tile_id_op — thread the count through as a sentinel attribute
def _make_compute_tile_id_op(result, expected_result_count=None):
    attrs = {}
    if expected_result_count is not None:
        attrs["_result_count"] = expected_result_count
    return Operation(
        result=result,
        op_type="ktdp.get_compute_tile_id",
        operands=[],
        attributes=attrs,
        result_type="index",
    )
```

```python
# parser.py — _parse_operation_text, after op.result is stamped
if op is not None and result is not None:
    op.result = result
    expected = op.attributes.pop("_result_count", None)
    if expected is not None:
        actual = len(result) if isinstance(result, list) else 1
        if actual != expected:
            raise ValueError(
                f"{op.op_type}: {actual} result name(s) but "
                f"{expected} result type(s) in: {op_text!r}"
            )
return op
```

The two deleted tests (`test_get_compute_tile_id_bundled_count_mismatch_rejected`, `test_get_compute_tile_id_comma_count_mismatch_rejected`) should be restored.

The `_result_count` sentinel is opt-in, so no other ops are affected — but any future op that declares a known result count can use the same mechanism.

yuhaohaoyu:
@fabianlim Cool, prepared with this, thus documented the 2 stages. 

Adding heavy tests to the main code path pushed me to look into the conversation between you and derrick on this. I was expecting to see the higher-level code to be straight-lined and clean minded. Now you are saying the said "%x, %y = ktdp.get_compute_tile_id : index (2 LHS names, 1 result type) now" is a valid input from the examples MLIR code that'd be parsing the frontend verifier ? This was your argument in pr 45 review cycles. The sentiments shifted since ? or this is a real issue that can encounter in real load. 

"
The PR https://github.com/torch-spyre/ktir-cpu/pull/45 reviewer (@fabianlim) noted at the time that this validation https://github.com/torch-spyre/ktir-cpu/pull/45#discussion_r3250141507.
"

yuhaohaoyu:
@fabianlim We want to do a deeper check on the mlir frontend parser side ? my sense were supported by the PR45 conversations and a shallow scan of the mlir frontend parser on the production rule of the get_compute_tile_id op. 

1. https://github.com/torch-spyre/ktir-cpu/pull/45#discussion_r3250141507 

The suggested fix certainly make sense. The core question is whether the two synthetic tests are needed? The RE may produce the unmatched cases ?

yuhaohaoyu:
This was the answer from my claude got back to me on how synthetic the two tests are: 

"
Code for the verification from front-end parser, `adapt_op()`: 

```
n_results = len(mlir_op.results)    # from the MLIR C++ IR
if n_results == 0:
    result = None
elif n_results == 1:
    result = mlir_op.results[0].get_name()
else:
    result = [r.get_name() for r in mlir_op.results]
```

The frontend gets `n_results` from `mlir_op.results` — populated by the MLIR verifier during `Module.parse()`. The MLIR C++ infrastructure validates that the number of results declared by the op definition matches the LHS. A mismatch like `%x, %y = ktdp.get_compute_tile_id : index` (2 names, 1 result type) would fail MLIR verification before `adapt_op` ever runs.
"

The mismatch cannot reach the regex parser via the frontend path **if we use the frontend to verify for the MLIR code we feed into ktir-cpu's regex parser**. If follow this step, test-guarded error scenario can only occur if someone hand-writes malformed MLIR text and feeds it directly to the regex parser (which is what the 2 deleted tests do). Of course, there is an if. 

The reason I am iterating on this is that the suggested fix was exactly the post 'stage 1' code (more or less). Then we looked deeper and the stage 2 sounded reasonable to me with many iterations. 

@fabianlim

fabianlim:
The count-mismatch validation is worth keeping. `KTIRParser` is the default parser — `KTIRInterpreter` uses it directly, and any caller passing hand-written or programmatically generated MLIR bypasses the frontend entirely. The "all MLIR comes through the frontend" assumption only holds if `MLIRFrontendParser` is explicitly selected. The two deleted tests are covering a real error path on the default code path.

yuhaohaoyu:
@fabianlim That concludes it then, a false assumption from my side. let me dig the stashed code back.

yuhaohaoyu:
@fabianlim Restored as suggested above. Let me see what I need to do with the out-of-date branch.

yuhaohaoyu:
Sorry for forced push, did rebase without thinking. @fabianlim

fabianlim:
`_extract_operands` still accepts a `result: Optional[str]` parameter but the `result` parameter is now dead — `_parse_general_operation` is the only caller and always passes `None` since the LHS is pre-stripped. The filter on line 631 never fires. Can simplify to `_extract_operands(self, text: str)` and drop the filter.

yuhaohaoyu:
> `_extract_operands` still accepts a `result: Optional[str]` parameter but the `result` parameter is now dead — `_parse_general_operation` is the only caller and always passes `None` since the LHS is pre-stripped. The filter on line 631 never fires. Can simplify to `_extract_operands(self, text: str)` and drop the filter.

@fabianlim done and conflict solved.

fabianlim (approved):
LGTM —  dead arg removal is clean.

fabianlim (approved):
LGTM — _extract_operands dead arg removal is clean.

PR #133, opened 2026-06-19
fix: comm ops inside scf.for / scf.if bodies (issue #131)

### LINKED ISSUE #131: Comm ops (ktdp.inter_tile_*) inside scf.for/scf.if regions are not driven by the scheduler
## Summary

Cross-core communication ops (`ktdp.inter_tile_produce` / `ktdp.inter_tile_reduce`) placed **inside an `scf.for` or `scf.if` region** are not executed correctly. The comm handler returns a generator (the cooperative-scheduler `yield from` / `RecvRequest` protocol), but the region executor runs **synchronously** and binds the generator object itself as the SSA result instead of driving it to completion. A downstream op then receives a Python generator instead of a `Tile`.

## Repro

A kernel that performs an inter-tile reduce inside an `scf.for` loop and uses the reduced tile:

```mlir
scf.for %i = %c0 to %hi step %step iter_args(...) -> (...) {
  // ... compute %partial : tensor<1x64xf16> ...
  %fut = ktdp.inter_tile_produce producer_tiles_per_group = #red_tiles, groups = #red_group
      : tensor<1x64xf16> -> !ktdp.tile_future<tensor<1x64xf16>> { ^bb0(%g: index): ktdp.yield_partial %partial : tensor<1x64xf16> }
  %reduced = ktdp.inter_tile_reduce(%fut) consumer_tiles_per_group = #red_tiles, groups = #red_group,
      identity(%id : tensor<1x64xf16>) : !ktdp.tile_future<tensor<1x64xf16>> -> tensor<1x64xf16> { ^bb0(%l, %r): ... }
  %y = arith.mulf %reduced, %something : tensor<1x64xf16>   // consumes %reduced
  scf.yield ...
}
```

Result:

```
TypeError: float() argument must be a string or a real number, not 'generator'
```

(the op consuming `%reduced` gets the undriven generator).

## Root cause

A comm op's handler is a generator: it `yield`s `RecvRequest`s to the cooperative scheduler and only **returns** its final `Tile` (via `StopIteration.value`) once driven to exhaustion with `yield from`.

`KTIRInterpreter.execute_region` is synchronous and assumes comm ops only appear at the top level of a function body — its comment states *"Comm ops cannot appear inside nested regions in the current spec, so this stays sync."* Only the top-level function body is driven via `yield from`; `scf.for` / `scf.if` bodies are run synchronously.

So when the comm op sits inside a loop/branch, nothing drives its generator — **the generator never returns from inside the loop**. The SSA result is bound to the generator object itself rather than the reduced tile, and the next op that consumes it fails.

## Impact

Any kernel that performs a cross-core reduce **per output tile** cannot run — e.g. split-K matmul / FFN, where the all-reduce of the K-partials naturally sits inside the output-tile (`scf.for`) loop. Top-level comm kernels (e.g. `examples/ktir/ring_reduce.mlir`) work only because their comm ops sit outside any region.

## Expected

Comm ops should be drivable inside `scf.for` / `scf.if` regions: the region executor and the `scf` handlers should propagate the comm `yield from` / `RecvRequest` protocol rather than running synchronously.

fabianlim:
## Summary

Fixes #131 — comm ops inside control-flow region bodies.

**Problem**

`ktdp.inter_tile_reduce` placed inside an `scf.for` or `scf.if` body was silently broken. The interpreter assumed comm ops only appear at the top level of a function: `execute_region` was synchronous, so a comm op's generator was returned as a plain value and never reached the scheduler.

**Solution — two parts** (full design in `docs/cross_core_scheduling.md` § "Comm ops inside control-flow region bodies")

**1. Two-speed region executor on `KTIRInterpreter`**

A new `execute_region_with_comms` generator method sits alongside the original sync `execute_region`. When an op inside the region returns a generator, it does `yield from` — forwarding `RecvRequest`s up to the scheduler. When no comm op fires, it returns immediately at no extra cost.

The original `execute_region` is unchanged and still used for *compute-only* regions (`linalg.reduce` combiners, `tensor.generate`, ktdp combiner bodies). These regions are spec-defined to never contain comm ops; making them generators would break callers like `_run_combiner` that expect a plain value back immediately.

**2. `*_with_comms` handler variants on `ControlOps`**

`ControlOps` gains generator variants (`for_op_with_comms`, `if_op_with_comms`, `while_op_with_comms`) that use `yield from region_executor(...)` instead of a plain call. The sync variants are preserved unchanged. The `scf` dialect handlers always use the comms variants and pass `execute_region_with_comms`.

Only control-flow ops need the comms variant because only their bodies can legally contain comm ops per the spec. Compute ops (`linalg.generic`, `linalg.reduce`, `tensor.generate`, ktdp combiners) operate on local tile data only — a comm op inside one would be a spec violation, not a supported pattern. The routing is therefore static per op type.

Any future op whose body may contain comm ops follows the same pattern: add a `*_with_comms` variant in `ControlOps` and pass `execute_region_with_comms` in the handler.

**Also fixes**: a pre-existing bug where `_for_op_rebind` wrote iter_arg updates into the child scope before `pop_scope()`, causing carried state to reset each iteration.

## Test plan

- `tests/test_grid_scheduler.py::test_inner_loop_comm` — scheduler unit tests: single-iter, multi-iter, plain-loop regression
- `tests/test_examples.py::TestRingReduceInnerLoopExecution` — end-to-end via `examples/ktir/ring_reduce_inner_loop.mlir` (4-core ring reduce inside `scf.for`, K iterations)
- Full suite: 1000 passed, 1 skipped, 13 xfailed

### DIFF
--- a/ktir_cpu/dialects/scf_ops.py
+++ b/ktir_cpu/dialects/scf_ops.py
@@ -28,7 +28,7 @@ def scf__if(op, context, env):
     condition = context.get_value(op.operands[0])
     then_region = op.regions[0] if len(op.regions) > 0 else []
     else_region = op.regions[1] if len(op.regions) > 1 else []
-    return ControlOps.if_op(context, condition, then_region, else_region, env.execute_region)
+    return (yield from ControlOps.if_op_with_comms(context, condition, then_region, else_region, env.execute_region_with_comms))
 
 
 @register("scf.for")
@@ -43,8 +43,8 @@ def scf__for(op, context, env):
     iter_init_operands = op.operands[3:]
     iter_init_values = [context.get_value(n) for n in iter_init_operands]
 
-    result = ControlOps.for_op(
-        context, lb, ub, step, iter_var, body_region, env.execute_region,
+    result = yield from ControlOps.for_op_with_comms(
+        context, lb, ub, step, iter_var, body_region, env.execute_region_with_comms,
         iter_arg_names=iter_arg_names,
         iter_init_values=iter_init_values,
     )
--- a/ktir_cpu/ops/control_ops.py
+++ b/ktir_cpu/ops/control_ops.py
@@ -17,6 +17,14 @@ Control flow compute helpers.
 
 Conditional, loop, and yield primitives used by dialect handlers
 in ``ktir_cpu.dialects``.
+
+Each op has two variants:
+  - Sync (``for_op``, ``if_op``, ``while_op``): plain functions, used by
+    tests and any caller that passes a synchronous region executor.
+  - Comms (``for_op_with_comms``, ``if_op_with_comms``, ``while_op_with_comms``):
+    generator functions, used by ``scf__for`` / ``scf__if`` / ``scf__while``
+    when the body may contain comm ops.  They propagate ``RecvRequest`` yields
+    up to the scheduler via ``yield from execute_region_with_comms(...)``.
 """
 
 from typing import List, Any, Callable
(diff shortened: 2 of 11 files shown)


fabianlim:
@nwang-ibm @WarningRan FYI — PR for review.

ktir_cpu/ops/control_ops.py
```diff
@@ -36,9 +44,53 @@ def __init__(self, values: List[Any]):
...
+    current_values = list(iter_init_values)
+    for name, val in zip(iter_arg_names, current_values):
+        context.set_value(name, val)
+    return iter_arg_names, current_values
+
+
+def _for_op_rebind(result, iter_arg_names, current_values, context):
+    """Re-bind yielded values as iter_args in the current (parent) scope.
+
+    Must be called *after* pop_scope() so the binding lands in the parent scope
```
nwang-ibm, line 64: Should we do pop_scope() here or add an assertion to make sure pop_scope() is called before rebind?

ktir_cpu/dialects/scf_ops.py
```diff
@@ -43,8 +43,8 @@ def scf__for(op, context, env):
     iter_init_operands = op.operands[3:]
     iter_init_values = [context.get_value(n) for n in iter_init_operands]
 
-    result = ControlOps.for_op(
-        context, lb, ub, step, iter_var, body_region, env.execute_region,
+    result = yield from ControlOps.for_op_with_comms(
```
nwang-ibm, line 46: now, result always returns a generator, would that be a concern?

ktir_cpu/ops/control_ops.py
fabianlim: I think its better to call it outside. this is an internal helper that is not meant for general use. calling it outside also makes it read better as "scoped is popped -> rebind output"

ktir_cpu/dialects/scf_ops.py
fabianlim: a generator will run all the way to the end if the conditions to yield are not triggered, so it will behave functionality like a synchronous function, there is no concern there. There is a slight overhead of handling the generator as compared to a sync function, but I feel that is ok

nwang-ibm:
LGTM.

nwang-ibm (approved):

Issue #130, opened 2026-06-18
LX overflow issue


nwang-ibm:
## Summary

MatMul with `M=1024, K=4096, N=4096` and block sizes `(BM=384, BN=512, BK=128)` produces error:

```
Error executing scf.for on core 0: LX scratchpad overflow on core 0:
  requested 393216 bytes, available 294912 bytes (capacity 2097152 bytes)
```

**Summary:** This is NOT a double-counting bug. PR #118's refcount fix works correctly.
The kernel genuinely exceeds 2MB LX under scope-based tracking because multiple dot product outputs are counted in Lx usage. There are gaps between Interpreter and Real Hardware

---
## LX Budget Breakdown (Iteration 2)

```
Parent scope:
  %accum_zero    = 384×512×2 =    393,216 B  (dead but not freed)
  %accum_itr     = 384×512×2 =    393,216 B  (new tile from yield)
  Parent subtotal:                 786,432 B

Body scope (iteration 2):
  %a             = 384×128×2 =     98,304 B
  %b             = 128×512×2 =    131,072 B
  %c_init        = 384×512×2 =    393,216 B
  %a_dot_b       = 384×512×2 =    393,216 B
  Subtotal before %accum_next:   1,015,808 B

Total at overflow point:          1,802,240 B
Requesting %accum_next:             393,216 B
Would-be total:                   2,195,456 B > 2,097,152 B (2 MB)
Overflow by:                         98,304 B
```

---

## Why Iteration 1 Succeeds but Iteration 2 Fails

- **Iteration 1:** `%accum_itr` aliases `%accum_zero` (same `id()`). Refcount = 2, no
  double charge. Parent scope = 393KB (one charge for both names). Body tiles:
  `%a`(98KB) + `%b`(131KB) + `%c_init`(393KB) + `%a_dot_b`(393KB) + `%accum_next`(393KB)
  = 1,408KB. Total = 393 + 1,408 = 1,802KB < 2MB. ✅

- **Iteration 2:** After yield rebind, `%accum_itr` points to a NEW object (from
  `arith.addf` in iter 1). Parent scope = 786KB (two distinct tiles: `%accum_zero`=393KB
  + `%accum_itr`=393KB). Body tiles before overflow:
  `%a`(98KB) + `%b`(131KB) + `%c_init`(393KB) + `%a_dot_b`(393KB) = 1,016KB.
[...]

PR #134, opened 2026-06-19
fix: consume single-use tiles at last-use to reduce peak LX (#130)

fabianlim:
## Background

Issue #130 documents an LX overflow in a matmul kernel that should fit in 2 MB on real hardware but exhausts the interpreter's 2 MB budget. The root cause is that the interpreter's scope-based LX tracking is too conservative in two distinct ways:

**1. Aliased tiles inflate `lx.used`** — `linalg.matmul` binds its output to both `%a_dot_b` and the `outs` buffer `%c_init`. Without tracking that these are the same Python object, `set_value` charges LX twice for a single allocation.

**2. Single-use tiles stay live until scope exit** — within a loop body, tiles that are produced and consumed within a single iteration (e.g. `%accum_zero` initialising the iter-arg) linger in scope until `pop_scope` at the end of the iteration, needlessly inflating peak LX. The fix for (1) in PR #118 corrected the double-charge but left the second gap open. This PR closes gap 2.

### Parse-time use-count map

At `load()` time, `KTIRInterpreter._build_use_counts` does a single recursive walk over all ops and nested regions, counting how many times each SSA name appears as an operand:

```python
_use_counts: Dict[str, int]   # SSA name → total operand occurrences
```

SSA names are unique across a function, so a flat global map suffices — no per-region threading is needed.

### Consume-on-last-use in `get_value`

When `use_count == 1`, the name is being fetched for the last time. `get_value` removes it from scope and frees LX before the consuming op runs, so the new result tile can reuse the freed slot at no net increase in `lx.used`:

```python
if (isinstance(value, Tile)
        and self._use_counts.get(name, 0) == 1
        and scope is self._scope_stack[-1]):   # topmost-scope guard
    del scope[name]
    self._tile_refcount[id(value)] -= 1
    if self._tile_refcount[id(value)] == 0:
        self.lx.used -= value.size_bytes()
```

The **topmost-scope guard** (`scope is self._scope_stack[-1]`) is essential. An outer-scope name with a global `use_count == 1` that is fetched per iteration inside a loop (e.g. `%accum_itr`) would otherwise be consumed on the first iteration and raise `KeyError` on subsequent ones. The guard restricts early-freeing to names that live in the currently-active scope.

### LXOptions — independently toggleable flags

Both mechanisms (alias_dedup from #118 and consume_last_use from this PR) are now controlled by a single `LXOptions` dataclass:

```python
@dataclass
class LXOptions:
    alias_dedup: bool = True       # refcount by id(Tile) — PR #118
    consume_last_use: bool = True  # free at last fetch — this PR
```

Tests use explicit presets to isolate each feature:

```python
_LX_BASELINE = LXOptions(alias_dedup=False, consume_last_use=False)
_LX_DEDUP    = LXOptions(alias_dedup=True,  consume_last_use=False)
_LX_FULL     = LXOptions(alias_dedup=True,  consume_last_use=True)
```

## Known limitation: simultaneous window for iter-arg tiles

Consume-on-last-use cannot eliminate the peak-LX cost of loop carry variables. In each iteration, `%accum_itr` (parent scope) and `%accum_next` (new tile from `arith.addf`) must both be live simultaneously at the moment `%accum_next` is charged. The topmost-scope guard correctly blocks early consumption of `%accum_itr` — it lives in the parent scope and is fetched on every iteration.

After the body scope pops, the `scf.for` rebinds `%accum_itr := %accum_next`, which frees the old carry tile promptly. But the peak has already occurred. See `docs/lx_tracking.md` for the full accounting.

## Changes

- `ktir_cpu/memory.py` — adds `LXOptions` dataclass
- `ktir_cpu/grid.py` — wires `lx_options` into `CoreContext`; branches on each flag in `set_value`, `get_value`, `pop_scope`
- `ktir_cpu/interpreter.py` — adds `_build_use_counts`; installs `_use_counts` on all cores before `execute_with_communication`
- `tests/test_lx_scoping.py` — restructured around `_LX_BASELINE`/`_LX_DEDUP`/`_LX_FULL` presets; module-level MLIR constants; ASCII-art accounting in docstrings
- `tests/test_examples.py` — removes `test_softmax_lx_overflow` (premise invalidated; overflow property tested directly in `test_lx_scoping.py`)
- `docs/lx_tracking.md` — new design doc covering both complications, the topmost-scope guard, the simultaneous-window limitation, and `LXOptions`

### DIFF
--- a/ktir_cpu/grid.py
+++ b/ktir_cpu/grid.py
@@ -23,7 +23,7 @@ from collections import deque
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Callable, Dict, Generator, List, Optional, Set, Tuple, Any
 from .ir_types import Tile
-from .memory import LXScratchpad, SpyreMemoryHierarchy, HBMSimulator
+from .memory import LXOptions, LXScratchpad, SpyreMemoryHierarchy, HBMSimulator
 
 if TYPE_CHECKING:
     from .ops.comm_ops import TransferBackend
@@ -84,11 +84,12 @@ class CoreContext:
         # set_value("%m_acc", %m_new) in parent scope: refcount tracks automatically.
     """
 
-    def __init__(self, core_id: int, grid_pos: Tuple[int, int, int], lx: LXScratchpad, hbm: HBMSimulator):
+    def __init__(self, core_id: int, grid_pos: Tuple[int, int, int], lx: LXScratchpad, hbm: HBMSimulator, lx_options: LXOptions = None):
         self.core_id = core_id
         self.grid_pos = grid_pos  # (x, y, z) position in grid
         self.lx = lx              # Core-local LX scratchpad
         self.hbm = hbm            # Shared HBM
+        self.lx_options = lx_options if lx_options is not None else LXOptions()
         self._scope_stack: List[Dict[str, Any]] = [{}]  # bottom = function body
         self._tile_refcount: Dict[int, int] = {}  # id(Tile) -> refcount
         self._lx_next_ptr_stack: List[int] = []  # watermarks for lx.next_ptr rewind on pop_scope
@@ -224,9 +225,12 @@ class CoreContext:
         scope = self._scope_stack.pop()
         for value in scope.values():
             if isinstance(value, Tile):
-                self._tile_refcount[id(value)] -= 1
-                if self._tile_refcount[id(value)] == 0:
-                    del self._tile_refcount[id(value)]
+                if self.lx_options.alias_dedup:
+                    self._tile_refcount[id(value)] -= 1
+                    if self._tile_refcount[id(value)] == 0:
+                        del self._tile_refcount[id(value)]
+                        self.lx.used -= value.size_bytes()
+                else:
                     self.lx.used -= value.size_bytes()
         self.lx.next_ptr = self._lx_next_ptr_stack.pop()
         assert len(self._lx_next_ptr_stack) == len(self._scope_stack) - 1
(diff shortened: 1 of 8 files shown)


fabianlim:
@WarningRan @nwang-ibm FYI

ktir_cpu/grid.py
```diff
@@ -278,11 +297,17 @@ def set_value(self, name: str, value: Any):
...
-            self._tile_refcount[id(value)] = self._tile_refcount.get(id(value), 0) + 1
 
-    def get_value(self, name: str) -> Any:
+    def get_value(self, name: str, *, peek: bool = False) -> Any:
         """Get SSA value, searching top-to-bottom.
 
+        On last use (use_count == 1), pops the value from scope and frees LX
+        immediately so the caller can reuse or replace the buffer at no net cost.
+
+        Args:
```
nwang-ibm, line 307: two Args: sections here.

nwang-ibm:
Not critical, but may worth of adding a test for peek=True followed by normal get_value, i.e. call get_value(name, peek=True) then get_value(name). Verify peek doesn't affect subsequent consumption. Currently tested only via integration (latency tracker path).

nwang-ibm:
While reviewing and testing this PR, I notice another potential bug, which counting 3 copies of Matmul output starting from second iterations, due to the linalg.matmul handler. I created a new issue [135](https://github.com/torch-spyre/ktir-cpu/issues/135), since current PR fixed the issues it intended to fix.

Also, make a note here. the matmul kernel example (matmul_small.mlir) inside of examples folder does not use fused matmul which also limits the lx usage.  (cc @WarningRan )

fabianlim:
@nwang-ibm 

> Also, make a note here. the matmul kernel example (matmul_small.mlir) inside of examples folder does not use fused matmul which also limits the lx usage.

what do you mean by fused matmul

nwang-ibm:
In the current kernel, %c_init is initialized with an empty tensor, which will stay in lx.
```
%c_init = tensor.empty() : tensor<8x32xf16>
%a_dot_b = linalg.matmul ins(%a, %b : tensor<8x32xf16>, tensor<32x32xf16>)
outs(%c_init : tensor<8x32xf16>) -> tensor<8x32xf16>
```
I tested this way, and seems working fine.
```
// CHANGED: fused matmul — accumulates directly into %accum_itr
      // Interpreter computes: outs + ins[0] @ ins[1] = %accum_itr + %a @ %b
      // Result is a NEW tile object (not in-place), so %accum_next costs BM×BN×2 bytes
      %accum_next = linalg.matmul ins(%a, %b : tensor<{BLOCK_M}x{BLOCK_K}xf16>, tensor<{BLOCK_K}x{BLOCK_N}xf16>)
                                  outs(%accum_itr : tensor<{BLOCK_M}x{BLOCK_N}xf16>) -> tensor<{BLOCK_M}x{BLOCK_N}xf16>
```

fabianlim:
@nwang-ibm 
 > In the current kernel, %c_init is initialized with an empty tensor, which will stay in lx.

Ok i see so there is a workaround, but we should handle #135 regardless, and I proposed a more universal way to handle it in that comment

nwang-ibm (approved):
LGTM

PR #136, opened 2026-06-21
fix: in-place outs invariant + structural assertion (#135)

### LINKED ISSUE #135: linalg.matmul handler breaks accumulator alias chain, causing 3× LX overcount
## Summary

The `linalg.matmul` handler creates a **new** `Tile` object for the result, breaking the alias chain between `%accum_zero`, `%accum_itr`, and `%accum_next`. This causes `alias_dedup` (PR #118) to charge LX separately for each SSA name, even though they all map to the same physical accumulator buffer on hardware.

Compute-bound matmul configs (AI > 384) that should fit in 2 MB LX overflow because peak is 3×BM×BN×2 instead of 1×BM×BN×2.

## Context

With the fused matmul kernel pattern:
```mlir
%accum_next = linalg.matmul ins(%a, %b) outs(%accum_itr) -> tensor<BM×BN×f16>
scf.yield %accum_next
```

On hardware, `outs(%accum_itr)` is an in-place accumulation — the result overwrites the same physical buffer. But the interpreter (`ktir_cpu/dialects/linalg_ops.py:323`) creates a new Python object:

```python
result = Tile(acc.data + result.data, acc.dtype, acc.shape)
```

This gives `%accum_next` a different `id()` from `%accum_itr`, so `alias_dedup` treats them as separate allocations.

## Impact

In iteration 2+, three distinct objects coexist in LX:

| SSA Name | Object | LX Charged |
|----------|--------|------------|
| `%accum_zero` | id=A (original, parent scope) | 1×BM×BN×2 |
| `%accum_itr` | id=B (prev iteration's result) | 1×BM×BN×2 |
| `%accum_next` | id=C (this iteration's new Tile) | 1×BM×BN×2 |

**Example:** BM=640, BN=1024, BK=32 (AI=393.8, compute-bound)
- Accumulator: 640×1024×2 = 1280 KB
- Peak: 3×1280 + 104 = **3944 KB → overflows 2 MB**
- Hardware reality: 1×1280 + 104 = **1384 KB → fits easily**

## Proposed Fix

```python
# ktir_cpu/dialects/linalg_ops.py, linalg__matmul handler

# Before:
if isinstance(acc, Tile):
    result = Tile(acc.data + result.data, acc.dtype, acc.shape)
return result

# After:
if isinstance(acc, Tile):
    acc.data += result.data
    return acc
return result
```

`acc.data += result.data` is a NumPy in-place operation. Returning `acc` preserves the same `id()`, so `alias_dedup` sees one refcount key for all three SSA names and charges LX only once.

### Test script

```python
import numpy as np
from ktir_cpu.interpreter import KTIRInterpreter
from ktir_cpu.latency import HardwareConfig
from ktir_cpu.memory import LXOptions

hw = HardwareConfig(
    num_cores=32, clock_ghz=1.2,
    hbm_bandwidth_tb_s=0.2048, ring_bandwidth_tb_s=0.154,
    simd_elements_per_cycle=64*8, systolic_flops_per_cycle=2048,
    transcendental_penalty=4,
)

_LX_DEDUP = LXOptions(alias_dedup=True, consume_last_use=False)
_LX_FULL  = LXOptions(alias_dedup=True, consume_last_use=True)

M, K, N = 2048, 4096, 4096
BM, BN, BK = 640, 1024, 32

mlir = f"""module {{
  func.func @matmul_sweep(
      %a_ptr: index, %b_ptr: index, %c_ptr: index,
      %K_param: index, %BLOCK_SIZE_M: index, %BLOCK_SIZE_N: index, %BLOCK_SIZE_K: index
  ) attributes {{grid = [1, 1]}} {{
    %pid_m, %pid_n = ktdp.get_compute_tile_id : index, index
    %a_view = ktdp.construct_memory_view %a_ptr, sizes: [{M}, {K}], strides: [{K}, 1] {{
      coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + {M-1} >= 0, d1 >= 0, -d1 + {K-1} >= 0)>,
      memory_space = #ktdp.spyre_memory_space<HBM>
    }} : memref<{M}x{K}xf16>
    %b_view = ktdp.construct_memory_view %b_ptr, sizes: [{K}, {N}], strides: [{N}, 1] {{
      coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + {K-1} >= 0, d1 >= 0, -d1 + {N-1} >= 0)>,
      memory_space = #ktdp.spyre_memory_space<HBM>
    }} : memref<{K}x{N}xf16>
    %c_view = ktdp.construct_memory_view %c_ptr, sizes: [{M}, {N}], strides: [{N}, 1] {{
      coordinate_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + {M-1} >= 0, d1 >= 0, -d1 + {N-1} >= 0)>,
      memory_space = #ktdp.spyre_memory_space<HBM>
    }} : memref<{M}x{N}xf16>
    %offs_am = arith.muli %pid_m, %BLOCK_SIZE_M : index
    %offs_bn = arith.muli %pid_n, %BLOCK_SIZE_N : index
    %accum_zero = arith.constant dense<0.0> : tensor<{BM}x{BN}xf16>
    %c = scf.for %off_k = %c0 to %K_param step %BLOCK_SIZE_K
        iter_args(%accum_itr = %accum_zero) -> (tensor<{BM}x{BN}xf16>) {{
      %a_acc = ktdp.construct_access_tile %a_view[%offs_am, %off_k] {{
        access_tile_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + {BM-1} >= 0, d1 >= 0, -d1 + {BK-1} >= 0)>,
        access_tile_order = affine_map<(d0, d1) -> (d0, d1)>
      }} : memref<{M}x{K}xf16> -> !ktdp.access_tile<{BM}x{BK}xindex>
      %b_acc = ktdp.construct_access_tile %b_view[%off_k, %offs_bn] {{
        access_tile_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + {BK-1} >= 0, d1 >= 0, -d1 + {BN-1} >= 0)>,
        access_tile_order = affine_map<(d0, d1) -> (d0, d1)>
      }} : memref<{K}x{N}xf16> -> !ktdp.access_tile<{BK}x{BN}xindex>
      %a = ktdp.load %a_acc : !ktdp.access_tile<{BM}x{BK}xindex> -> tensor<{BM}x{BK}xf16>
      %b = ktdp.load %b_acc : !ktdp.access_tile<{BK}x{BN}xindex> -> tensor<{BK}x{BN}xf16>
      %accum_next = linalg.matmul ins(%a, %b : tensor<{BM}x{BK}xf16>, tensor<{BK}x{BN}xf16>)
                                  outs(%accum_itr : tensor<{BM}x{BN}xf16>) -> tensor<{BM}x{BN}xf16>
      scf.yield %accum_next : tensor<{BM}x{BN}xf16>
    }}
    %c_acc = ktdp.construct_access_tile %c_view[%offs_am, %offs_bn] {{
        access_tile_set = affine_set<(d0, d1) : (d0 >= 0, -d0 + {BM-1} >= 0, d1 >= 0, -d1 + {BN-1} >= 0)>,
        access_tile_order = affine_map<(d0, d1) -> (d0, d1)>
    }} : memref<{M}x{N}xf16> -> !ktdp.access_tile<{BM}x{BN}xindex>
    ktdp.store %c, %c_acc : tensor<{BM}x{BN}xf16>, !ktdp.access_tile<{BM}x{BN}xindex>
    return
  }}
}}"""


def try_run(lx_options, label):
    interp = KTIRInterpreter(latency_config=hw)
    interp.load(mlir)
    orig_prepare = interp._prepare_execution
    def patched(grid_shape):
        orig_prepare(grid_shape)
        for core in interp.grid_executor.cores:
            core.lx_options = lx_options
    interp._prepare_execution = patched

    a = np.random.rand(M, K).astype(np.float16)
    b = np.random.rand(K, N).astype(np.float16)
    c = np.zeros((M, N), dtype=np.float16)
    try:
        interp.execute_function(
            "matmul_sweep", a_ptr=a, b_ptr=b, c_ptr=c,
            K_param=K, BLOCK_SIZE_M=BM, BLOCK_SIZE_N=BN, BLOCK_SIZE_K=BK,
        )
        report = interp.get_latency_report()
        rf = report.roofline()
        print(f"{label}: OK  cycles={report.kernel_cycles:.0f}  "
              f"AI={rf['arithmetic_intensity']:.1f}  "
              f"eff={rf['efficiency']:.1%}  "
              f"bottleneck={report.bottleneck}")
    except MemoryError as e:
        print(f"{label}: OVERFLOW — {e}")


print(f"Config: BM={BM}, BN={BN}, BK={BK}")
print(f"AI = {BM*BN/(BM+BN):.1f} FLOP/B (ridge=384)")
print()
try_run(_LX_DEDUP, "_LX_DEDUP")
try_run(_LX_FULL,  "_LX_FULL ")
```

### Results (after fix applied)

```
Config: BM=640, BN=1024, BK=32
AI = 393.8 FLOP/B (ridge=384)

_LX_DEDUP: OK  cycles=6406144  AI=266.0  eff=59.1%  bottleneck=memory
_LX_FULL : OK  cycles=6406144  AI=266.0  eff=59.1%  bottleneck=memory
```

Broader sweep:

```
    BM     BN     BK      AI   accum     a+b     DEDUP      FULL
----------------------------------------------------------------------
   640   1024     32   393.8  1280 KB   104 KB       OK        OK
   640   1024     64   393.8  1280 KB   208 KB       OK        OK
   640   1024    128   393.8  1280 KB   416 KB       OK        OK
   768   1024     32   438.9  1536 KB   112 KB       OK        OK
   768   1024     64   438.9  1536 KB   224 KB       OK        OK
  1024    768     32   438.9  1536 KB   112 KB       OK        OK
  1024   1024     32   512.0  2048 KB   128 KB OVERFLOW  OVERFLOW
```

Full test suite: **1008 passed, 0 failures**.

## How the Three Mechanisms Compose

| Mechanism | What it does | LX saved |
|-----------|-------------|----------|
| PR #118 `alias_dedup` | `%accum_zero` = `%accum_itr` in iter 1 (same object, refcount) | 1×BM×BN×2 |
| PR #134 `consume_last_use` | Free `%a`, `%b` when matmul fetches them | (BM×BK + BK×BN)×2 |
| **This fix** (in-place matmul) | `%accum_next` = same object as `%accum_itr` (extends alias chain) | 1–2×BM×BN×2 |

- PR #118 already solved `%accum_zero`/`%accum_itr` aliasing (same object in iter 1).
- This fix extends the chain: matmul returns the same object → `%accum_next` shares `id()` with `%accum_itr` → dedup collapses all three to one charge.
- In iteration 2+, without this fix, the yielded result becomes a new object (id=B ≠ id=A), breaking the chain. With this fix, it stays id=A throughout all iterations.

nwang-ibm:
## Background

The `linalg.matmul` and `linalg.batch_matmul` handlers create a **new `Tile` object** for the accumulator result. This breaks the alias chain between `%accum_itr` and `%accum_next` in `scf.for` iter_arg loops. Since `alias_dedup` (PR #118) tracks allocations by `id(Tile)`, each new object receives a separate LX charge — causing 3× overcount of the accumulator buffer.

**Impact:** Compute-bound matmul configs (AI > 384) that should fit in 2 MB LX overflow because peak is `3×BM×BN×2` instead of `1×BM×BN×2`.

Example: BM=640, BN=1024, BK=32 (AI=393.8):
- Before fix: peak = 3×1280 + 104 = 3944 KB → **overflows 2 MB**
- After fix: peak = 1×1280 + 104 = 1384 KB → **fits easily**

## Strategy

Two complementary measures, following @fabianlim's [structural check proposal](https://github.com/torch-spyre/ktir-cpu/issues/135):

**1. Fix the handlers (immediate cure):**
Both `linalg.matmul` and `linalg.batch_matmul` now accumulate in-place (`acc.data += result.data; return acc`), preserving Python object identity. On hardware, the `outs` buffer is a physical address written in-place — the interpreter must mirror this.

**2. Structural assertion (systemic safeguard):**
Rather than relying on developer discipline for every future outs-bearing handler, we assert the invariant centrally in `_execute_op`: any leaf op (no regions) with `outs_operands` must return the **same Tile object** as its outs buffer. This catches violations at the point of origin — not downstream in LX tracking anomalies.

To support the assertion, a new `outs_operands` field is added to the `Operation` dataclass. It is populated by both the MLIR frontend parser (via a new `_adapt_linalg_matmul_like` adapter) and the fallback regex parser. Currently scoped to ops where `outs` is semantically an in-place accumulator (matmul, batch_matmul); ops where `outs` is just a destination shape (linalg.add, linalg.fill, linalg.generic) are excluded.

## How the three LX mechanisms compose

| Mechanism | What it does | LX saved |
|-----------|-------------|----------|
| PR #118 `alias_dedup` | Same object across SSA names → single charge | 1×BM×BN×2 |
| PR #134 `consume_last_use` | Free `%a`, `%b` after matmul fetches them | (BM×BK + BK×BN)×2 |
| **This fix** (in-place + assertion) | `%accum_next` = same object as `%accum_itr` | 1–2×BM×BN×2 |

## Changes

| File | Change |
|------|--------|
| `ktir_cpu/dialects/linalg_ops.py` | In-place accumulation for both `matmul` and `batch_matmul` |
| `ktir_cpu/ir_types.py` | `outs_operands: List[str]` field on `Operation` |
| `ktir_cpu/mlir_frontend/parser.py` | `_adapt_linalg_matmul_like` adapter + `adapt_op` extracts `_outs_operands` |
| `ktir_cpu/parser.py` | Fallback parser populates `outs_operands` for matmul/batch_matmul |
| `ktir_cpu/interpreter.py` | Structural assertion in `_execute_op` |
| `tests/test_lx_scoping.py` | 7 new tests: end-to-end fused matmul LX, unit-level batch_matmul, assertion verification |

## Verification sweep

```
  BM     BN     BK      AI   accum     a+b     DEDUP      FULL   bound
-------------------------------------------------------------------------
  640   1024     32   393.8   1280 KB    104 KB        OK        OK   compute
  640   1024     64   393.8   1280 KB    208 KB        OK        OK   compute
  640   1024    128   393.8   1280 KB    416 KB        OK        OK   compute
  768   1024     32   438.9   1536 KB    112 KB        OK        OK   compute
  768   1024     64   438.9   1536 KB    224 KB        OK        OK   compute
 1024    768     32   438.9   1536 KB    112 KB        OK        OK   compute
 1024   1024     32   512.0   2048 KB    128 KB  OVERFLOW  OVERFLOW   compute (legitimate — accum = full LX)
```

## Test plan

- [x] Full test suite: `uv run pytest tests/ -q` → 1029 passed, 0 failures
- [x] Issue verification script (BM=640, BN=1024, BK=32): both `_LX_DEDUP` and `_LX_FULL` pass
- [x] Broader sweep: all configs with accum < 2 MB fit; 1024×1024 legitimately overflows
- [x] Structural assertion fires on broken handler (test confirms `AssertionError` with identifying message)
- [x] Correct handler does not trigger assertion

Closes #135

### DIFF
--- a/ktir_cpu/dialects/registry.py
+++ b/ktir_cpu/dialects/registry.py
@@ -41,6 +41,10 @@ _REGISTRY: Dict[str, HandlerFn] = {}
 
 _LATENCY_CATEGORIES: Dict[str, str] = {}
 
+# Ops registered with inplace_outs=True — their outs buffer is an in-place
+# accumulator and the handler must return the same Tile object.
+_INPLACE_OPS_SET: set = set()
+
 
 def get_latency_category(op_name: str) -> str:
     """Return the latency category for *op_name*, defaulting to ``"zero"``.
@@ -84,7 +88,7 @@ def make_parse_context(aliases: Dict[str, str]) -> "ParseContext":
     return ParseContext(aliases=aliases)
 
 
-def register(*op_names: str, latency_category: str = "zero"):
+def register(*op_names: str, latency_category: str = "zero", inplace_outs: bool = False):
     """Decorator that registers a dialect operation handler.
 
     Usage::
@@ -100,14 +104,27 @@ def register(*op_names: str, latency_category: str = "zero"):
         @register()  # infers "arith.addf" from function name "arith__addf"
         def arith__addf(op, context, env):
             ...
+
+    Args:
+        inplace_outs: When True, the op's outs buffer is an in-place accumulator
+            and parsers will populate ``Operation.outs_operands``.  The structural
+            assertion in ``_execute_op`` then verifies the handler returns the
+            same Tile object.
     """
     def decorator(fn: HandlerFn) -> HandlerFn:
         names = op_names or (fn.__name__.replace("__", "."),)
         for name in names:
             _REGISTRY[name] = fn
             _LATENCY_CATEGORIES[name] = latency_category
+            if inplace_outs:
+                _INPLACE_OPS_SET.add(name)
         return fn
     return decorator
+
+
+def is_inplace_outs(op_name: str) -> bool:
+    """Return whether *op_name* was registered with ``inplace_outs=True``."""
+    return op_name in _INPLACE_OPS_SET
 
 
 def dispatch(op_name: str) -> Optional[HandlerFn]:
@@ -116,13 +133,14 @@ def dispatch(op_name: str) -> Optional[HandlerFn]:
 
 
 def temp_registry():
-    """Context manager that restores _REGISTRY and _LATENCY_CATEGORIES on exit."""
+    """Context manager that restores _REGISTRY, _LATENCY_CATEGORIES, and _INPLACE_OPS_SET on exit."""
     from contextlib import contextmanager
 
     @contextmanager
     def _ctx():
         saved_registry = _REGISTRY.copy()
         saved_latency = _LATENCY_CATEGORIES.copy()
+        saved_inplace = _INPLACE_OPS_SET.copy()
         try:
             yield
         finally:
@@ -130,6 +148,8 @@ def temp_registry():
             _REGISTRY.update(saved_registry)
             _LATENCY_CATEGORIES.clear()
             _LATENCY_CATEGORIES.update(saved_latency)
+            _INPLACE_OPS_SET.clear()
+            _INPLACE_OPS_SET.update(saved_inplace)
 
     return _ctx()
 
--- a/ktir_cpu/dialects/registry.py
+++ b/ktir_cpu/dialects/registry.py
@@ -115,6 +115,25 @@ def dispatch(op_name: str) -> Optional[HandlerFn]:
     return _REGISTRY.get(op_name)
 
 
+def temp_registry():
+    """Context manager that restores _REGISTRY and _LATENCY_CATEGORIES on exit."""
+    from contextlib import contextmanager
+
+    @contextmanager
+    def _ctx():
+        saved_registry = _REGISTRY.copy()
+        saved_latency = _LATENCY_CATEGORIES.copy()
+        try:
+            yield
+        finally:
+            _REGISTRY.clear()
+            _REGISTRY.update(saved_registry)
+            _LATENCY_CATEGORIES.clear()
+            _LATENCY_CATEGORIES.update(saved_latency)
+
+    return _ctx()
+
+
 @dataclass
 class ParseContext:
     """Parse-time context passed to dialect parsers alongside the op text.
--- a/ktir_cpu/interpreter.py
+++ b/ktir_cpu/interpreter.py
@@ -36,10 +36,6 @@ from .dtypes import to_ktir_dtype as _ktir_dtype, stick_to_elem_idx as _stick_to
 from .memory import HBMSimulator
 from .ops.memory_ops import hbm_read, hbm_write
 
-# Ops where outs is an in-place accumulator (result includes outs values).
-# The structural assertion only applies to these ops.
-_INPLACE_OPS = frozenset({"linalg.matmul", "linalg.batch_matmul"})
-
 
 class KTIRInterpreter:
     """Main KTIR interpreter.
@@ -230,7 +226,6 @@ class KTIRInterpreter:
         # Structural invariant: accumulating ops must return the same outs object.
         # Only applies to ops where result = f(ins) + outs (in-place accumulation).
         if (op.outs_operands and not op.regions
-                and op.op_type in _INPLACE_OPS
                 and result is not None and not inspect.isgenerator(result)):
             for outs_name in op.outs_operands:
                 try:
--- a/ktir_cpu/interpreter.py
+++ b/ktir_cpu/interpreter.py
@@ -223,6 +223,20 @@ class KTIRInterpreter:
             print(f"Error executing {op.op_type} on core {context.core_id}: {e}")
             raise
 
+        # Structural invariant: leaf ops with outs must return the same object.
+        if (op.outs_operands and not op.regions
+                and result is not None and not inspect.isgenerator(result)):
+            for outs_name in op.outs_operands:
+                try:
+                    outs_tile = context.get_value(outs_name, peek=True)
+                except KeyError:
+                    continue
+                if isinstance(outs_tile, Tile) and isinstance(result, Tile):
+                    assert id(result) == id(outs_tile), (
+                        f"{op.op_type}: handler returned new Tile instead of "
+                        f"mutating outs {outs_name!r}"
+                    )
+
         # Store result in context.
         # Only Tile values (tensor data backed by NumPy arrays) occupy LX.
         # Other result types — TileRef (metadata), AccessTile (coordinates),
--- a/ktir_cpu/interpreter.py
+++ b/ktir_cpu/interpreter.py
@@ -36,6 +36,10 @@ from .dtypes import to_ktir_dtype as _ktir_dtype, stick_to_elem_idx as _stick_to
 from .memory import HBMSimulator
 from .ops.memory_ops import hbm_read, hbm_write
 
+# Ops where outs is an in-place accumulator (result includes outs values).
+# The structural assertion only applies to these ops.
+_INPLACE_OPS = frozenset({"linalg.matmul", "linalg.batch_matmul"})
+
 
 class KTIRInterpreter:
     """Main KTIR interpreter.
@@ -223,8 +227,10 @@ class KTIRInterpreter:
             print(f"Error executing {op.op_type} on core {context.core_id}: {e}")
             raise
 
-        # Structural invariant: leaf ops with outs must return the same object.
+        # Structural invariant: accumulating ops must return the same outs object.
+        # Only applies to ops where result = f(ins) + outs (in-place accumulation).
         if (op.outs_operands and not op.regions
+                and op.op_type in _INPLACE_OPS
                 and result is not None and not inspect.isgenerator(result)):
             for outs_name in op.outs_operands:
                 try:
--- a/ktir_cpu/mlir_frontend/parser.py
+++ b/ktir_cpu/mlir_frontend/parser.py
@@ -134,6 +134,11 @@ class MLIRTypeAdapter:
         handler(mlir_op, attributes, result_type, operands)
 
         outs_operands = attributes.pop("_outs_operands", [])
+        if not outs_operands:
+            from ..dialects.registry import is_inplace_outs
+            if is_inplace_outs(mlir_op.name):
+                from ..parser_utils import extract_outs_operands
+                outs_operands = extract_outs_operands(mlir_op.get_asm())
 
         return Operation(
             result=result,
@@ -216,29 +221,11 @@ MLIRTypeAdapter.install(
     "ktdp.get_compute_tile_id",
     "ktdp.load",
     "ktdp.store",
+    "linalg.matmul",
+    "linalg.batch_matmul",
     # emitted by the bindings walk but not present in text IR
     "ktdp.region_terminator",
 )(_no_attrs)
-
-
-def _adapt_linalg_matmul_like(mlir_op, attributes, result_type, operands):
-    """Handler for linalg.matmul/batch_matmul.
-
-    Populates _outs_operands so the structural assertion in _execute_op can
-    verify the handler returns the same Tile object as outs (in-place invariant).
-    If a future op has the same accumulate-into-outs semantics (e.g. linalg.conv),
-    add it to the MLIRTypeAdapter.install() call below.
-    """
-    from ..parser_utils import extract_outs_operands
-    outs = extract_outs_operands(mlir_op.get_asm())
-    if outs:
-        attributes["_outs_operands"] = outs
-
-
-MLIRTypeAdapter.install(
-    "linalg.matmul",
-    "linalg.batch_matmul",
-)(_adapt_linalg_matmul_like)
 
 
 @MLIRTypeAdapter.install("scf.for")
--- a/ktir_cpu/mlir_frontend/parser.py
+++ b/ktir_cpu/mlir_frontend/parser.py
@@ -133,6 +133,8 @@ class MLIRTypeAdapter:
             )
         handler(mlir_op, attributes, result_type, operands)
 
+        outs_operands = attributes.pop("_outs_operands", [])
+
         return Operation(
             result=result,
             op_type=mlir_op.name,
@@ -140,6 +142,7 @@ class MLIRTypeAdapter:
             attributes=attributes,
             result_type=result_type,
             regions=regions,
+            outs_operands=outs_operands,
         )
 
     def adapt_block(self, block) -> List[Operation]:
@@ -158,8 +161,6 @@ MLIRTypeAdapter.install(
     "func.return",
     "linalg.add",
     "linalg.fill",
-    "linalg.matmul",
-    "linalg.batch_matmul",
     "linalg.yield",
     "scf.yield",
     "scf.if",
@@ -218,6 +219,24 @@ MLIRTypeAdapter.install(
     # emitted by the bindings walk but not present in text IR
     "ktdp.region_terminator",
 )(_no_attrs)
+
+
+def _adapt_linalg_matmul_like(mlir_op, attributes, result_type, operands):
+    """Handler for linalg.matmul/batch_matmul: fixed 2 ins, 1 outs.
+
+    Populates _outs_operands so the structural assertion in _execute_op can
+    verify the handler returns the same Tile object as outs (in-place invariant).
+    If a future op has the same accumulate-into-outs semantics (e.g. linalg.conv),
+    add it to the MLIRTypeAdapter.install() call below.
+    """
+    if len(operands) > 2:
+        attributes["_outs_operands"] = operands[2:]
+
+
+MLIRTypeAdapter.install(
+    "linalg.matmul",
+    "linalg.batch_matmul",
+)(_adapt_linalg_matmul_like)
 
 
 @MLIRTypeAdapter.install("scf.for")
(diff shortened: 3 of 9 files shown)


nwang-ibm:
@fabianlim @WarningRan FYI — PR for review.

tests/test_lx_scoping.py
```diff
@@ -1193,3 +1193,287 @@ def test_addf_loop_body_lx_accounting(self):
...
+        interp = KTIRInterpreter()
+        interp.load('module { func.func @dummy() attributes {grid = [1]} { return } }')
+
+        with pytest.raises(AssertionError, match="handler returned new Tile"):
+            interp._execute_op(op, ctx)
+
+        # Clean up registry
+        from ktir_cpu.dialects.registry import _REGISTRY
+        del _REGISTRY["test.broken_matmul"]
+
```
fabianlim, line 1439: `del _REGISTRY[...]` after the assertion fires leaves the registry dirty if the test raises before reaching this line. Better to use a context manager that snapshots both `_REGISTRY` and `_LATENCY_CATEGORIES` on enter and restores them on exit, so cleanup is guaranteed even on test failure:

```python
# In registry.py
from contextlib import contextmanager

@contextmanager
def temp_registry():
    saved_registry = _REGISTRY.copy()
    saved_latency = _LATENCY_CATEGORIES.copy()
    try:
        yield
    finally:
        _REGISTRY.clear()
        _REGISTRY.update(saved_registry)
        _LATENCY_CATEGORIES.clear()
        _LATENCY_CATEGORIES.update(saved_latency)
```

Then each test becomes:

```python
with temp_registry():
    @register("test.broken_matmul", latency_category=None)
    def broken_handler(...): ...
    ...
    with pytest.raises(AssertionError, match="handler returned new Tile"):
        interp._execute_op(op, ctx)
# registry restored here regardless of whether the assert fired
```

Note `_LATENCY_CATEGORIES` also needs to be included — `register()` writes to both dicts.

nwang-ibm: addressed accordingly.

ktir_cpu/mlir_frontend/parser.py
```diff
@@ -220,6 +221,24 @@ def _no_attrs(mlir_op, attributes, result_type, operands):
...
+    """Handler for linalg.matmul/batch_matmul: fixed 2 ins, 1 outs.
+
+    Populates _outs_operands so the structural assertion in _execute_op can
+    verify the handler returns the same Tile object as outs (in-place invariant).
+    If a future op has the same accumulate-into-outs semantics (e.g. linalg.conv),
+    add it to the MLIRTypeAdapter.install() call below.
+    """
+    if len(operands) > 2:
+        attributes["_outs_operands"] = operands[2:]
+
```
fabianlim: `operands[2:]` hardcodes the assumption that there are exactly 2 ins operands, which breaks for any op with a different ins count.

`get_asm()` returns the canonical op text with normalized SSA names — for `linalg.matmul` that is:
```
# %0 = linalg.matmul ins(%arg0, %arg1 : tensor<...>, tensor<...>) outs(%arg2 : tensor<...>) -> tensor<...>
```

Parsing `outs(...)` from this is more direct — it reads the outs names from their explicit syntactic group rather than computing a positional slice over a flat operand list. No need for `operandSegmentSizes`:

```python
import re
from ..parser_utils import find_ssa_names

def _adapt_linalg_matmul_like(mlir_op, attributes, result_type, operands):
    outs_match = re.search(r'\bouts\s*\(([^)]+)\)', mlir_op.get_asm())
    if outs_match:
        attributes["_outs_operands"] = find_ssa_names(outs_match.group(1).split(':')[0])
```

ktir_cpu/parser.py
```diff
@@ -620,12 +620,19 @@ def _parse_general_operation(self, text: str) -> Optional[Operation]:
                 attributes["_result_shape"] = type_info["shape"]
                 attributes["_result_dtype"] = type_info.get("dtype", "f16")
 
+        outs_ops = []
+        if op_type in ("linalg.matmul", "linalg.batch_matmul"):
+            outs_match = re.search(r'\bouts\s*\(([^)]+)\)', rest)
```
fabianlim: The `op_type in (...)` guard is a hardcoded whitelist that needs updating every time a new accumulating op is added. Since the `outs(...)` regex already correctly identifies destination-style operands for any op, just run it unconditionally — any op without `outs(...)` in its text will naturally produce an empty list:

```python
outs_ops = []
outs_match = re.search(r'\bouts\s*\(([^)]+)\)', rest)
if outs_match:
    outs_ops = find_ssa_names(outs_match.group(1).split(':')[0])
```

This covers `linalg.matmul`, `linalg.batch_matmul`, `linalg.reduce`, `linalg.generic`, and any future op with `outs(...)` without a whitelist.

Also, the same `outs(...)` extraction logic now lives in two places — here and in `_adapt_linalg_matmul_like` in `mlir_frontend/parser.py`. Worth factoring into a shared helper in `parser_utils.py`, e.g.:

```python
# parser_utils.py
def extract_outs_operands(op_text: str) -> List[str]:
    """Extract SSA names from the outs(...) clause of a linalg op text."""
    m = re.search(r'\bouts\s*\(([^)]+)\)', op_text)
    return find_ssa_names(m.group(1).split(':')[0]) if m else []
```

Both parsers then call `extract_outs_operands(...)` — one on `rest`, the other on `mlir_op.get_asm()`.

fabianlim (changes requested):
adding out_operands to Operation looks good, but populating the outs_operands is a little akward. Gave few comments but wondering if it could be better improved.

ktir_cpu/parser_utils.py
```diff
@@ -30,6 +30,15 @@ def find_ssa_names(text: str) -> list[str]:
     return _SSA_RE.findall(text)
 
 
+_OUTS_RE = re.compile(r'\bouts\s*\(([^)]+)\)')
+
+
+def extract_outs_operands(op_text: str) -> list[str]:
+    """Extract SSA names from the ``outs(...)`` clause of a linalg op text."""
+    m = _OUTS_RE.search(op_text)
+    return find_ssa_names(m.group(1).split(':')[0]) if m else []
```
fabianlim: `split(':')[0]` only captures the first outs operand. For `outs(%C : T1, %D : T2)`, the captured group is `%C : T1, %D : T2`; splitting on `:` and taking `[0]` gives `%C` only — `%D` is dropped.

No current op hits this (matmul/batch_matmul both have a single outs), but the utility is wrong by construction. Suggested fix — split on `,` at the top level, then strip the type from each segment:

```python
def extract_outs_operands(op_text: str) -> list[str]:
    m = _OUTS_RE.search(op_text)
    if not m:
        return []
    names = []
    for segment in split_top_level(m.group(1)):
        names.extend(find_ssa_names(segment.split(':')[0]))
    return names
```

And a unit test to pin the behavior:

```python
def test_extract_outs_operands_single():
    assert extract_outs_operands(
        "linalg.matmul ins(%a, %b : ...) outs(%c : tensor<4x4xf16>)"
    ) == ["%c"]

def test_extract_outs_operands_multi():
    assert extract_outs_operands(
        "linalg.generic ins(%a : ...) outs(%c : tensor<4xf16>, %d : tensor<4xf16>)"
    ) == ["%c", "%d"]
```

ktir_cpu/interpreter.py
```diff
@@ -36,6 +36,10 @@
 from .memory import HBMSimulator
 from .ops.memory_ops import hbm_read, hbm_write
 
+# Ops where outs is an in-place accumulator (result includes outs values).
+# The structural assertion only applies to these ops.
+_INPLACE_OPS = frozenset({"linalg.matmul", "linalg.batch_matmul"})
```
fabianlim: Looking at this more carefully, `_INPLACE_OPS` in the interpreter exists because the fallback parser calls `extract_outs_operands` unconditionally — so `linalg.fill`, `linalg.add` etc. all get a non-empty `outs_operands`, and without the gate the assertion would fire on them incorrectly. The mlir_frontend path does not have this problem since `_no_attrs` simply never sets `_outs_operands`.

The root issue is that `outs_operands` extraction should be opt-in, not automatic. A cleaner design: move the opt-in to `register()` via an `inplace_outs=True` flag. When set, the registry automatically handles `outs_operands` extraction at parse time for that op. Everything else gets `[]` by default.

Concretely:
- `register("linalg.matmul", ..., inplace_outs=True)` — registry populates `outs_operands`
- fallback parser no longer calls `extract_outs_operands` unconditionally
- `_adapt_linalg_matmul_like` in the mlir_frontend becomes unnecessary
- `_INPLACE_OPS` in the interpreter disappears — assertion just checks `op.outs_operands`
- both parsers are now symmetric: `outs_operands` is only non-empty when the op was registered with `inplace_outs=True`

nwang-ibm: make sense. redesign it accordingly.

ktir_cpu/dialects/registry.py
```diff
@@ -100,21 +104,56 @@ def arith_maxf(op, context, env):
...
 
 
+def is_inplace_outs(op_name: str) -> bool:
+    """Return whether *op_name* was registered with ``inplace_outs=True``."""
+    return op_name in _INPLACE_OPS_SET
+
+
 def dispatch(op_name: str) -> Optional[HandlerFn]:
     """Look up the handler for *op_name*, or return ``None``."""
     return _REGISTRY.get(op_name)
```
fabianlim, line 134: Nit: `temp_registry` is a plain function that constructs and returns a context manager via a nested inner function. The idiomatic spelling is `@contextmanager` directly on `temp_registry` with a `yield` — one less layer of indirection.

fabianlim (approved):
LGTM. The `inplace_outs=True` on `register()` is the right design — opt-in at the registration site, both parsers symmetric, `_INPLACE_OPS` gone from the interpreter. `extract_outs_operands` fix and unit tests look good.

Issue #126, opened 2026-06-16
latency analysis: surface cross-core communication (comm) cost


WarningRan:
> **Purpose:** design and surface a **cross-core communication view** for the latency analysis — the metrics, their definitions, and the insights they give. Comm lives on its **own axis**, not on the FLOP roofline. Child of **#123** (which defers all comm analysis here).

## Problem

`notebooks/latency_demo.ipynb` reports per-kernel cycles, time, the compute / memory bottleneck, arithmetic intensity, and a roofline. It has a per-core table with a `comm_cycles` column and a `bottleneck` that can read `comm` — but no real **communication view**: every kernel it loads is embarrassingly parallel (matmul, softmax, attention), so the `comm` column is zero throughout; even with a communicating kernel loaded, the table prints cycles as integers (sub-cycle comm rounds to `0`) and the roofline is FLOP-only. So it cannot answer *"how much latency does a cross-core reduce add, and is comm ever the bottleneck?"*

The underlying cost is **already modelled** — `ktdp.inter_tile_reduce` is costed as `LC.COMM`, `comm_cycles`
is folded into `total_cycles` / `kernel_time_us`, and `bottleneck` already recognises `comm`. What is missing is
a set of **comm metrics** derived from those counters and surfaced.

## Metrics we designed

All fractions are taken on the **critical-path core** (the core with `max(total_cycles)` — the core
`kernel_time_us` comes from). Comm on any other core is hidden behind that core's own work and adds no wall
time, so the critical-core comm is the **exposed communication** (PyTorch **HTA** framing).

| metric | definition | formula | unit |
| --- | --- | --- | --- |
| `comm_fraction` | share of the critical core's time spent on the ring — exposed communication | `comm_cycles / total_cycles` (critical core) | ratio ∈ [0, 1] (shown as %) |
| `comm_time` | absolute exposed comm time on the critical core | `comm_cycles / (clock_ghz · 1e3)` | µs |
| `comm_bytes` | ring bytes the critical core moved | Σ over that core's hops (per-core counter) | bytes |
| temporal breakdown | per-core split of time into compute / memory / comm | `{compute, memory, comm}_cycles / total_cycles` | 3 ratios summing to 1 (shown as %) |

The comm cost itself is **bandwidth-only**: `comm_cycles = comm_bytes / ring_bytes_per_cycle` (no
per-hop startup-latency term — step 3 below).

**Insights these give:**

[...]

PR #138, opened 2026-06-23
fix: exclude cross-core comm bytes from roofline arithmetic intensity

WarningRan:
**Fix roofline arithmetic intensity to exclude cross-core comm (ring) bytes**

Surfaced while adding the comm-latency view (#126); continues the roofline-correctness line of #122 / #129.

`roofline()` computed arithmetic intensity as `flops / total_bytes`, but `total_bytes` lumps in cross-core comm (ring) bytes. Ring bytes don't cross HBM, yet the roofline's bandwidth ceiling is HBM-only — so a comm kernel's attained point exceeds the HBM ceiling (`efficiency > 1`, point above the roof), breaking the invariant `test_roofline_returns_sane_values` pins (`efficiency in [0,1]`). It held until now only because comm-free kernels (matmul / vector-add) carry no ring bytes.

Where comm bytes are used — before → after (❌ marks the bug, ✅ the fix):

| comm bytes used for | before | after |
|---|---|---|
| comm latency (`bytes / ring_bw`) | `result.comm_bytes` → `comm_cycles` | unchanged (already correct) |
| **roofline AI denominator** | ❌ `total_bytes` (incl. ring) → efficiency can exceed 1 | ✅ `dram_bytes` (HBM only) → efficiency ≤ 1 |
| `total_bytes` aggregate | sum over all transports | unchanged contract (now a derived property) |
| ring-byte readout | buried in `total_bytes`, not separable | ✅ `comm_bytes` first-class (newly exposed) |

**Change.** Bytes are bucketed by transport on `CoreLatencyCounters` (`bytes_by_category`, mirroring the existing `flops_by_category` / `cycles_by_category`), exposing `total_bytes` (sum, unchanged contract), `comm_bytes` (ring), and `dram_bytes` (HBM, the `"memory"` category). `roofline()` uses `dram_bytes` for arithmetic intensity — AI = FLOP / DRAM-byte, the standard (Williams) definition. `dram_bytes` is a whitelist (the `"memory"` category), so any non-HBM transport is excluded by default; comm **cycles** are untouched (still in wall time).

**Scope.** No effect on comm-free kernels (`dram_bytes == total_bytes`); the matmul sweep is unaffected.

**Tests.** Existing byte/roofline tests pass (`total_bytes` contract intact). The real-kernel `ring_reduce` latency test now also pins `comm_bytes == per-core ring bytes`, `total_bytes == dram_bytes + comm_bytes`, and `efficiency ≤ 1` (was > 1 before the fix).

### DIFF
--- a/ktir_cpu/latency.py
+++ b/ktir_cpu/latency.py
@@ -116,11 +116,13 @@ class CoreLatencyCounters:
     """Per-core cycle counters."""
     memory_cycles: float = 0.0
     comm_cycles: float = 0.0
-    total_bytes: int = 0
-    # Per-category flops and cycles — keys are LatencyCategory string values.
+    # Per-category flops, cycles, and bytes — keys are LatencyCategory string values.
     # Compute categories: "compute_matmul", "compute_float", "compute_transcendental", "compute_int".
     flops_by_category: Dict[str, float] = field(default_factory=dict)
     cycles_by_category: Dict[str, float] = field(default_factory=dict)
+    # Bytes split by transport: the roofline uses DRAM-only traffic for arithmetic
+    # intensity, while comm/ring bytes stay separately readable (dram_bytes / comm_bytes).
+    bytes_by_category: Dict[str, int] = field(default_factory=dict)
     trace: Optional[List[_TraceEntry]] = None
 
     @property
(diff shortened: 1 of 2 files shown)


WarningRan:

fabianlim:
The `__str__` guard at line 709 uses `total_bytes` (HBM + comm) while `roofline()` now uses `dram_bytes` (HBM only) as the denominator. For a comm-only kernel with zero HBM bytes, `total_bytes > 0` causes the roofline section to print, but `dram_bytes == 0` so every unit shows `AI = inf` and the label reads "inf (no memory traffic)" — technically accurate but slightly misleading since there is comm traffic. Not a correctness issue; just flagging for awareness. @WarningRan

WarningRan:
@fabianlim Good catch. Kept the total_bytes guard — it only gates whether the section prints, and a comm-only kernel is worth showing. Fixed the label instead: on AI == inf it now splits on total_bytes — > 0 → inf (no HBM traffic), == 0 → inf (no memory traffic). So a comm-only kernel reads "no HBM traffic".

fabianlim (approved):
I only reviewed this superficially, but I have a feeling that removing comm bytes is not the correct thing to do. Suggest to @WarningRan to properly come up with a latency modeling design document (we already have a document in the doc/s, so we could choose to extend that)

PR #141, opened 2026-06-26
[Experimental] feat: func.call support — multi-func KTIR modules with persistent LX state

fabianlim:
## Summary

cc: @tnakaike @Prasanth-Chatarasi 

- **`func.call` execution** (`ktir_cpu/dialects/func_ops.py`): callee runs in a
  pushed scope with swapped `_use_counts`; LX scratchpad writes survive scope
  pop so data passed between stages via LX persists across call boundaries.
- **Dialect separation**: `func.return` / `func.call` handlers moved out of
  `scf_ops.py` into a dedicated `func_ops.py`; registered in
  `dialects/__init__.py`.
- **MLIR frontend** (`mlir_frontend/parser.py`):
  `MLIRTypeAdapter.install("func.call")` handler extracts callee name from
  `FlatSymbolRefAttr`. Closes the registry-consistency gap that would surface in
  `test_registry_consistency` once `mlir_ktdp` is available.
- **Examples** (`examples/triton-ktir/`): two LX-fused softmax variants —
  `softmax_lx_fused.ktir` (single func, 5 stages inlined) and
  `softmax_lx_fused_calls.ktir` (6 funcs, `@softmax` calls 5 stage subfunctions
  via top-level `func.call`). Both achieve identical latency (2 HBM passes,
  ~14 466 cycles).
- **Tests**: `TestFuncParsers` in `test_dialects_parse.py`, `TestFunc` in
  `test_dialects_exec.py`, `TestFuncAdapt` in `test_parse_adapt.py`.

## Constraint

`func.call` is only supported at the **top level** of a function body — not
inside `scf.for` / `scf.if` regions. This matches current usage and avoids
re-entrant scope/LX issues that haven't been designed for.

## Test plan

- [ ] `uv run pytest tests/ -v` — 1059 passed (adapt tests skip without `mlir_ktdp`)
- [ ] `softmax_lx_fused.ktir` and `softmax_lx_fused_calls.ktir` both run via
      `test_examples`; max softmax error < 1e-4, row sums 0.999–1.001

### DIFF
--- a/examples/triton-ktir/softmax_lx_fused.ktir
+++ b/examples/triton-ktir/softmax_lx_fused.ktir
@@ -1,0 +1,135 @@
+// Single-func, on-chip-fused KTIR — softmax(x) over [512, 1024] (reduce over the 1024 axis).
+//
+// Same 5 stages as softmax_multi.ktir (max, sub, exp, sum, div), but collapsed into ONE func with
+// ONE fixed grid = [32], and with NO intermediate HBM round-trips: the only HBM traffic is the
+// input load and the output store. The function therefore takes only %X (input) and %C (output) —
+// there are no %mx/%sb/%ex/%sm intermediate pointers (contrast softmax_multi.ktir).
+//
+// How the round-trips are removed: instead of letting each stage repartition (the planner's
+// {mb:16,out:2} / {mb:32,out:1} / {mb:1,out:32}), the kernel PINS one partition for the whole
+// kernel — the reduction layout {mb:16, out:2}:
+//     core = out*16 + mb ;  out in [0,2) -> 256 rows ;  mb in [0,16) -> one feature stick (64 wide)
+// so every core holds the SAME tile [256, 64] (its stick of its row-group) across all 5 stages.
+// Consequences:
+//   - max / sum are the only cross-core ops: a local reduce over the 64 width, then an
+//     inter_tile ALL-REDUCE over the 16-stick `mb` cohort (consumer set == producer set, so every
+//     core in the cohort ends with the full row max/sum — no reduce-to-one, no reload).
+//   - sub / exp / div are PURELY LOCAL: each core already holds its x tile and the all-reduced
+//     row max/sum, so they never touch HBM or another core.
+// This is the fusion "case (a)" win (fusing-triton-kernels.md §4.2): consecutive stages share the
+// slicing, so the hand-off stays on-chip in SSA rather than round-tripping HBM.
+//
+// PHYSICAL layout (data_dti.json): logical [512,1024] is stick-on-inner,
+//   sizes [16, 512, 64] = [stick, row, stick_width], strides [64, 1024, 1].
+// The stick index (dim 0) is the `mb` reduction-cohort axis, so the all-reduce runs over dim 0.
+
+#id3 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
+
+// Cross-core cohorts ({mb:16, out:2}): 16 contiguous tiles per group, 2 groups.
+// Tile i is in group g iff 16*g <= i <= 16*g + 15.  (i = core id = out*16 + mb)
+#cohort16 = affine_set<(i)[g] : (i - 16 * g >= 0, -i + 16 * g + 15 >= 0)>
+#groups2  = affine_set<(g) : (g >= 0, -g + 1 >= 0)>
+
+module @softmax_fused {
+  func.func @softmax(%X: index, %C: index)
+      attributes {grid = [32],
+                  // the reduction partition this kernel is pinned to (the other stages fuse onto it)
+                  numWkSlicesPerDim = {mb = 16, out = 2}} {
+    %c16 = arith.constant 16 : index
+    %c64 = arith.constant 64 : index
+    %c256 = arith.constant 256 : index
+    %c0 = arith.constant 0 : index
+    %ninf = arith.constant 0xFC00 : f16            // -inf, max identity
+    %zero = arith.constant 0.000000e+00 : f16       // 0, add identity
+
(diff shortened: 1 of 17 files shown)


fabianlim:
closing this as its not being considered for now

fabianlim:
reopening at @tnakaike's request

yuhaohaoyu:
**Approve** w/ minor condition:

- Approve note: Recommend this to be merged for it does not introduce new bug, yet expanded ktir-cpu's capacity to life-up the need for e2e emulation of fused software super-kernel. 
- Minor conditional/prerequisite: Extract the code-set of PR-144 (fixing a minor issue of data-consistency) from here

---

I have given it a through review (aggregated time: 10 hrs, 500K tokens.), humanizing the review write-ups. Will post today. 

--- 

Difficult and jungling issues to follow up, **NOT** to muddy this PR:
1. storage space abstraction ? there are HBM, LX, LX-bump-allocated, uncharged. It shows muddy design and abstraction given the co-existence of 'LX' and 'LX-bump-allocated'. 
2. bare minimal in func_ops for the softmax example: call/return implemented; not or not fully for return values, no call in for or if, no multi-level test, etc.

yuhaohaoyu:
My review note is below, provided some background leading to decision to let this PR pass so we can attack the issues exposed by this PR that existed much earlier for the problem of LX-management. 

## PR-141 Scope

To support the processing of ktir-mlir kernels with mlir's func.call op appeared in a composite kernel with calling into sub-kernels. A motivation example is the softmax composite-op which contains 5 steps (max, exp, etc.) where each maybe wrapped into a MLIR function (func.func op). Thus have ktir-cpu processing a kernel with sub-kernel invocations can lead to high-level kernel fussion. For a mem-bounded op softmax, the reuse of LX space across contiguous kernels is the lowest hanging target for runtime (Spyre) optimization. So there is 2 appects. 

## Assessments

1. **support func.call and necessary ops**: The key tech of runtime support is stack-frame. ktir-cpu has implemented such stack-frame for the support of regions in SCF IR (control flow: if, for, etc.), which matches the basic needs for the in-same-segment func call-return needs of non-recursive, single-return-arg funcs. 

2. **support reuse of LX-residing data**: The data flow emulated for the reference patterns in the softmax super kernel is correct but lack of precision. In single-level kernel context, the LX-bump-allocated space is the runtime LX-management moulde with the exclusive ownership of LX space, correct for all load is from HBM to LX. In multi-level kernel context, for the mixed load pattern, the LX-bump space and LX-space together did not model the actual HW scratchpad (software mamnaged cache) well. 

| features  | Notes | Assessment |
| --- | --- | --- |
| func.call | functionality: basics - def, call, return | good build up, approve |
|           | resolved: softmax super kernel from triton-ktir got <br> emulated e2e, small tensor size, no overflow | priority, approve |
|           | not-resolved: incomplete compared with `func` dialect. <br> what about other kernel fusion cases? | non-trivial, follow-up tasks |
| LX data reuse | functionality: emulation correct (by observing data-flow) | good build up, approve | 
|               | not-resolved: questionalble data-space model. space overflow | non-trivial, follow-up task |

## Critical high-level notes

### Complex features on reuse of LX residing data

**Keep-in-mind Input Code**

This was an pseudo code, modeled after the softmax superkernel mlir example code, 

```
# def of funcs
func.func _f1( a, b )
   ...
func.func _f2( a, b )
   ...

# buffer alloc and population in LX
x, y, z = tensor(2, 4, 8)
load x, from HBM(whatever)

# invocations
func.call _f1( x, y )
func.call _f2( y, z )

store z, to HBM(results)
```

**Storage-Spaces**

The existing emulation stack maintains 4 relavent storage spaces: HBM, LX (input pgm specified), LX-emulator-alloc'd, uncharged (interpreter plumbing, nothing to do with real runtime. 


| # | Space | Description | In LX? | 
|---|-------|-------------|:------:|
| 1 | **HBM** | External memory. Data enters/leaves the chip here. The only true off-chip storage. | No | 
| 2 | **LX, program-specified** | Persistent buffers at compile-time addresses (arith.constant). Inter-stage communication channel. | Yes (2MB) | 
| 3 | **LX, bump** | Emulator-assigned working space for active computation. | Yes (same 2MB) | 
| 4 | **_uncharged** | Interpreter plumbing. Not in LX at all (no bump address, no explicit address). Pure Python objects to satisfy the "everything is a Tile" execution model. MLIR-to-numpy bridging. | No | 

### Key Observations

- **2 and 3 share the same physical 2MB** but only 3 is capacity-checked.
- The capacity validator should be `persistent_bytes + bump_bytes ≤ 2MB`, but currently only checks `bump_bytes ≤ 2MB`.
- Maybe other implicit issues. 

### Gist 

Does have a historic reason. Before adding func.call, all loads are HBM-2-LX, all stores are LX-2-HBM. Now func.call examples added LX-2-LX copy, but there **should never be LX-2-LX copy**. This led to multiple bad concepts: 

1. One never do LX-2-LX copy. It's nvidia's `shared_mem + L1_cache`, one would never duplicate in here. 
2. In func.call context, with 'load' statement's source residing in LX, the runtime would turn the 'load' to 'no-op'. Store to 'LX location' can be meaningful, but it'd be a pipeline stage of the math-op dictated by the data-flow core.
3. there are 2 space managers for LX, one that wrote the MLIR input file (tensors are produced into LX with addresses statically generated), one as spyre-runtime (the LX bump-allocator in ktir-cpu). 

### Verdict

The issues existed before func.call introduction, fixing it would across many examples, ops, should be a dedicated and focused LX-emulation model.

fabianlim:
@yuhaohaoyu this PR is experimental we may close it if @tnakaike does not need to use it anymore. If its closed then we shouldnt need to persue the complications added from func.call.

fabianlim:
Closing as discussed in the #160

Issue #25, opened 2026-05-02
[Feature] Fix cross-core communication: ktdp.reduce and ktdp.transfer correctness


fabianlim:
## Summary

The `ktdp.reduce` and `ktdp.transfer` ops are implemented but have known simulation correctness issues that prevent them from being exercised by real kernels. The current BSP-replay execution model in `grid.py` produces incorrect results for any kernel that uses inter-core communication.

**Scope:** This issue addresses ring communication on the SFP ring — the communication pattern used for reduction, top-k, and split-k matmul. Other communication patterns may be addressed in separate issues as needed.

## Current State

- `ktdp.transfer` and `ktdp.reduce` are registered in `ktir_cpu/dialects/scf_ops.py`
- `CommOps` and `RingNetwork` are implemented in `ktir_cpu/ops/comm_ops.py`
- `execute_with_communication` in `ktir_cpu/grid.py` handles multi-pass execution
- All current example kernels are embarrassingly parallel (no comm ops), so these code paths are untested end-to-end

### 1. Multi-round re-execution inflates latency (gap K1)

`execute_with_communication` re-executes the **entire function body** on every core for each communication round. The latency tracker accumulates costs from every round, so kernels using `ktdp.transfer` or `ktdp.reduce` have inflated latency — all operations (loads, compute, stores) counted N times instead of once.

### 2. Cyclic communication produces incorrect results (gap K2)

The BSP replay model breaks for bidirectional exchanges and `ktdp.reduce`:

- `CommOps.reduce` calls `ring.recv_from()` with `pop(0)`, consuming messages. On the next round the consumed messages are gone, producing different intermediate results.
- `CommOps.transfer` calls `ring.send()` every round, so duplicate messages accumulate in the buffer.
- Neither `ring_network` nor `core.values` are reset between rounds.

### 3. No end-to-end test coverage

No MLIR example exercises cross-core communication. We need at least:
- A simple all-reduce example (e.g., sum partial results across cores)
- A point-to-point transfer example (e.g., halo exchange)

## Proposed Fix

Replace the BSP replay loop with a single-pass execution model where each core executes at most once, scheduled in dependency order. Cores blocked on a `recv_from` are retried after other cores complete. Cyclic dependencies are detected and raised as errors.

## Acceptance Criteria

- [ ] `ktdp.reduce` produces correct results for a multi-core sum reduction
- [ ] `ktdp.transfer` delivers tiles without duplication
[...]

PR #144, opened 2026-06-30
MLIR frontend: inter-tile yield adaptors and bb0 region fix

fabianlim:
Adds MLIR-frontend support for the inter-tile communication ops introduced in the previous parser work, and fixes a correctness issue with `^bb0` block argument handling.

**Inter-tile yield adaptors**

- Registers `ktdp.inter_tile_produce` and `ktdp.inter_tile_reduce` with `MLIRTypeAdapter`, parsing their affine-set attributes (`producer_tiles_per_group`, `groups`, `consumer_tiles_per_group`, `producer_dependency_per_consumer`) and unpacking the `!ktdp.tile_future<...>` result type into `partial_tensor_types`.
- Registers `ktdp.yield_partial`, `ktdp.yield_reduced`, and `linalg.max` as no-attr ops.
- Adds `linalg.max` interpreter support alongside `linalg.add` via a shared `_LINALG_BINOP` dispatch table (incidentally touches #2).

**bb0 region fix**

- Ops whose regions carry explicit `^bb0(...)` block argument lists were silently dropping those names, causing mis-resolved SSA names inside the region body. The fix extracts the names via a new `extract_bb0_arg_names` helper and injects a synthetic `region.bb0_args` op at the head of the region so downstream consumers see them.
- Guards `attach_reshape` against a `None` reduced value from the generator to avoid a downstream `AttributeError`.

### DIFF
--- a/ktir_cpu/dialects/linalg_ops.py
+++ b/ktir_cpu/dialects/linalg_ops.py
@@ -246,20 +246,27 @@ def linalg__reduce(op, context, env):
     return result
 
 
-@register("linalg.add", latency_category=LC.COMPUTE_FLOAT)
-def linalg__add(op, context, env):
-    """Elementwise tensor add — ``%c = linalg.add ins(%a, %b) outs(%init)``.
+_LINALG_BINOP = {
+    "linalg.add": lambda a, b: a + b,
+    "linalg.max": np.maximum,
+}
 
-    Standard MLIR named op: result = a + b (the outs buffer provides the
-    destination shape only; its values are not accumulated into in v1).
+
+@register("linalg.add", "linalg.max", latency_category=LC.COMPUTE_FLOAT)
+def linalg__binop(op, context, env):
+    """Elementwise binary named ops (add, max) — ins(%a, %b) outs(%init).
+
+    The outs buffer provides the destination shape only; its values are
+    not accumulated into the result.
     """
     tile_a = context.get_value(op.operands[0])
     tile_b = context.get_value(op.operands[1])
     if not isinstance(tile_a, Tile) or not isinstance(tile_b, Tile):
         raise TypeError(
-            f"linalg.add: ins must be Tiles, got {type(tile_a)} and {type(tile_b)}"
+            f"{op.op_type}: ins must be Tiles, got {type(tile_a)} and {type(tile_b)}"
         )
-    return Tile(tile_a.data + tile_b.data, tile_a.dtype, tile_a.shape)
+    fn = _LINALG_BINOP[op.op_type]
+    return Tile(fn(tile_a.data, tile_b.data), tile_a.dtype, tile_a.shape)
 
 
 @register("linalg.fill")
--- a/ktir_cpu/mlir_frontend/parser.py
+++ b/ktir_cpu/mlir_frontend/parser.py
@@ -713,7 +713,7 @@ def _adapt_inter_tile_reduce(mlir_op, attributes, result_type, operands):
     """Extract consumer_tiles_per_group, groups, optional producer_dependency_per_consumer,
     and _result_shape from the result type, matching the regex parser's output.
     """
-    from ..parser_utils import parse_tensor_type
+    from ..parser_utils import parse_tensor_or_memref_type
     attributes["consumer_tiles_per_group"] = parse_affine_set(
         str(mlir_op.attributes["consumer_tiles_per_group"])
     )
@@ -725,7 +725,7 @@ def _adapt_inter_tile_reduce(mlir_op, attributes, result_type, operands):
             str(mlir_op.attributes["producer_dependency_per_consumer"])
         )
     if result_type is not None:
-        parsed = parse_tensor_type(result_type)
+        parsed = parse_tensor_or_memref_type(result_type)
         if parsed is not None:
             attributes["_result_shape"] = parsed["shape"]
 
--- a/ktir_cpu/mlir_frontend/parser.py
+++ b/ktir_cpu/mlir_frontend/parser.py
@@ -134,9 +134,19 @@ class MLIRTypeAdapter:
         handler(mlir_op, attributes, result_type, operands)
 
         from ..dialects.registry import is_inplace_outs
-        from ..parser_utils import extract_outs_operands
-        outs_operands = (extract_outs_operands(mlir_op.get_asm())
+        from ..parser_utils import extract_outs_operands, extract_bb0_arg_names
+        op_asm = mlir_op.get_asm()
+        outs_operands = (extract_outs_operands(op_asm)
                          if is_inplace_outs(mlir_op.name) else [])
+        bb0_names = extract_bb0_arg_names(op_asm)
+        if bb0_names and regions:
+            regions[0].insert(0, Operation(
+                result=None,
+                op_type="region.bb0_args",
+                operands=[],
+                attributes={"names": bb0_names},
+                result_type=None,
+            ))
 
         return Operation(
             result=result,
@@ -149,18 +159,7 @@ class MLIRTypeAdapter:
         )
 
     def adapt_block(self, block) -> List[Operation]:
-        ops = []
-        block_args = list(block.arguments)
-        if block_args:
-            ops.append(Operation(
-                result=None,
-                op_type="region.bb0_args",
-                operands=[],
-                attributes={"names": [a.get_name() for a in block_args]},
-                result_type=None,
-            ))
-        ops.extend(self.adapt_op(op) for op in block.operations)
-        return ops
+        return [self.adapt_op(op) for op in block.operations]
 
 
 # ---------------------------------------------------------------------------
@@ -174,6 +173,7 @@ def _no_attrs(mlir_op, attributes, result_type, operands):
 MLIRTypeAdapter.install(
     "func.return",
     "linalg.add",
+    "linalg.max",
     "linalg.fill",
     "linalg.yield",
     "scf.yield",
--- a/ktir_cpu/mlir_frontend/parser.py
+++ b/ktir_cpu/mlir_frontend/parser.py
@@ -149,7 +149,18 @@ class MLIRTypeAdapter:
         )
 
     def adapt_block(self, block) -> List[Operation]:
-        return [self.adapt_op(op) for op in block.operations]
+        ops = []
+        block_args = list(block.arguments)
+        if block_args:
+            ops.append(Operation(
+                result=None,
+                op_type="region.bb0_args",
+                operands=[],
+                attributes={"names": [a.get_name() for a in block_args]},
+                result_type=None,
+            ))
+        ops.extend(self.adapt_op(op) for op in block.operations)
+        return ops
 
 
 # ---------------------------------------------------------------------------
@@ -167,6 +178,8 @@ MLIRTypeAdapter.install(
     "linalg.yield",
     "scf.yield",
     "scf.if",
+    "ktdp.yield_partial",
+    "ktdp.yield_reduced",
     # float binary
     "arith.addf", "arith.subf", "arith.mulf", "arith.divf", "arith.remf",
     # float unary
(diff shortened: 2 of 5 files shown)


fabianlim:
CC @kiszk @lchu6 @nwang-ibm — tagging also @lasch on the `_LINALG_BINOP` refactoring bit (touches #2) — worth considering whether that dispatch-table approach fits into the broader linalg ops refactoring tracked there.

yuhaohaoyu:
@fabianlim @lasch 

I spent a day trying to dissect the PR, got a clean picture, I think this PR should be merged so we can move pass the bottleneck and address the "hacky" piece for long term. I am composing my reviewing messaging, summarizing from 20 Claude Code responses. 

In the incoming positive review, the only hole for the current shape of the approach and code, a e2e test looked like a need-to-have in this PR. 

More concise notes will come today.

yuhaohaoyu:
## Overview

**approving** 
(the e2e test will be one of the following PR when the setting is ready, depending on pr-25 from FE repo (link as below). 

Top level assessment statements:

1. Support group-wise collective communication of compute cores is essential to deliver realistic latency projection for effective kernel decomposition. This way, collectively, groups of collaborative cores can collectively hold larger or more data tiles thus reduce the need to have data repeated exhausting the HBM-core data move path. 
2. The feature is not yet a 'documented spec', with open PR for RFC [#23](https://github.com/torch-spyre/ktir-mlir-frontend/pull/23) and for implementation [#25](https://github.com/torch-spyre/ktir-mlir-frontend/pull/25) in ktir-mlir-frontend. There are merged PRs in ktir-cpu that were targetting this desirable feature (#72 <-- #71, #70), and this PR will bring up the feature of group-all-reduce (thought analogy: MPI Group, https://theartofhpc.com/pcse/graphics/recursive-transpose.jpeg)
3. The PR correctly wires inter-tile ops into the MLIR frontend path and fixes a  correctness bug (bb0 block argument loss). Two design comments below on the bb0 fix approach that smells better but should not be blocking while worth serious consideration for follow-up.

---

## Explaining the issue: Hyper simplified usage of inter-tile ops for all_reduce, for MPI people

A very high-level and conceptual pseudo sequence of the inter-tile ops, for MPI people. It's like iAllReduce and Wait.
```
  # pseudo^2 code to align with mpi_iallreduce and mpi_wait
  msg_handle = inter_tile_produce( data, group, ... )
  ret = inter_tile_reduce( msg_handle )
```

Some test input code segment that worth of explanation for this complex yet powerful addition is actually how most linear-algebra kernels got decomposed into computation units.

1. Setup of the data for all-reduce, within a 4-core (tile) group. the num_total_tiles = num_cores (let's assume 32)
```
tile 0 has:  [a₀, a₁, ..., a₆₃]
tile 1 has:  [b₀, b₁, ..., b₆₃]
tile 2 has:  [c₀, c₁, ..., c₆₃]
tile 3 has:  [d₀, d₁, ..., d₆₃]

after step 2/3 below, ALL tiles hold:  [a₀+b₀+c₀+d₀, a₁+b₁+c₁+d₁, ..., a₆₃+b₆₃+c₆₃+d₆₃]
```

2. Define the **async.** collective SEND, data producers

```
#all_tiles = affine_set<(i)[g] : (i >= 0, -i + 3 >= 0)>   // tiles/cores/nodes 0,1,2,3
#one_group = affine_set<(g) : (g == 0)>                   // single group, id'd 0

# fut(ure) here can be taken as a handle for messaging, a ticket for runtime to fulfill, etc.
%fut = ktdp.inter_tile_produce
    producer_tiles_per_group = #all_tiles,
    groups = #one_group
    : tensor<1x64xf16> -> !ktdp.tile_future<tensor<1x64xf16>>
{
  ^bb0(%gid: index):
    ktdp.yield_partial %my_partial : tensor<1x64xf16>
}
```

3. Define the receiving side (in a way)

```
%total = ktdp.inter_tile_reduce(%fut)
    consumer_tiles_per_group = #all_tiles,
    groups = #one_group,
    identity(%zeros : tensor<1x64xf16>)
    : !ktdp.tile_future<tensor<1x64xf16>> -> tensor<64xf16>
{
  ^bb0(%lhs: tensor<1x64xf16>, %rhs: tensor<1x64xf16>):
    %sum = linalg.add ins(%lhs, %rhs ...) ...
    ktdp.yield_reduced %sum : tensor<1x64xf16>
}
```

4. What is bb0 here ? 

It's a MLIR region construct (basic block), named, with arguments. From my understanding (shorter) it has these uses
- match the produce/reduce pair
- define the functor (tiny function easy to be generalized) for the data producer/provider and the reduction. 
  
## Functional changes in the PR

1. Register 2 ops, produce/reduce, in the MLIR adaptor module of ktir-cpu
2. Parse and construct the necessities for bb0 region. 

## Thought process for approval

1. The PR will bringup an important addition to ktir op family for grouped cross-core communication/data-sharing.
2. The bb0 inclusion works and unblocks the feature. 

## Test needed (by AI)

**What's untested:** whether the MLIR frontend adapters produce correctly-shaped objects, same as through the regex-parsing path.

The two parser paths are independent code that must produce identical output:

```
regex parser        → Operation(op_type="ktdp.inter_tile_produce", attributes={...})
                                                                        ↓   
                                                                   interpreter ← tested ✅
                                                                        ↑   
MLIR frontend path  → Operation(op_type="ktdp.inter_tile_produce", attributes={...})
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                      This translation is new code with zero test coverage
```

If `_adapt_inter_tile_produce` emits a slightly wrong attribute key, or packs the affine set differently, or the bb0 fix injects names in wrong order — the interpreter gets malformed input. No existing test catches that.

**Need**
A round-trip test should be in place: e.g. an existing inter-tile test case through the MLIR frontend path and assert same execution result. 

---

### Reference 1: PR Linearage 

```
Layer                    Repo                    PR      Status
─────                    ────                    ──      ──────
Spec (RFC)               ktir-mlir-frontend      #23     OPEN (doc only)
Compiler (tablegen+C++)  ktir-mlir-frontend      #25     OPEN (produce + reduce)
Emulator (Python)        ktir-cpu                #72     MERGED (produce + reduce)
                         ktir-cpu                #144    OPEN (frontend wiring + bb0-fix)
```

### Reference2: Changes at a glance by AI
 
| File | Function/Area | +/- | Purpose |
|------|--------------|-----|---------|
| `ktir_cpu/dialects/ktdp_helpers.py` | `attach_reshape` | +2 | Guard against `None` from generator (non-consumer tiles return `None` from ring) |
| `ktir_cpu/dialects/linalg_ops.py` | `linalg__binop` (was `linalg__add`) | +14/−7 | Refactor to `_LINALG_BINOP` dispatch table; adds `linalg.max` alongside `linalg.add` |
| `ktir_cpu/mlir_frontend/parser.py` | `adapt_op` (generic path) | +10 | Inject synthetic `region.bb0_args` op for ops with `^bb0(...)` block arguments |
| `ktir_cpu/mlir_frontend/parser.py` | `_no_attrs` registration | +3 | Register `linalg.max`, `ktdp.yield_partial`, `ktdp.yield_reduced` as no-attr ops |
| `ktir_cpu/mlir_frontend/parser.py` | `_adapt_inter_tile_produce` | +22 | Parse `producer_tiles_per_group`, `groups`, unpack `!ktdp.tile_future<...>` result |
| `ktir_cpu/mlir_frontend/parser.py` | `_adapt_inter_tile_reduce` | +20 | Parse `consumer_tiles_per_group`, `groups`, optional `producer_dependency_per_consumer`, `_result_shape` |
| `ktir_cpu/parser_utils.py` | `extract_bb0_arg_names` | +15 | New helper: regex extracts SSA names from `^bb0(...)` in op ASM text |
| `tests/mlir_frontend/test_registry_consistency.py` | exclusion list | −10 | Remove "blocked on #23" entries now that adapters exist |

yuhaohaoyu:
## Follow up needed

the bb0 inclusion uses `extract_bb0_arg_names(mlir_op.get_asm())` — regex on the op's text representation — to recover block argument names that are already available via the structured `mlir.ir` API (aka. the same API that `scf.for`'s handler already uses: `list(block.arguments)`).

While `scf.for` gets block args from the structured API; `inter_tile_produce`/`inter_tile_reduce` read them from regex on text, a not so great smell. Two options drafted here make the bb0-treatment more structural that are mostly composed by AI.

---

### Option A — Structured API in `adapt_block` (small change)

Move block-argument extraction into `adapt_block`, using the API `scf.for` already proves works:

```python
def adapt_block(self, block) -> List[Operation]:
    ops = [self.adapt_op(op) for op in block.operations]
    if block.arguments:
        names = [a.get_name() for a in block.arguments]
        ops.insert(0, Operation(
            result=None,
            op_type="region.bb0_args",
            operands=[],
            attributes={"names": names},
            result_type=None,
        ))
    return ops
```

**Effect:**
- Rid of the usage of regex in (`extract_bb0_arg_names`) and the `.get_asm()` call
- Generic for ops with region block arguments
- `scf.for`'s special-case block-arg handling could also be removed (now redundant ?? )
- ~5 line diff, no new dependencies

---

### Option B — `Region` dataclass (long-term design)

The root issue: `Operation.regions` is `List[List[Operation]]` — no slot for block arguments. Modeling regions properly eliminates the synthetic op entirely:

```python
@dataclass
class Region:
    args: List[str]       # block argument SSA names
    ops: List[Operation]  # ops in the block body

@dataclass
class Operation:
    ...
    regions: List[Region]  # replaces List[List[Operation]]
```

Adaptation becomes:

```python
def adapt_region(self, region) -> Region:
    block = list(region.blocks)[0]
    args = [a.get_name() for a in block.arguments]
    ops = [self.adapt_op(op) for op in block.operations]
    return Region(args=args, ops=ops)
```

**Effect:**
- No synthetic `region.bb0_args` op needed anywhere
- Mirrors MLIR's native data model (`Block` has `arguments` + `operations`)
- Eliminates `scf.for`'s special-case handler
- Interpreter's region execution reads `region.args` directly to bind names
- **Larger refactor** — touches `Operation` dataclass, all region consumers in the interpreter, both parser paths

**Not for this PR** — this is a separate architectural cleanup. But noting it here as the direction that would prevent this class of bug from recurring.

ktir_cpu/dialects/linalg_ops.py
```diff
@@ -246,20 +246,27 @@ def linalg__reduce(op, context, env):
...
 
-    Standard MLIR named op: result = a + b (the outs buffer provides the
-    destination shape only; its values are not accumulated into in v1).
+
+@register("linalg.add", "linalg.max", latency_category=LC.COMPUTE_FLOAT)
+def linalg__binop(op, context, env):
+    """Elementwise binary named ops (add, max) — ins(%a, %b) outs(%init).
+
+    The outs buffer provides the destination shape only; its values are
+    not accumulated into the result.
```
yuhaohaoyu, line 260: ok dedup

ktir_cpu/mlir_frontend/parser.py
```diff
@@ -134,9 +134,19 @@ def adapt_op(self, mlir_op) -> Operation:
...
                          if is_inplace_outs(mlir_op.name) else [])
+        bb0_names = extract_bb0_arg_names(op_asm)
+        if bb0_names and regions:
+            regions[0].insert(0, Operation(
+                result=None,
+                op_type="region.bb0_args",
+                operands=[],
+                attributes={"names": bb0_names},
+                result_type=None,
+            ))
```
yuhaohaoyu, line 149: mlir_frontend parser yet using modules from regex-parser: (1) should be no such need; (2) tangled code for the two path. 

vote to address for later. See the long comment on 'follow up'.

ktir_cpu/parser_utils.py
```diff
@@ -31,6 +31,21 @@ def find_ssa_names(text: str) -> list[str]:
...
+
+    Returns the list of SSA names (e.g. ``['%arg3', '%arg4']``) when the op's
+    ASM contains a region body with explicit block arguments, or an empty list
+    when no ``^bb0(...)`` label is present (e.g. scf.for, which carries its
+    block-arg names as op attributes instead).
+    """
+    m = _BB0_RE.search(op_text)
+    if not m:
+        return []
+    return find_ssa_names(m.group(1))
```
yuhaohaoyu, line 48: see 'follow up' comment for a better-smelled bridging with ktir-mlir-frontend. no need for action for this PR.

fabianlim:
@yuhaohaoyu thanks for the thorough review.

**On the e2e test:** the adapt test infrastructure is ready (the `InterpreterTestMixin` refactor is in place for `TestRingReduceExecution` / `TestRingReduceInnerLoopExecution`), but the round-trip test is blocked on ktir-mlir-frontend#25 — `Module.parse()` rejects `ktdp.inter_tile_produce`/`reduce` since the C++ dialect doesn't define them yet. Once #25 merges, adding the adapt classes is a 3-line diff.

**On the bb0 / region follow-up:** shall we open an issue to track the two options you outlined (Option A: structured API in `adapt_block`, Option B: `Region` dataclass)? Happy to file it with your write-up as the description.

ktir_cpu/dialects/linalg_ops.py
fabianlim: Ack.

ktir_cpu/mlir_frontend/parser.py
fabianlim: Agreed — will address in the follow-up issue for the region/bb0 refactor.

ktir_cpu/parser_utils.py
fabianlim: Agreed, no action here. Will track in the follow-up issue with the two options you described.

yuhaohaoyu:
@fabianlim In that case, I think we can merge this one. I saw the implementation issue in the FE repo is being made progressing by you. This is a good way to break the chicken-egg cycle and have the snowball grow. Keep my approving of the PR. @lasch I guess we can go ahead with the merging. :)

Issue #98, opened 2026-06-09
[Feature] Vectorized indirect coord build, reusing one subscript dispatch


fabianlim:
## Background

Split out from #52. PR #79 round-1 added an indirect-path fast path (`_eval_subscript_vectorized`, `_build_indirect_coords_box`, `_enumerate_box_pts`) that was **reverted** in round 2 (review Issue 3). The reason for the revert: it wrote a *second copy* of the whole subscript-kind handling — a parallel `direct` / `direct_expr` / `indirect` evaluator for the array case. Two copies of the same logic drift apart and become a correctness hazard.

This issue lands the same speedup the *right* way: write the dispatch **once** and call it for both the slow (per-point) and fast (whole-tile) cases.

## Problem

`_build_indirect_coords` (`ktir_cpu/ops/memory_ops.py:264`) builds the coordinate list one point at a time:

```python
points = _enumerate_in_vso_order(iat)
coords = []
for pt in points:                                  # <-- runs N times in Python (N can be ~1M)
    coord = []
    for sub in iat.dim_subscripts:
        kind = sub["kind"]
        if kind == "direct":
            coord.append(pt[sub["var_index"]])
        elif kind == "direct_expr":
            coord.append(eval_subscript_expr(sub["subscript"], pt))
        elif kind == "indirect":
            raw_idx = int(next(idx_iters[iv_idx]))
            if raw_idx < 0:
                raise IndexError(...)
            coord.append(raw_idx)
    coords.append(tuple(coord))
```

For an axis-aligned `variables_space_set` (`BoxSet`) with identity `variables_space_order`, the points form a regular grid and the whole loop can be done in numpy at once.

## What to build

Write the `if kind == ...` dispatch **once**, in a helper that operates on columns. The slow path is just this with `N == 1`:

```python
def _coords_for(iat, points, idx_values):
    """Build one coordinate column per dim.

    `points` is shaped (N, ndim): N loop-variable points, each ndim wide.
    Slow path: N == 1. Fast path: N == whole tile. No per-point Python loop.
    """
    cols = []
    for sub in iat.dim_subscripts:
        kind = sub["kind"]
        if kind == "direct":
            cols.append(points[:, sub["var_index"]])
        elif kind == "direct_expr":
            # _eval_node already works on numpy columns unchanged
            cols.append(
                _eval_node(sub["subscript"],
                           [points[:, k] for k in range(points.shape[1])])
            )
        elif kind == "indirect":
[...]

Issue #97, opened 2026-06-09
[Feature] MemoryOps.load/store accept ndarray coords to skip tolist()/tuple() conversion


fabianlim:
## Background

Split out from #52. The indirect fast path (when reinstated, see the BoxSet indirect follow-up) builds coordinates as NumPy arrays, then converts them to `List[Tuple[int, ...]]` via `.tolist()` / `tuple()` because `MemoryOps.load(coords=...)` / `_flat_memory_offsets` consume a list of tuples.

## Problem

That closing conversion is where the entire Python-side overhead of a vectorised coord build lives — it round-trips an ndarray back to Python objects only for `_flat_memory_offsets` to iterate them per coord. Accepting an `ndarray` directly would let the vectorised build skip the conversion and `_flat_memory_offsets` compute offsets via a single `coords @ strides`.

## Scope

- Allow `coords` to be an `(N, ndim)` ndarray in `MemoryOps.load` / `store`.
- Vectorise `_flat_memory_offsets` for the ndarray case (matrix-multiply by strides instead of per-coord Python sum).
- Preserve the list-of-tuples path for back-compat.

## Out of scope

- The vectorised indirect coord build that produces the ndarray (separate follow-up; this is the callee-side enabler).

## References

- `ktir_cpu/ops/memory_ops.py` — `load:439`, `store:523`, `_flat_memory_offsets:407`.
- #52, PR #79 follow-ups.

fabianlim:
cc @nwang-ibm @WarningRan

fabianlim:
Related: #98 (vectorized coord build) is the producer that benefits from the ndarray `coords` interface here; either order is fine, doing both together avoids writing-then-deleting the `tuple()` conversion.

Issue #96, opened 2026-06-09
[Feature] Preserve BoxSet structure under permuted variables_space_order in indirect access


fabianlim:
## Background

Split out from #52 (Issue 4 in the PR #79 review). The indirect access paths (`indirect_load` / `indirect_store`) fall back to full point-list enumeration whenever `variables_space_order` (vso) is non-identity, even when `variables_space_set` is a `BoxSet`.

## Problem

A permuted box is still a box. Pre-permuting the box's `lo` / `hi` by the vso permutation preserves the rectangular structure, so the variable-space iteration could stay O(ndim) construction instead of materialising the full O(∏shape) point list. The current guard treats any non-identity vso as disqualifying, which is overly conservative.

Note: this builds on the indirect BoxSet fast path that was **reverted** in PR #79 (Issue 3). That fast path should land first (or together) — there is nothing to optimise until the indirect BoxSet path exists again.

## Scope

- When `variables_space_set` is `BoxSet` and `vso` is a permutation, pre-permute `lo`/`hi` and drive variable-space iteration from the permuted box rather than enumerating points.
- Per-element indirect index lookups stay per-point (stick locality unpredictable).

## Out of scope

- Non-permutation vso (already correctly rejected with `ValueError`).
- The base indirect BoxSet fast path itself (separate follow-up; prerequisite).

## References

- `ktir_cpu/ops/memory_ops.py` — `indirect_load:589`, `indirect_store`, `_enumerate_in_vso_order:170`, `_build_indirect_coords:264`.
- #52, PR #79 (Issue 3 revert + Issue 4 deferral).

fabianlim:
cc @nwang-ibm @WarningRan

fabianlim:
Depends on #98 (base vectorized coord build) — that must land first; this extends it to the permuted-vso case.

PR #147, opened 2026-06-30
Add block-gather fast path for indirect memory access

### LINKED ISSUE #143: [Feature] Fast Indirection Emulation targeting Blocked Indirection
## Problem

The current implementation of the memory_op indirect-load and indirect-store assumes the worst case scenario the accesses into the data array is fully random, e.g. there is no spatial locality in the data read/write. The killer patterns from MoE expert funneling/selection, paged attention page-level access follow indirection over continuous blocked data. The real LLM-related patterns will need to deal with much smaller space of indexing, more efficient blocked data moving. 

## Opportunity

A brief PoC to show that **400x** speedup for a single CPU run on MacBook-M4. A speedup table is as below. 

 ### Workload B: 64 MB (256 experts, 8 active, hidden dim 2048)

```
Data tensor:    256 x 64 x 2048  (f16) = 64.0 MB
Index tensor:   [8]  (i32)
Gather output:  8 x 64 x 2048   (f16) = 2.0 MB
Total points:   1,048,576
Unique lookups: 8
Ratio:          131,072x
```

**Old path (general inspector-executor, 7 steps):**

| Step | Description                  |    Time (ms) |     % | 
|-----:|------------------------------|-----------:|------:|
|    1 | Enumerate iteration space    |      27.076 |  2.3% |
|    2 | Read index tensors           |     772.084 | 65.9% |
|    3 | Build coordinate list        |     245.077 | 20.9% |
|    4 | Linearize flat offsets       |     126.027 | 10.8% |
|    5 | Gather from HBM              |       0.932 |  0.1% |
|    6 | Reshape                      |       0.079 |  0.0% |
|    7 | Write to LX                  |       0.003 |  0.0% |
|      | **TOTAL**                    | **1171.279** |       |   

**New path (block-gather fast path, 3 steps):**

| Step | Description                  | Time (ms) |     % | 
|-----:|------------------------------|----------:|------:|
|    1 | Read K=8 index values        |     0.025 |  0.9% |
|    2 | Numpy broadcast ?~F~R flat offsets |   1.823 | 68.0% |
|    3 | Gather + reshape + LX write  |     0.834 | 31.1% |
|      | **TOTAL**                    | **2.681** |       |   

**End-to-end speedup: 437x**

---

- **Workload A** (16 MB, 128 experts, 8 active, hidden 1024): 566ms ?~F~R 1.3ms ?~@~T **442x**
- **Workload B** (64 MB, 256 experts, 8 active, hidden 2048): 1171ms ?~F~R 2.7ms ?~@~T **437x**

Both show the same profile: steps 2 and 3 (Python iteration) dominate the old path at ~87%, completely eliminated by the numpy broadcast in the new path.

## Issue Overlap Table

| Coverage | Issue | Title | Overlapped or Related | Explanation |
|:--------:|-------|-------|----------------------|-------------|
| **75%** | #78 | Parallel GridExecutor | Related (motivation removed) | Indirect access was about 99% of per-core wall-clock; now 3ms instead of 1100ms. The primary driver for parallelizing cores was this bottleneck - do not need for MoE workloads now. Still valid for compute-heavy kernels. |
| **50%** | #98 | Vectorized indirect coord build | Directly overlapped | We solved the same problem (eliminate per-point Python loop) for the block-gather subset via numpy meshgrid. The general high-cardinality case (all points truly random) still needs the "single dispatch, column-wise" approach described in the issue. |
| **35%** | #97 | ndarray coords in MemoryOps.load/store | Bypassed for dominant case | Our fast path never enters `MemoryOps.load(coords=...)` -  it goes straight to `mgr.gather(offsets)`. The tuple-conversion overhead is eliminated for block-gather patterns. General indirect path still pays it. |
| **20%** | #67 | read_scattered cross-allocation safety | Avoided by new path | `gather()` fancy-indexes the allocation directly, sidestepping `read_scattered` and its cross-allocation hazard entirely. But the existing general-path callers still use `read_scattered` - the bug remains latent there. |
| **15%** | #102 | HBM/LX memory bootstrap | Tangential (same classes) | We added `gather()` to `HBMSimulator`/`LXScratchpad` and `_place_in_lx` for zero-copy writes - touching the same modules. But the placement API and allocation-aware `next_ptr` are unrelated concerns. |
| **10%** | #96 | Preserve BoxSet under permuted VSO | Known gap in fast path | Our gate requires identity VSO (falls back otherwise). A permuted-VSO block-gather misses the fast path today. If #96 lands the pre-permute logic, our gate could accept it - but we didn't implement that. |
| **10%** | #60 | indirect_store collision semantics | Same behavior inherited | `_block_gather_store` uses the same `flat[offsets] = data` assignment - last-writer-wins. If this issue resolves toward "raise on collision" or reduction semantics, the fast path needs updating too. |
| **5%** | #95 | Strided sub-rectangle fast path | Unrelated (direct access path) | That issue targets non-contiguous rectangular *direct* access. Our work is entirely on the *indirect* access path. No overlap in code or pattern. |
| **5%** | #56 | Non-rectangular VSS | Not addressed | Our fast path requires `BoxSet` (axis-aligned rectangle). Triangular or arbitrary constraint sets fall through to the general path. Orthogonal problem. |

yuhaohaoyu:
## 1. Problem and Motives

**Defining Issues**: #143 

**Motivating issues**: #98 : Vectorized indirect coord build | Solved for block-gather subset; general high-cardinality case remains

## 2. Commit message: 
Block-gather exploits the fact that MoE/paged-attn patterns select few index values over large contiguous blocks.

1. Numpy broadcast replaces per-point Python loop in indirect_load/store
2. Add gather() to memory simulators and zero-copy _place_in_lx
3. Add 23 tests in test_block_gather_fast_path.py
4. Add benchmark scripts under benchmarks/

---

## 3. Code Changes Summary (with commits b83fc3b and e856f84)

| File | +/− | Net |
|------|-----|-----|
| `ktir_cpu/ops/memory_ops.py` | +401 / −11 | **+390** |
| `ktir_cpu/memory.py` | +46 / −0 | +46 |
| `tests/test_block_gather_fast_path.py` | +590 / −0 | +590 |
| `benchmarks/bench_utils.py` | +524 / −0 | +524 |
| `benchmarks/bench_indirect_emul_time.py` | +295 / −0 | +295 |
| `benchmarks/README.md` | +196 / −0 | +196 |
| `benchmarks/configs/indirect_emul.toml` | +99 / −0 | +99 |
| **Total** | **+2,140 / −11** | **+2,129** |

Breakdown by category:

| Category | Lines | Share |
|----------|-------|-------|
| Functional code (`memory_ops.py` + `memory.py`) | +436 | 20% |
| Tests | +590 | 28% |
| Benchmarks (runner + utils + config + README) | +1,114 | 52% |

---

### 4.1 Dispatch: Gate → Fast Path → Fallthrough

```
indirect_load(context, iat)
    │
    ├─ _is_block_gather(iat)?
    │       │
    │       ├─ YES → _block_gather_load(context, iat)      ← fast path
    │       │
    │       └─ NO  → general inspector-executor path       ← unchanged
    │
    └─ (returns Tile)
```

**Conditions gating the "fast path"** (`_is_block_gather` → `_block_gather_analyze`):

1. Exactly **1 indirect dim** among `dim_subscripts`
2. `variables_space_set` is a `BoxSet` (axis-aligned rectangle)
3. **Ratio of Size-of-data-array Size-of-index-array**: `unique_lookups x 16 <= total_points`
   - `unique_lookups` = product of extents for `dep_vars` (variables that drive the index lookup)
   - `total_points` = product of extents of all iteration variable 
   - Ratio >= 16 means each index lookup serves more than 16 output elements (the "block")

**Gate example** — `A[B[i], j]` with `i ∈ [0,4), j ∈ [0,32)`:

```
dep_vars = {0}               # B's address depends on variable i
unique_lookups = extent(i) = 4
total_points = 4 x 32 = 128
ratio = 128 / 4 = 32x >= 16x # --> qualifies
```

The indirect dim, only one, can be anywhere (outermost, middle, innermost). The gate condition only measures iteration-space reduction.  If the condition returns False, indicating multiple indirect dims, non-BoxSet VSS, or ratio < 16, execution falls through to the general inspector-executor path.

### 4.2 Fast Path for Blocked Indirect-Accesses : 4 Steps

Indirection is hard to interprete generally, thus we include an elaborated example to depict the process of the blocked-gather for load and store of indirect-accesses. 

**Step 0: Input to the algorithm**

The exampler LOAD op involving indirection assumes: `A[8, 32]`, strides `[32, 1]`, index `B[4] = [3, 7, 1, 5]`, pattern `Y[i,j] = A[B[i], j]`.

---

**Step 1: Read K unique index values**

Read K=4 elements from HBM, regardless of total output size. Inside the `memory_op.py`

```
_block_gather_read_idx(context, iat, ...)
  - reads B[0..3] from HBM
  - returns idx_values_arr = [3, 7, 1, 5]   (shape (4,))
```

---

**Step 2: Build indirect coordinate grid (reshape for numpy broadcast)**

```python
# dep_var 0 has extent 4; iteration has 2 vars (i, j)
# Place extent-4 in axis 0, size-1 in axis 1:
idx_lookup_shape = [4, 1]
indirect_coord_grid = [[3], [7], [1], [5]]     # shape (4, 1)
```

Exampler Output of the step:

`indirect_coord_grid[i, :]` = the row of A that variable `i` selects. The size-1 axis numpy.broadcasts across all `j`.

---

**Step 3: Compute flat offsets via numpy broadcast addition**

Each dimension contributes one term to a separable sum:

```python
offsets = zeros(4, 32)

# dim 0 (indirect), stride=32 --> broadcast for dim-expansion
offsets += indirect_coord_grid * 32     # boradcast: shape (4,1) --> (4,32)
# output of the numpy broadcast #1
#	offsets = [[96,96,...96 ], [224,224,...224], [32,32,...32 ], [160,160,... 160]]    

# dim 1 (direct, var j), stride=1 --> broadcast-add
offsets += arange(32).reshape(1, 32)    # broadcast rhs: shape (1,32) --> (4,32)

# output of the numpy broadcast #2
# offsets[i, j] = B[i] x 32 + j x 1
#               = [[96,97,...,127], [224,225,...,255], [32,33,...,63], [160,161,...,191]]

offsets = offsets.ravel()   # flatten: 2D --> 1D --> shape (128,)
```

No **Python loops**, and no cpu load/store ops. The broadcast addition is highly optimized. 

---

**Step 4: Gather from HBM + place in LX**

```python
gathered = mgr.gather(offsets, "f16")    # numpy fancy-index: A.flat[offsets]
data = gathered.reshape(4, 32)           # restore output shape
_place_in_lx(context, data)              # zero-copy write to LX scratchpad
```

---

### 4.3 3D data array, Middle-dim example 

Access pattern: `A[i, B[j], k]` with shape `(16, 8, 32)`, `B[4] = [5, 3, 7, 1]`:

Same 4 steps, different shapes:
- `indirect_coord_grid` shape `(1, 4, 1)` - size-4 in axis 1, size-1 elsewhere
- `offsets[i, j, k] = ix256 + B[j]x32 + kx1`
- 2048 elements gathered in one read of indireciton array.

### 4.4 Takeaways: 

- The algorithm does not care if the indirection is the outer-most dimension by construction. We just emulate not really deal with blocked copy. The key is to create a flat `offsets[N]` array fast, where `N = output of the blocked gathor or selection`.  
- A small matter is that the hardware prefetcher, on my laptop, makes the gather near-memcpy speed.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,0 +1,74 @@
+# Benchmarks: Indirect Access Emulation Timing
+
+## Usage
+
+```bash
+cd ktir-cpu
+
+# Fast-path vs general-path comparison + per-step breakdown
+uv run python benchmarks/bench_indirect_emul_time.py block
+
+# Gather memcpy optimization (3-copy vs 1-copy)
+uv run python benchmarks/bench_indirect_emul_time.py gather
+
+# Custom config
+uv run python benchmarks/bench_indirect_emul_time.py block --config configs/custom.toml
+```
+
+## Output
+
+### `block` — fast-path vs general-path + per-step breakdown
+
+```
+Workload |       Shape |    Points | General (ms) | Fast (ms) | Speedup
+---------+-------------+-----------+--------------+-----------+--------
+ xl-262K | 128x256x128 |   262,144 |       293.89 |      2.74 |  107.3x
+    16MB | 128x64x1000 |   512,000 |       616.80 |      5.99 |  102.9x
+    64MB | 256x64x2000 | 1,024,000 |      1154.30 |     10.70 |  107.9x
+
+Step breakdown (64MB workload):
+------------------------------------------------------------
+  Old path (7 steps):
+    1. Enumerate iteration space:         24.668 ms  (  2.1%)
+    2. Read index tensors:               779.872 ms  ( 66.8%)
+    3. Build coordinate list:            232.776 ms  ( 19.9%)
+    4. Linearize flat offsets:           128.898 ms  ( 11.0%)
+    5. Gather from HBM:                    1.187 ms  (  0.1%)
+    6. Reshape:                            0.003 ms  (  0.0%)
+    7. Write to LX:                        0.005 ms  (  0.0%)
+    TOTAL                              1167.409 ms
+
+  New path (3 steps):
+    1. Read K index values:                0.032 ms  (  0.8%)
+    2. Numpy broadcast offsets:            2.578 ms  ( 67.6%)
+    3. Gather + reshape + LX:              1.202 ms  ( 31.5%)
+    TOTAL                                 3.812 ms
+
+  Speedup: 306x
+```
+
+### `gather` — 3-memcpy vs 1-memcpy data movement
+
+```
+Workload | Gather (elems) | Span (elems) | Old (ms) | New (ms) | Speedup
+---------+----------------+--------------+----------+----------+--------
+ xl-262K |        262,144 |    3,211,264 |   0.5312 |   0.2526 |    2.1x
+    16MB |        512,000 |    7,744,000 |   1.3399 |   0.5175 |    2.6x
+    64MB |      1,024,000 |   28,544,000 |   1.5259 |   1.1510 |    1.3x
+
+  Old: _read_flat(span) + flat[offsets] + write_flat(lx) — 3 memcpys
+  New: mgr.gather(offsets) + place_in_lx — 1 memcpy
+```
+
+## Config
+
+`configs/indirect_emul.toml` — workloads specified as `(num_experts, M, N, n_selected)`. Size strings like `"2K"` are auto-parsed. List fields require explicit `mode = "product"` or `"zip"`.
+
+## Code Structure
+
+```
+bench_indirect_emul_time.py   — CLI entry point (block/gather subcommands)
+bench_utils.py                — parse_size, BenchTimer, BenchTable, load_config,
+                                make_bench_context, build_moe_iat
+configs/indirect_emul.toml    — workload parameter matrix
+```
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,11 +1,176 @@
-# Benchmarks: Indirect Access Emulation Timing
+# Benchmarks: Block-Gather Fast Path
+
+Measures the blocked-gather optimization against the general 7-step
+inspector-executor path for indirect memory access emulation.
+
+## Named Patterns
+
+The fast path applies to any indirect access with at least one indirect
+dimension (looked up via an index tensor) and at least one direct dimension
+(the contiguous "block"), where the ratio of total iteration points to unique
+index lookups exceeds 16×.
+
+| Pattern | Access expression | Indirect | Direct | Realistic workload |
+|---------|------------------|----------|--------|-------------------|
+| `moe_ffn` | `W[expert_idx[e], m, n]` | 1 | 2 | MoE feed-forward: select k expert weight matrices from a pool |
+| `paged_attn` | `cache[page_idx[p], heads, bs, hd]` | 1 | 3 | Paged KV-cache: gather selected pages for attention decode |
+| `sparse_attn` | `cache[page_idx[b], token_idx[t], d]` | 2 | 1 | Sparse attention: select page+token pairs from a large cache |
+| `multi_head` | `W[expert_idx[e], head_idx[h], m, n]` | 2 | 2 | Multi-head expert routing: select (expert, head) weight blocks |
+
+## Summary
+
+All patterns show 100–190× end-to-end speedup over the general path.
+The blocked-gather replaces 7 sequential steps with 3 steps whose cost is
+dominated by a single NumPy broadcast (offset computation) and a flat gather.
+
+| Pattern | Points | General (ms) | Fast (ms) | Speedup |
+|---------|--------|--------------|-----------|---------|
+| `moe_ffn` | 1,024,000 | 1,084 | 9.4 | 115× |
+| `paged_attn` | 1,048,576 | 1,176 | 10.5 | 112× |
+| `sparse_attn` | 128,000 | 227 | 1.2 | 190× |
+| `multi_head` | 4,096,000 | 7,745 | 40.5 | 191× |
+
+## Step Breakdown by Pattern
+
+### `moe_ffn` — MoE feed-forward weight selection
+
+1M points (256 experts × 64 × 2K, selecting 8).
+
+The general path spends 67% of its time resolving index tensor reads
+element-by-element. The fast path reads all 8 index values in one call,
+then computes offsets via a single broadcast multiply-add.
+
+**General path (7 steps):**
+
+| Step | Time (ms) | Share |
+|------|-----------|-------|
+| Enumerate iteration space | 23.7 | 2.1% |
+| Read index tensors | 765.5 | 67.1% |
+| Build coordinate list | 230.9 | 20.2% |
+| Linearize flat offsets | 120.2 | 10.5% |
+| Gather from HBM | 1.2 | 0.1% |
+| Reshape | 0.003 | — |
+| Write to LX | 0.004 | — |
+| **Total** | **1,141** | |
+
+**Fast path (3 steps):**
+
+| Step | Time (ms) | Share |
+|------|-----------|-------|
+| Read K index values | 0.023 | 0.7% |
+| NumPy broadcast offsets | 2.12 | 68.6% |
+| Gather + reshape + LX | 0.95 | 30.7% |
+| **Total** | **3.1** | |
+
+**Step-level speedup: 369×**
+
+---
+
+### `paged_attn` — Paged KV-cache page gather
+
+1M points (256 pages × 16 heads × 16 block_size × 128 head_dim, selecting 32 pages).
+
+Same bottleneck structure as MoE. The 3-direct-dim block (heads × bs × hd = 32K
+elements per page) gives high reuse per index lookup.
+
+**General path (7 steps):**
+
+| Step | Time (ms) | Share |
+|------|-----------|-------|
+| Enumerate iteration space | 23.5 | 2.0% |
+| Read index tensors | 758.1 | 63.5% |
+| Build coordinate list | 266.6 | 22.3% |
+| Linearize flat offsets | 143.8 | 12.1% |
+| Gather from HBM | 1.0 | 0.1% |
+| Reshape | 0.002 | — |
+| Write to LX | 0.003 | — |
+| **Total** | **1,193** | |
+
+**Fast path (3 steps):**
+
+| Step | Time (ms) | Share |
+|------|-----------|-------|
+| Read K index values | 0.038 | 1.2% |
+| NumPy broadcast offsets | 2.19 | 71.7% |
+| Gather + reshape + LX | 0.82 | 27.0% |
+| **Total** | **3.0** | |
+
+**Step-level speedup: 391×**
+
+---
+
+### `sparse_attn` — Sparse attention page+token selection
+
+128K points (128 pages × 64 tokens × 2K hidden, selecting 8 pages × 8 tokens).
+
+Two indirect dimensions multiply the general path's per-element index resolution
+cost. The fast path reads each index tensor once (8 + 8 = 16 values total), then
+broadcasts both into a combined offset grid.
+
+**General path (7 steps):**
+
+| Step | Time (ms) | Share |
+|------|-----------|-------|
+| Enumerate iteration space | 2.3 | 1.0% |
+| Read index tensors | 174.7 | 76.6% |
+| Build coordinate list | 36.2 | 15.9% |
+| Linearize flat offsets | 14.7 | 6.4% |
+| Gather from HBM | 0.08 | — |
+| Reshape | 0.002 | — |
+| Write to LX | 0.002 | — |
+| **Total** | **228** | |
+
+**Fast path (3 steps):**
+
+| Step | Time (ms) | Share |
+|------|-----------|-------|
+| Read K index values | 0.11 | 27.8% |
+| NumPy broadcast offsets | 0.21 | 52.9% |
+| Gather + reshape + LX | 0.08 | 19.3% |
+| **Total** | **0.41** | |
+
+**Step-level speedup: 563×**
+
+---
+
+### `multi_head` — Multi-head expert weight routing
+
+4M points (32 experts × 8 heads × 64 × 2K, selecting 8 experts × 4 heads).
+
+Largest workload: the general path takes nearly 8 seconds due to 5.8s spent
+resolving index reads across 4M points. The fast path reads 8 + 4 = 12 index
+values and computes the full 4M-element offset grid in under 5 ms.
+
+**General path (7 steps):**
+
+| Step | Time (ms) | Share |
+|------|-----------|-------|
+| Enumerate iteration space | 95.5 | 1.2% |
+| Read index tensors | 5,822 | 74.7% |
+| Build coordinate list | 1,328 | 17.0% |
+| Linearize flat offsets | 544.5 | 7.0% |
+| Gather from HBM | 3.9 | 0.1% |
+| Reshape | 0.003 | — |
+| Write to LX | 0.004 | — |
+| **Total** | **7,794** | |
+
+**Fast path (3 steps):**
+
+| Step | Time (ms) | Share |
+|------|-----------|-------|
+| Read K index values | 0.076 | 1.1% |
+| NumPy broadcast offsets | 4.54 | 63.2% |
+| Gather + reshape + LX | 2.57 | 35.8% |
+| **Total** | **7.2** | |
+
+**Step-level speedup: 1,084×**
 
 ## Usage
 
 ```bash
 cd ktir-cpu
 
-# Fast-path vs general-path comparison + per-step breakdown
+# Summary table + step breakdown (all patterns)
 uv run python benchmarks/bench_indirect_emul_time.py block
 
 # Gather memcpy optimization (3-copy vs 1-copy)
@@ -15,60 +180,17 @@ uv run python benchmarks/bench_indirect_emul_time.py gather
 uv run python benchmarks/bench_indirect_emul_time.py block --config configs/custom.toml
 ```
 
-## Output
-
-### `block` — fast-path vs general-path + per-step breakdown
-
-```
-Workload |       Shape |    Points | General (ms) | Fast (ms) | Speedup
----------+-------------+-----------+--------------+-----------+--------
- xl-262K | 128x256x128 |   262,144 |       293.89 |      2.74 |  107.3x
-    16MB | 128x64x1000 |   512,000 |       616.80 |      5.99 |  102.9x
-    64MB | 256x64x2000 | 1,024,000 |      1154.30 |     10.70 |  107.9x
-
-Step breakdown (64MB workload):
-------------------------------------------------------------
-  Old path (7 steps):
-    1. Enumerate iteration space:         24.668 ms  (  2.1%)
-    2. Read index tensors:               779.872 ms  ( 66.8%)
-    3. Build coordinate list:            232.776 ms  ( 19.9%)
-    4. Linearize flat offsets:           128.898 ms  ( 11.0%)
-    5. Gather from HBM:                    1.187 ms  (  0.1%)
-    6. Reshape:                            0.003 ms  (  0.0%)
-    7. Write to LX:                        0.005 ms  (  0.0%)
-    TOTAL                              1167.409 ms
-
-  New path (3 steps):
-    1. Read K index values:                0.032 ms  (  0.8%)
-    2. Numpy broadcast offsets:            2.578 ms  ( 67.6%)
-    3. Gather + reshape + LX:              1.202 ms  ( 31.5%)
-    TOTAL                                 3.812 ms
-
-  Speedup: 306x
-```
-
-### `gather` — 3-memcpy vs 1-memcpy data movement
-
-```
-Workload | Gather (elems) | Span (elems) | Old (ms) | New (ms) | Speedup
----------+----------------+--------------+----------+----------+--------
- xl-262K |        262,144 |    3,211,264 |   0.5312 |   0.2526 |    2.1x
-    16MB |        512,000 |    7,744,000 |   1.3399 |   0.5175 |    2.6x
-    64MB |      1,024,000 |   28,544,000 |   1.5259 |   1.1510 |    1.3x
-
-  Old: _read_flat(span) + flat[offsets] + write_flat(lx) — 3 memcpys
-  New: mgr.gather(offsets) + place_in_lx — 1 memcpy
-```
-
 ## Config
 
-`configs/indirect_emul.toml` — workloads specified as `(num_experts, M, N, n_selected)`. Size strings like `"2K"` are auto-parsed. List fields require explicit `mode = "product"` or `"zip"`.
+`configs/indirect_emul.toml` defines workloads with a `pattern` field that
+selects the IAT builder. Each entry specifies pattern-specific parameters
+(expert counts, page sizes, head dims, etc.). Size strings like `"2K"` are
+auto-parsed. List fields require explicit `mode = "product"` or `"zip"`.
 
 ## Code Structure
 
 ```
 bench_indirect_emul_time.py   — CLI entry point (block/gather subcommands)
-bench_utils.py                — parse_size, BenchTimer, BenchTable, load_config,
-                                make_bench_context, build_moe_iat
-configs/indirect_emul.toml    — workload parameter matrix
+bench_utils.py                — BenchTimer, BenchTable, IAT builders, config loader
+configs/indirect_emul.toml    — workload definitions (4 patterns × 2 sizes)
 ```
--- a/benchmarks/bench_indirect_emul_time.py
+++ b/benchmarks/bench_indirect_emul_time.py
@@ -1,0 +1,264 @@
+"""Indirect memory access emulation timing — MoE block-gather patterns.
+
+Subcommands:
+  block   — block-gather fast path vs general inspector (summary + breakdown)
+  gather  — 3-memcpy (old) vs 1-memcpy (new) gather step comparison
+
+Usage:
+    uv run python benchmarks/bench_indirect_emul_time.py block
+    uv run python benchmarks/bench_indirect_emul_time.py gather
+    uv run python benchmarks/bench_indirect_emul_time.py block --config configs/custom.toml
+"""
+
+import time
+import numpy as np
+
+from bench_utils import (
+    load_config, make_bench_context, build_moe_iat, reset_lx,
+    flush_cache, format_size, BenchTimer, BenchTable,
+)
+from ktir_cpu.ops.memory_ops import (
+    MemoryOps, _MemAccessor,
+    _is_block_gather, _block_gather_load,
+    _resolve_idx_reads, _build_indirect_coords,
+    _enumerate_in_vso_order,
+)
+from ktir_cpu.memory import _read_flat
+from ktir_cpu.dtypes import bytes_per_elem, to_np_dtype
+
+
+# ---------------------------------------------------------------------------
+# Shared helpers
+# ---------------------------------------------------------------------------
+
+def _run_general_path(ctx, iat):
+    reset_lx(ctx)
+    idx_values, _ = _resolve_idx_reads(ctx, iat)
+    coords = _build_indirect_coords(iat, idx_values)
+    MemoryOps.load(ctx, iat.parent_ref.to_tile_ref(), coords=coords, result_shape=iat.shape)
+
+
+def _run_fast_path(ctx, iat):
+    reset_lx(ctx)
+    _block_gather_load(ctx, iat)
+
+
+# ---------------------------------------------------------------------------
+# block subcommand
+# ---------------------------------------------------------------------------
+
+def _old_7_steps(ctx, iat) -> dict:
+    """One iteration of old 7-step path, returns per-step ms."""
+    reset_lx(ctx)
+    t0 = time.perf_counter()
+    _enumerate_in_vso_order(iat)
+    t1 = time.perf_counter()
+    idx_values, _ = _resolve_idx_reads(ctx, iat)
+    t2 = time.perf_counter()
+    coords = _build_indirect_coords(iat, idx_values)
+    t3 = time.perf_counter()
+    tile_ref = iat.parent_ref.to_tile_ref()
+    mgr = _MemAccessor(ctx, tile_ref.memref.memory_space, tile_ref.base_ptr, tile_ref.memref.lx_core_id)
+    offsets, _ = MemoryOps._flat_memory_offsets(
+        tile_ref.base_ptr, tile_ref.shape, tile_ref.strides, tile_ref.dtype,
+        coords, stick_bytes=mgr.stick_bytes,
+    )
+    t4 = time.perf_counter()
+    gathered = mgr.gather(offsets, tile_ref.dtype)
+    t5 = time.perf_counter()
+    data = gathered.reshape(iat.shape)
+    t6 = time.perf_counter()
+    MemoryOps._place_in_lx(ctx, data)
+    t7 = time.perf_counter()
+    return {
+        "1. Enumerate iteration space": (t1 - t0) * 1000,
+        "2. Read index tensors": (t2 - t1) * 1000,
+        "3. Build coordinate list": (t3 - t2) * 1000,
+        "4. Linearize flat offsets": (t4 - t3) * 1000,
+        "5. Gather from HBM": (t5 - t4) * 1000,
+        "6. Reshape": (t6 - t5) * 1000,
+        "7. Write to LX": (t7 - t6) * 1000,
+    }
+
+
+def _new_3_steps(ctx, iat) -> dict:
+    """One iteration of new 3-step path, returns per-step ms."""
+    from ktir_cpu.ops.memory_ops import _block_gather_analyze, _block_gather_read_idx, _block_gather_offsets
+
+    reset_lx(ctx)
+    info = _block_gather_analyze(iat)
+    indirect_sub, dep_vars, dep_var_list, dep_extents, dep_los = info
+
+    t0 = time.perf_counter()
+    idx_values_arr, _ = _block_gather_read_idx(ctx, iat, indirect_sub, dep_vars, dep_var_list)
+    t1 = time.perf_counter()
+    offsets = _block_gather_offsets(iat, idx_values_arr, dep_vars, dep_var_list, dep_extents, dep_los)
+    t2 = time.perf_counter()
+    tile_ref = iat.parent_ref.to_tile_ref()
+    mgr = _MemAccessor(ctx, tile_ref.memref.memory_space, tile_ref.base_ptr, tile_ref.memref.lx_core_id)
+    gathered = mgr.gather(offsets, tile_ref.dtype)
+    data = gathered.reshape(iat.shape)
+    MemoryOps._place_in_lx(ctx, data)
+    t3 = time.perf_counter()
+    return {
+        "1. Read K index values": (t1 - t0) * 1000,
+        "2. Numpy broadcast offsets": (t2 - t1) * 1000,
+        "3. Gather + reshape + LX": (t3 - t2) * 1000,
+    }
+
+
+def cmd_block(config):
+    """Fast-path vs general-path summary + per-step breakdown."""
+    print(f"{config.name}: block-gather fast path")
+    print()
+
+    # --- Summary table ---
+    table = BenchTable(
+        headers=["Workload", "Shape", "Points", "General (ms)", "Fast (ms)", "Speedup"],
+    )
+
+    for w in config.workloads:
+        ctx = make_bench_context()
+        iat = build_moe_iat(ctx, w["num_experts"], w["M"], w["N"], w["n_selected"], w.get("dtype", "f16"))
+        assert _is_block_gather(iat), f"Workload {w['label']} does not qualify for fast path"
+
+        timer = BenchTimer(
+            n_warmup=w.get("warmup", config.defaults.get("warmup", 2)),
+            n_rounds=w.get("n_rounds", config.defaults.get("n_rounds", 5)),
+            cache_flush=True,
+        )
+        general_ms, fast_ms = timer.measure_pair(
+            lambda: _run_general_path(ctx, iat),
+            lambda: _run_fast_path(ctx, iat),
+        )
+        total_pts = w["n_selected"] * w["M"] * w["N"]
+        speedup = general_ms / fast_ms if fast_ms > 0 else float("inf")
+        shape_str = f"{w['num_experts']}x{w['M']}x{w['N']}"
+        table.add_row([
+            w["label"], shape_str, f"{total_pts:,}",
+            f"{general_ms:.2f}", f"{fast_ms:.2f}", f"{speedup:.1f}x",
+        ])
+
+    table.print()
+
+    # --- Per-step breakdown on last (largest) workload ---
+    if config.modes.get("breakdown"):
+        w = config.workloads[-1]
+        print(f"Step breakdown ({w['label']} workload):")
+        print("-" * 60)
+
+        ctx = make_bench_context()
+        iat = build_moe_iat(ctx, w["num_experts"], w["M"], w["N"], w["n_selected"], w.get("dtype", "f16"))
+        timer = BenchTimer(
+            n_warmup=w.get("warmup", config.defaults.get("warmup", 2)),
+            n_rounds=w.get("n_rounds", config.defaults.get("n_rounds", 5)),
+        )
+
+        old = timer.measure_steps(lambda: _old_7_steps(ctx, iat))
+        new = timer.measure_steps(lambda: _new_3_steps(ctx, iat))
+
+        old_total = sum(old.values())
+        new_total = sum(new.values())
+
+        print("  Old path (7 steps):")
+        for step, ms in old.items():
+            print(f"    {step}:{' ' * (35 - len(step))}{ms:>8.3f} ms  ({ms/old_total*100:>5.1f}%)")
+        print(f"    {'TOTAL':{35}}{old_total:>8.3f} ms")
+        print()
+
+        print("  New path (3 steps):")
+        for step, ms in new.items():
+            print(f"    {step}:{' ' * (35 - len(step))}{ms:>8.3f} ms  ({ms/new_total*100:>5.1f}%)")
+        print(f"    {'TOTAL':{35}}{new_total:>8.3f} ms")
+        print()
+
+        print(f"  Speedup: {old_total / new_total:.0f}x")
+        print()
+
+
+# ---------------------------------------------------------------------------
+# gather subcommand
+# ---------------------------------------------------------------------------
+
+def cmd_gather(config):
+    """3-memcpy (old) vs 1-memcpy (new) gather step comparison."""
+    print(f"{config.name}: gather memcpy comparison (3-copy vs 1-copy)")
+    print()
+
+    table = BenchTable(
+        headers=["Workload", "Gather (elems)", "Span (elems)", "Old (ms)", "New (ms)", "Speedup"],
+    )
+
+    for w in config.workloads:
+        ctx = make_bench_context()
+        iat = build_moe_iat(ctx, w["num_experts"], w["M"], w["N"], w["n_selected"], w.get("dtype", "f16"))
+        dtype = w.get("dtype", "f16")
+
+        # Compute offsets once (use fast path to get them)
+        from ktir_cpu.ops.memory_ops import _block_gather_analyze, _block_gather_read_idx, _block_gather_offsets
+
+        info = _block_gather_analyze(iat)
+        indirect_sub, dep_vars, dep_var_list, dep_extents, dep_los = info
+        idx_values_arr, _ = _block_gather_read_idx(ctx, iat, indirect_sub, dep_vars, dep_var_list)
+        offsets = _block_gather_offsets(iat, idx_values_arr, dep_vars, dep_var_list, dep_extents, dep_los)
+
+        tile_ref = iat.parent_ref.to_tile_ref()
+        mgr = _MemAccessor(ctx, tile_ref.memref.memory_space, tile_ref.base_ptr, tile_ref.memref.lx_core_id)
+        span = int(offsets.max()) + 1
+        np_dtype = to_np_dtype(dtype)
+        elem_size = bytes_per_elem(dtype)
+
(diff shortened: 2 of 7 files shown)


yuhaohaoyu:
---

## 5. Benchmarks and Speedup

### `block` — fast-path vs general-path + per-step breakdown

```
Workload |       Shape |    Points | General (ms) | Fast (ms) | Speedup
---------+-------------+-----------+--------------+-----------+--------
 xl-262K | 128x256x128 |   262,144 |       293.89 |      2.74 |  107.3x
    16MB | 128x64x1000 |   512,000 |       616.80 |      5.99 |  102.9x
    64MB | 256x64x2000 | 1,024,000 |      1154.30 |     10.70 |  107.9x

Step breakdown (64MB workload):
------------------------------------------------------------
  Old path (7 steps):
    1. Enumerate iteration space:         24.668 ms  (  2.1%)
    2. Read index tensors:               779.872 ms  ( 66.8%)
    3. Build coordinate list:            232.776 ms  ( 19.9%)
    4. Linearize flat offsets:           128.898 ms  ( 11.0%)
    5. Gather from HBM:                    1.187 ms  (  0.1%)
    6. Reshape:                            0.003 ms  (  0.0%)
    7. Write to LX:                        0.005 ms  (  0.0%)
    TOTAL                              1167.409 ms

  New path (3 steps):
    1. Read K index values:                0.032 ms  (  0.8%)
    2. Numpy broadcast offsets:            2.578 ms  ( 67.6%)
    3. Gather + reshape + LX:              1.202 ms  ( 31.5%)
    TOTAL                                 3.812 ms

  Speedup: 306x
```

### `gather` — 3-memcpy vs 1-memcpy data movement

```
Workload | Gather (elems) | Span (elems) | Old (ms) | New (ms) | Speedup
---------+----------------+--------------+----------+----------+--------
 xl-262K |        262,144 |    3,211,264 |   0.5312 |   0.2526 |    2.1x
    16MB |        512,000 |    7,744,000 |   1.3399 |   0.5175 |    2.6x
    64MB |      1,024,000 |   28,544,000 |   1.5259 |   1.1510 |    1.3x

  Old: _read_flat(span) + flat[offsets] + write_flat(lx) — 3 memcpys
  New: mgr.gather(offsets) + place_in_lx — 1 memcpy
```

**Where the speedup comes from:**

Steps 2–3 of the old path (per-point Python iteration: `_resolve_idx_reads` + `_build_indirect_coords`) account for ~87% of total time. The fast path eliminates them entirely with O(ndim) numpy broadcast operations.

---

## 6. Reference: Why Not Generalize to #95 (Strided Sub-Rectangles)

**Question:** Could block-gather serve as the general mechanism for all algebraic blocked accesses (including #95's strided sub-rectangle case)?

**Answer:** No — the two optimizations target different bottlenecks.

| | Block-gather (this PR) | Strided sub-rectangle (#95) |
|---|---|---|
| **Path** | `indirect_load` / `indirect_store` | `MemoryOps.load` (direct access) |
| **Where time goes** | 99% in Python address computation | 99% in data movement |
| **What we eliminate** | O(N) Python loops → O(ndim) numpy broadcast | Nothing to eliminate — addresses are trivial |
| **Best tool** | meshgrid → offsets → gather | `as_strided` → zero-copy view → bulk copy |

Generalizing our approach to #95's domain would **pessimize** it:

- **Memory**: O(N) int64 offset array vs O(1) view descriptor for `as_strided`
- **Speed**: element-by-element fancy-index gather vs SIMD-vectorized strided copy
- **Store**: per-element scatter vs bulk strided assignment

The block-gather approach is optimal when **computing addresses is the bottleneck** (indirect: millions of Python-level index lookups). `as_strided` is optimal when **data movement is the bottleneck** (direct: addresses derivable from shape+strides). They're complementary rows in the dispatch table.

---

## 7. Related Issues

I found there has been heavy and scattered planning on the emulation speed worries. Thus did a search and let's have a consideration if want.

| Coverage | Issue | Relation |
|:--------:|-------|----------|
| **75%** | [#78](https://github.com/torch-spyre/ktir-cpu/issues/78) — Parallel GridExecutor | Primary motivation (indirect access bottleneck) largely removed for MoE workloads |
| **50%** | [#98](https://github.com/torch-spyre/ktir-cpu/issues/98) — Vectorized indirect coord build | Solved for block-gather subset; general high-cardinality case remains |
| **35%** | [#97](https://github.com/torch-spyre/ktir-cpu/issues/97) — ndarray coords in MemoryOps.load/store | Bypassed for dominant case (fast path skips coord construction entirely) |
| **20%** | [#67](https://github.com/torch-spyre/ktir-cpu/issues/67) — read_scattered cross-allocation safety | Avoided by new path (`gather()` fancy-indexes allocation directly) |
| **15%** | [#102](https://github.com/torch-spyre/ktir-cpu/issues/102) — HBM/LX memory bootstrap | Tangential — same classes modified (`gather()` methods added) |
| **10%** | [#96](https://github.com/torch-spyre/ktir-cpu/issues/96) — Preserve BoxSet under permuted VSO | Known gap: fast path requires identity VSO |
| **10%** | [#60](https://github.com/torch-spyre/ktir-cpu/issues/60) — indirect_store collision semantics | Fast path inherits same last-writer-wins behavior |
| **5%** | [#95](https://github.com/torch-spyre/ktir-cpu/issues/95) — Strided sub-rectangle fast path | Unrelated: targets direct access path (see §5) |
| **5%** | [#56](https://github.com/torch-spyre/ktir-cpu/issues/56) — Non-rectangular VSS | Not addressed: fast path requires BoxSet |

---

yuhaohaoyu:
@fabianlim functional code only 400 lines in memory.py and memory_ops.py

yuhaohaoyu:
The initial implementation has a constraint on that the fast path only treat for tensor accesses (load/store) with one indirection-dimension. With a simple revisit, the patch generalizes the block-gather fast path to multiple indirection dimensions

1. +35 net lines functional code change (`memory_ops.py:` 91 ins, 56 del).
1. Adds support for patterns with 2+ indirect dims: `sparse_attn` (page+token selection) and `multi_head` (expert+head routing), alongside existing `moe_ffn` and `paged_attn`.
2. Consolidated tests, benchmarks, and updated benchmarks/README with per-pattern step breakdowns showing **112–1084×** speedup.

lasch (commented):
Some (more code-level comments and suggestions). Algorithmic/Conceptual review I need some more digging...


## Non-blocking — suggested improvements

**1. `_place_in_lx` aliasing contract is violated but does not affect existing paths (`memory_ops.py:917–922`, `491–494`, `1255–1261`)**

All three `_place_in_lx` call sites produce a C-contiguous alias: `data`, `LX.memory[lx_ptr]`, and `Tile.data` all point to the same backing buffer. `linalg.matmul` and `linalg.batch_matmul` (`linalg_ops.py:326, 349`) both mutate `acc.data` in-place with `+=`, writing through the alias. However, the mutation is intentional (`inplace_outs=True`) and is a last write before HBM storage — the store reads `tile.data` directly and produces the correct accumulated value. No existing path observes the pre-accumulation LX state. Latent risk remains if a future kernel reuses `%C` after the matmul as a distinct value, or if a transitive alias via `arith.bitcast` / `tensor.reshape` reaches an in-place mutation.

Recommended follow-up: change `_place_in_lx` to store `data.ravel().copy()`. This closes all three call sites and transitive paths atomically at the cost of one extra allocation per load. Track as a follow-up issue rather than a merge blocker.

**2. `_block_gather_load` / `_block_gather_store` crash on `None` from `_block_gather_analyze` (`memory_ops.py`)**

Both functions call `_block_gather_analyze(iat)` and immediately unpack the result without guarding against `None`. Internal production paths are safe (guarded by `_is_block_gather` at the call sites), but a direct caller (e.g. a future benchmark) that skips the guard will get a `TypeError`. A guard at the top of each function would make them self-documenting and defensive.

**3. `_block_gather_store` read-modify-write over full span (`memory_ops.py:534–537`)**

The store path materializes a `offsets.max()+1` element prefix, scatters into it, and writes it back — up to ~33× read amplification for sparse MoE patterns. The load path uses a direct `gather` with no span copy. Add a matching `scatter(offsets, data, dtype)` to both simulators (mirror of `gather`) and use it here. Suggest tracking as a follow-up issue if not addressed in this PR.

**4. `sub_dep_vars` in `_block_gather_read_idx` is dead code**

Computed but never used — either remove it or implement the per-sub dep-var filtering optimization it was scaffolded for.

**5. `_block_gather_offsets` fallback gate fires mid-loop after partial computation**

The `direct_expr` / non-identity VSO checks should move into `_block_gather_analyze` so the gate is definitive before any broadcast work begins. Currently the caller cannot know which path ran.

## Minor
- Private symbol imports in `bench_indirect_emul_time.py` (`_block_gather_analyze`, etc.) will silently break if these are refactored. Consider a stable `block_gather_steps()` shim or documenting the coupling.
- `bench_indirect_emul_time.py:358`: add a message to the `assert _is_block_gather(iat)` — e.g. `f"Workload {w['label']!r} doesn't qualify: ratio too low or non-BoxSet VSS"`.


-----------------

There were a few test gaps detected. I made a PR to your fork adding those. Feel free to reject and make your own, here are the listed gaps:

## Test coverage gaps

### No test for non-identity `variables_space_order`

The fallback path in `_block_gather_offsets` (triggered by `vso is not None and not
vso.is_identity()`) had no test. Issue #96 notes that permuted-VSO block-gathers miss the fast
path today — a test was needed confirming graceful fallback (correct result) and that
`_is_block_gather` still returns `True` for these.

### No adversarial test for the 16× ratio boundary

The fallback path in `_block_gather_offsets` (triggered by `vso is not None and not
vso.is_identity()`) had no test. Issue #96 notes that permuted-VSO block-gathers miss the fast
path today — a test was needed confirming graceful fallback (correct result) and that
`_is_block_gather` still returns `True` for these.

### Dynamic attribute `result.index_unique_sticks = idx_sticks` on `Tile`

`_block_gather_load` sets this attribute after creating the `Tile`. If `Tile` uses `__slots__`
(check `ir_types.py`), this would raise `AttributeError` at runtime - however, the `__slots__`-issue is a non-issue. BUT: None of the tests exercise
`index_unique_sticks` on the returned tile, so this couldn't have surfaced even if it was an issue or becomes an issue later.

yuhaohaoyu:
Had to add force push after adding signoffs in an additional step. @lasch 
 I forgot to sign off while merging Lars' PR to my origin branch.

yuhaohaoyu:
@lasch @fabianlim 

The issues suggested by Lars had been all addressed, with one tiny foot on the snake. Top level breakdown is below: 

The issues brought up Lars has 5 "major", and 2 "minor" issues. They are referred as such in below, w mapping to the commit hash: 

1. Major 1, 3076238
2. Major 3, 06d0e67
3. refactor test_indirect_access (deduplication), 2bff253
4. the rest of the issues, major-2/4/5 w minor-1/2, bea4f9d

yuhaohaoyu:
@lasch On dealing with Major 1 issue, we did not do 'copy/backup', instead just used built-in guard for read-only vs writeable over the data. Further explained as below: 

# Major 1: `_place_in_lx` Aliasing Fix

## Problem

Three references alias one buffer after `_block_gather_load`:

```python
gathered = mgr.gather(offsets, tile_ref.dtype)
data = gathered.reshape(out_shape)
MemoryOps._place_in_lx(context, data)
result = Tile(data, tile_ref.dtype, out_shape, unique_sticks)
```

| Alias | Access Type |
|-------|-------------|
| `lx.memory[ptr]` | write-once, then read-only |
| `data` / `Tile.data` (input) | read-only |
| `Tile.data` (accumulator) | read-write (matmul `+=` only) |

Aliasing is correct (tile IS the LX scratchpad region). Risk: unguarded mutation silently corrupts LX state. `.copy()` would break the hardware model and hide bugs.

## Solution: freeze via `flags.writeable = False`

Zero-cost (O(1) flag flip). Violations raise `ValueError` at point of mutation.

**`memory_ops.py` — `_place_in_lx`:**
```python
context.lx.memory[lx_ptr] = data.ravel()
data.flags.writeable = False
```

**`linalg_ops.py` — `_accumulate_tile` helper:**
```python
def _accumulate_tile(acc, addend):
    acc.data.flags.writeable = True
    acc.data += addend.data
    acc.data.flags.writeable = False
```

Called from `linalg.matmul` and `linalg.batch_matmul`:
```python
if isinstance(acc, Tile):
    _accumulate_tile(acc, result)
    return acc
```

## Test refactoring

Added `test_loaded_tile_is_read_only`.
Extracted `_hbm_store(hbm, arr)` helper — replaces the 4-line alloc/write/ptr-math
pattern across 11 call sites. Added little comments.

yuhaohaoyu:
Major 3 is straightforward, with additional things on code deduplication, tests splashing into the general indirect-access handling for scatter is such a common pattern. Further as below. @lasch 

# Major 3: Replace read-modify-write with `scatter`

## Problem

All store paths read the full element span before writing back:

```python
span = int(offsets.max()) + 1
flat = mgr.read(span, dtype)       # read entire prefix
flat[offsets] = data                # scatter locally
mgr.write(flat)                    # write back entire prefix
```

For sparse patterns (e.g. MoE selecting rows 3 and 127 from a 128×256 tensor),
`span` is orders of magnitude larger than the actual write set. Read amplification
up to 64× on production workloads.

## Solution: `scatter` as the inverse of `gather`

```python
mgr.scatter(offsets, data, dtype)   # writes ONLY the target elements
```

No intermediate buffer, no read-before-write. Fancy-index assignment directly
into the allocation's backing array.

## Implementation

Two free functions in `memory.py` (shared logic, no duplication):

```python
def _gather_from(memory, byte_addr, offsets, dtype):
    alloc = _find_allocation(memory, byte_addr, bytes_per_elem(dtype))
    _, data, elem_offset = alloc
    return data.ravel()[elem_offset + offsets]

def _scatter_into(memory, byte_addr, offsets, data, dtype):
    alloc = _find_allocation(memory, byte_addr, bytes_per_elem(dtype))
    _, buf, elem_offset = alloc
    buf.ravel()[elem_offset + offsets] = data
```

Both `HBMSimulator` and `LXScratchpad` delegate to these — 2-line methods each.
`_MemAccessor.scatter` exposes the unified interface to callers.

## Reuse: 3 store paths consolidated

| Store path | Before | After |
|-----------|--------|-------|
| `_block_gather_store` (fast path) | read span + scatter + write | `mgr.scatter(offsets, data, dtype)` |
| `MemoryOps.store` (general strided/coord) | read span + scatter + write | `mgr.scatter(offsets, data, dtype)` |
| `MemoryOps.indirect_store` (general indirect) | read span + scatter + write | `mgr.scatter(offsets, data, dtype)` |

Net: −12 lines across `memory_ops.py`. All three paths now O(k) where k = elements written,
not O(max_offset).

## Verification

1208 tests pass. Existing store tests (`test_scatter_write`, `test_indirect_access`,
`test_ops`) exercise all three paths — no new tests needed for the substitution.
3 new unit tests verify `scatter` primitives directly (HBM sparse, LX sparse, roundtrip).

yuhaohaoyu:
On refactoring (de-dup) of tests_indirect_accesses.py. @lasch @fabianlim I spotted some dups in the general indirect access tsts when developing for the scattering pattern. So in this separate commit, 2bff253, did the dedup. Quite some code swapping around, thus this tabled explanation below 

# test_indirect_access.py refactoring summary

## MLIR template unification — before → after

| Before (2 copies each) | After (1 template) | Parameters |
|---|---|---|
| `SMALL_INDIRECT_MLIR` (const, 52 lines) | `_small_indirect_4x4_mlir(func_name, indirect_side)` | `"load"` → IAT on X |
| `SMALL_INDIRECT_SCATTER_MLIR` (const, 52 lines) | ↑ same function | `"store"` → IAT on Y |
| `_gather_vso_mlir_4x4(vso_str, func_name)` (54 lines) | `_indirect_vso_mlir_4x4(vso_str, func_name, indirect_side)` | `"load"` → IAT on X |
| `_scatter_vso_mlir_4x4(vso_str, func_name)` (54 lines) | ↑ same function | `"store"` → IAT on Y |
| `_SMALL_3D_INDIRECT_GATHER_3CYCLE_MLIR` (const, 46 lines) | `_small_3d_indirect_3cycle_mlir(func_name, indirect_side)` | `"load"` → IAT on X |
| `_SMALL_3D_INDIRECT_SCATTER_3CYCLE_MLIR` (const, 46 lines) | ↑ same function | `"store"` → IAT on Y |

All 6 MLIR sources produced identical module structure — 4 memory views, 2 access tiles (1 IAT + 1 direct AT), load→store. The only variation was which view received the IAT. Each unified template uses a `header` (f-string, shared views) + `tiles` (plain string, selected by `indirect_side`) + `footer` (plain string, shared load/store) — no nested f-string escaping.

## Other deduplication

| Extraction | Sites affected | Lines saved |
|---|---|---|
| `_prep_with_hbm(mlir, seed_fn)` / `_run_with_hbm(mlir, func_name, seed_fn)` | 10+ test functions | ~40 |
| `_resolve_ssa(node, val=32)` | 2 inline `resolve(node)` closures in `TestSubscriptExpr` | ~7 |
| `_seed_rfc_zeros(hbm)` | 2 RFC smoke tests | ~6 |
| `_seed_swap_4x4(hbm)` simplified (takes `hbm` not `interp`) | 1 site | ~3 |
| `_seed_3cycle(hbm)` extracted | 2 (load + store 3-cycle tests) | ~6 |

## Parametrize style change

Before: pass callable factory as pytest parameter (`mlir_factory`), reader must chase the reference.
After: pass `indirect_side` string (`"load"` / `"store"`), self-documenting at call site.

## Inline comment policy

All `hbm.write()` calls retain same-line comments describing what is being seeded.
Redundant info (dtype, stick number) dropped where already visible in the code arguments.

## Net result

- **Before:** 1030 lines
- **After:** 865 lines (−165 net, +324/−489 in diff)
- **Tests:** 21 passed, 0 changed semantics

ktir_cpu/dialects/linalg_ops.py
```diff
@@ -27,6 +27,13 @@
 from .registry import register, register_parser
 
 
+def _accumulate_tile(acc, addend):
+    """In-place accumulate into a frozen LX tile: unfreeze, add, re-freeze."""
+    acc.data.flags.writeable = True
+    acc.data += addend.data
+    acc.data.flags.writeable = False
```
lasch: Do we want to reflect the 'frozen/freeze'-aspect in the name of the function?
`_accumulate_tile` looks benign compared to the fact that it overrides a 'read-only' memory region.

lasch (commented):
Thanks @yuhaohaoyu for addressing the review items. One more high-level comment/suggestion below.


Additional suggestions for subsequent PRs:
 * PR scope: The introduction of the benchmark could have been done in a separate PR to reduce the amount of code changes and scope of this one.
 * (in your comments above) you may want to avoid using `#NN` to reference review items because they're being turned into issue/pr references pointing at random unrelated github issues

ktir_cpu/dialects/linalg_ops.py
lasch: @yuhaohaoyu any comments/responses/thoughts about this?

yuhaohaoyu: Sorry @lasch for the delay. 

The comment is misleading. I updated the comment. Essentially, with very expanded wording: 
1) the HW model of LX accesses, there is no frozen or not, aka guarding for read-only or non-mutable. 
2) the unfreeze/freeze is to use the python writeable feature guarding potential stumbling across aliased references. For sequential execution , this is as safe as atomic ops or lock guarded.

lasch:
@yuhaohaoyu unfortunately, this has conflicts to resolve now :hammer:

fabianlim:
@yuhaohaoyu 

Digging into why this PR conflicts with main: main merged #149 (67d6fb6), which reworked the same two outs-accumulation lines in linalg.matmul/linalg.batch_matmul (acc.data += result.data) to call a new _accumulate_inplace helper (copy-on-write for shared 0-LX literal constants).

This branch also touches those exact two lines, but for a different reason: commit 3076238 ("Freeze LX tile data after placement to enforce read-only invariant") + 49893ae add LX write-protection (writeable = False) after placement, then introduce _accumulate_tile to unfreeze/add/re-freeze around the in-place accumulation so matmul/batch_matmul keep working under the new freeze invariant.

That freeze invariant isn't motivated by anything in the block-gather/gather-scatter work as far as I can tell — it reads as a separate hardening change that happened to land on this branch. Two questions:

1. Is enforcing LX read-only-after-placement important on its own merits? If so, it'd be cleaner as its own PR — it'll need to be reconciled with #149's _accumulate_inplace fix regardless (whoever resolves the conflict needs to compose both: copy-on-write for shared literals and unfreeze/re-freeze around the mutation, not pick one).
2. If it's not essential to ship now, would it simplify review/merge to drop 3076238/49893ae from this PR and keep it focused on the fast path?


cc: @lasch

yuhaohaoyu:
@lasch mechanical conflicts are the least we expected. :) 

@fabianlim the conflict is minor, let's take your suggestion here and do a separate PR for the potential aliasing issue brought up by Lars' review. 

Let me give the comments a non-expansive spin.

yuhaohaoyu:
@lasch @fabianlim Woo, that's easy. :) 

remove the freeze-lx-tile commits: 3076238, 49893ae . had to force push the rest.

fabianlim:
cc: @yuhaohaoyu @lasch @nwang-ibm 

I did a quick investigation into the PR with my agents and the below report is produced. I didnt look through it that carefully, but upon reading it, the suggestions does make sense. I suggest to split off the benchmark into a different repo, the memory_ops.py changes here are very naunced, and I feel should be done with as much reuse as possible

# PR #147 — block-gather fast path: design analysis & reuse proposal

## 1. Current design (pre-PR baseline)

### Layering

```
ktir_cpu/memory.py                          ktir_cpu/ops/memory_ops.py
──────────────────                          ───────────────────────────
_find_allocation(memory, ptr, elem_size)     _MemAccessor
  ↑ single place: byte_addr → (array,          .read(n, dtype)
    elem_offset). Bounds-checked.              .read_scattered(addrs, dtype)
                                                .write(data)
_read_flat(memory, ptr, n, ...)                    │
  - calls _find_allocation                        │ wraps HBMSimulator /
  - zero-pads past allocation end                  │ LXScratchpad + the
                                                    │ stick/intra_byte split
_write_flat(memory, ptr, data)                     ▼
  - calls _find_allocation               MemoryOps.load / .store /
  - truncates past allocation end         .distributed_load / .distributed_store
                                                    │
HBMSimulator.read/write                            │ for indirect access:
LXScratchpad.read/write                            ▼
  - thin wrappers over _read_flat/         _resolve_idx_reads      (idx tensor reads)
    _write_flat                            _build_indirect_coords  (coord materialization,
                                                                     negative-index guard)
                                            indirect_load / indirect_store
```

Two invariants the pre-PR code states explicitly in comments:

1. **`_find_allocation` is "the single place where byte addresses are translated to array indices. All read/write helpers call this rather than doing their own address arithmetic."** `_read_flat`/`_write_flat` build on top of it and add bounds-checking (zero-pad on read, truncate on write) — so the *safety* layer is `_read_flat`/`_write_flat`, not `_find_allocation` alone.
2. **`_build_indirect_coords` is "shared by `indirect_load` and `indirect_store` so their coord construction stays in lockstep,"** and it's the one place that rejects negative indirect-index values (`raw_idx < 0: raise IndexError`), because "NumPy fancy-indexing silently wraps negatives."

For indirect (gather/scatter-shaped) access, the pre-PR flow for *any* IAT was:

```
_resolve_idx_reads(iat)          # read index tensors → idx_values dict
_build_indirect_coords(iat, idx_values)   # → coordinate tuples, negative-index checked
_flat_memory_offsets(...)        # coords → flat element offsets (+ unique_sticks)
span = offsets.max() + 1
flat = mgr.read(span, dtype)     # → _read_flat, bounds-checked/zero-padded
gathered = flat[offsets]         # fancy-index in Python-owned memory
mgr.write(flat) / MemoryOps._write_to_lx(...)
```

One general-purpose implementation. Every indirect load/store — regardless of shape — walked through `_resolve_idx_reads` → `_build_indirect_coords` → `_flat_memory_offsets`, so every safety property (negative-index rejection, VSO ordering, `direct_expr` support) lived in exactly one place.

## 2. What PR #147 adds, and why it doesn't reuse the above

The PR's stated goal: MoE/paged-attention patterns look up a *small* set of unique indices (e.g. 4 expert ids) that each fan out over a *large* contiguous block (e.g. 32 columns each). The general path pays for this with an O(total_points) Python loop in `_resolve_idx_reads`/`_build_indirect_coords`. The PR's fix is to detect that shape and vectorize it with numpy broadcasting.

That part is sound. What's not reused:

| New code | What it duplicates | Why reuse didn't happen |
|---|---|---|
| `_block_gather_analyze` | Nothing pre-existing — new gating heuristic (`unique_lookups * 16 <= total_points`, single indirect dim, BoxSet VSS) | N/A, this part is genuinely new logic |
| `_block_gather_read_idx` | `_resolve_idx_reads` — reads index tensors via `_MemAccessor.read_scattered` | Re-implemented instead of parameterizing `_resolve_idx_reads` to iterate over `dep_points` (the reduced set) instead of all `points` |
| `_block_gather_offsets` / `_block_gather_offsets_fallback` | `_build_indirect_coords` + `_flat_memory_offsets` — coordinate materialization and negative-index validation | Built as a **new numpy-broadcast coordinate/offset computation from scratch**, bypassing `_build_indirect_coords` entirely — which is *where the negative-index guard lives*. The guard was never ported. |
| `_block_gather_load` / `_block_gather_store` | `MemoryOps.load` / `.store`'s strided/coord-set branch | New top-level functions instead of a new *branch* inside the existing `load`/`store`, so `indirect_load`/`indirect_store` now fork into two independent implementations of "resolve idx → coords → offsets → move data" |
| `HBMSimulator.gather`/`.scatter`, `LXScratchpad.gather`/`.scatter`, `_gather_from`/`_scatter_into` | `_read_flat`/`_write_flat` — both call `_find_allocation`, but skip the bounds-check/zero-pad/truncate wrapper `_read_flat`/`_write_flat` add on top | New sibling functions next to `_read_flat`/`_write_flat` that call `_find_allocation` directly, instead of adding an `offsets=` parameter to `_read_flat`/`_write_flat` themselves |
| `MemoryOps._place_in_lx` | `MemoryOps._write_to_lx` → `LXScratchpad.write` → `_write_flat` | New method that writes `context.lx.memory[lx_ptr] = data.ravel()` directly, skipping `_write_flat`'s `_find_allocation` + patch logic, justified by a comment ("always a freshly bumped pointer") rather than a code-level guarantee |

The pattern across all six rows is the same: **rather than teaching the one general implementation to also handle the low-cardinality case fast, the PR built a second, parallel implementation that happens to share input/output shapes with the first.** The result (confirmed in review):

- The negative-index guard in `_build_indirect_coords` (`ktir_cpu/ops/memory_ops.py:677-681`) has no counterpart in `_block_gather_offsets`/`_block_gather_offsets_fallback` — a fast-path IAT with a negative index silently wraps instead of raising.
- `_read_flat`/`_write_flat`'s bounds-check/zero-pad has no counterpart in `_gather_from`/`_scatter_into` — an out-of-range offset crashes with a raw numpy `IndexError` instead of the documented zero-pad/truncate behavior.
- `_block_gather_load`'s empty-iteration-space branch forgot to set `Tile.index_unique_sticks` (present in the non-empty branch two paths below it), because that bookkeeping isn't shared code — it's copy-pasted per branch, per function.

This is exactly what "shared by X and Y so they stay in lockstep" (the comment on `_build_indirect_coords`) was written to prevent, and the new fast path sits entirely outside that lockstep.

## 3. Proposed changes to reuse instead of fork

The guiding idea: **the low-cardinality optimization is a data-dependent fast path for computing *offsets*, not a reason to reimplement the read/write/validation stack around it.** Every new function should either become a parameter/branch of an existing function, or should call the existing function it currently duplicates.

### 3.1 `_read_flat` / `_write_flat`: add an `offsets=` fast path, delete `_gather_from`/`_scatter_into`

Extend the two functions that already own the bounds-check/zero-pad contract, instead of adding siblings that skip it:

```python
def _read_flat(memory, ptr, n_elements, np_dtype, elem_size, *, offsets=None):
    alloc = _find_allocation(memory, ptr, elem_size)
    if alloc is None:
        raise ValueError(...)
    _, data, elem_offset = alloc
    flat = data.ravel()
    if offsets is not None:
        idx = elem_offset + offsets
        in_bounds = idx < flat.size            # same "zero past the end" semantics
        out = np.zeros(offsets.shape, dtype=np_dtype)
        out[in_bounds] = flat[idx[in_bounds]].astype(np_dtype)
        return out
    ...                                          # existing n_elements path unchanged
```

Same shape for `_write_flat` (truncate/no-op writes for out-of-bounds `idx`, matching the existing truncate-on-overrun behavior). `HBMSimulator`/`LXScratchpad` gain `gather(ptr, offsets, dtype)`/`scatter(ptr, offsets, data, dtype)` that just forward to `_read_flat(..., offsets=offsets)` / `_write_flat(..., offsets=offsets)`. `_MemAccessor.gather`/`.scatter` stay as-is (they already just forward to the simulator).

Net effect: one bounds-check implementation, reused by scalar and gathered access alike. `_gather_from`/`_scatter_into` are deleted.

### 3.2 `_place_in_lx`: fold into `LXScratchpad.write`, delete the second entrypoint

Rather than a caller-selected fast path, make the *fresh-pointer* case an internal optimization of the one write path:

```python
def _write_flat(memory, ptr, data):
    if ptr not in memory and not _find_allocation(memory, ptr, data.itemsize):
        memory[ptr] = data.ravel() if data.flags['C_CONTIGUOUS'] else data.flatten()
        return
    ... # existing patch-in-place logic
```

This already exists in spirit — `_write_flat`'s `alloc is None` branch does `memory[ptr] = data.flatten()`. The PR's fast path is really just "use `.ravel()` instead of `.flatten()` when the caller can prove `data` isn't aliased." That's safe to do **inside** `_write_flat`'s existing no-existing-allocation branch (same condition it already checks), not as a separate `_place_in_lx` that callers must remember to pick. Every current call site (`_write_to_lx`, and the load paths this PR redirected to `_place_in_lx`) then gets the optimization automatically and `_place_in_lx` disappears.

### 3.3 `_block_gather_read_idx`: parameterize `_resolve_idx_reads` instead of duplicating it

`_resolve_idx_reads(iat)` builds `per_view_addrs` by iterating `points = _enumerate_in_vso_order(iat)`. Give it an optional `points` override:

```python
def _resolve_idx_reads(iat, points=None):
    points = points if points is not None else _enumerate_in_vso_order(iat)
    ...  # unchanged body
```

`_block_gather_read_idx` becomes a thin caller: compute `dep_points` (the reduced cardinality set), call `_resolve_idx_reads(iat, points=dep_points)`, and reshape. This removes the second hand-rolled `read_scattered`-driving loop entirely and guarantees both paths read index tensors identically (same accessor batching, same stick counting).

### 3.4 `_block_gather_offsets` / `_block_gather_offsets_fallback`: route through `_build_indirect_coords`'s validation, don't bypass it

This is the fix for the negative-index gap, and it's also the biggest simplification. `_build_indirect_coords` already knows how to turn `(iat, idx_values)` into coordinate tuples, validated. The fast path's actual novelty is *only* that it doesn't need to materialize `idx_values` per-point — it can broadcast the *small* `idx_values_map` result up to the dep-var grid shape first. Validation should still be shared:

Factor the negative-index check out of `_build_indirect_coords` into a tiny shared helper —

```python
def _check_nonnegative_idx(raw_idx, index_view):
    if raw_idx < 0:
        raise IndexError(f"indirect index {raw_idx} from {index_view} is negative")
```

— call it from `_build_indirect_coords` (per scalar `raw_idx`, unchanged behavior) AND from `_block_gather_read_idx`/`_block_gather_offsets` right after the raw index values are read back (vectorized: `if (idx_values_arr < 0).any(): raise IndexError(...)`, before any broadcasting happens). This guarantees one place defines "negative indirect index is an error," and both the scalar and vectorized paths enforce it — a future change to that policy (e.g. a configurable clamp mode) only needs to happen once.

The stride/offset math itself (`_block_gather_offsets`'s broadcast grid construction) is genuinely new and worth keeping — it's not a duplicate of anything, it's the actual performance work. Only the validation and the read-idx plumbing should be shared.

### 3.5 `_block_gather_load` / `_block_gather_store`: share the boilerplate, fix the empty-tile asymmetry as a side effect

Both functions repeat: compute `n_points`, early-return on zero, call `_block_gather_read_idx`, call `_block_gather_offsets`, build `_MemAccessor` + `tile_ref`, compute `unique_sticks` via `np.unique(...)`. Factor into one helper:

```python
def _block_gather_prepare(context, iat):
    """Returns (mgr, tile_ref, offsets, unique_sticks, idx_sticks), or
    (None, tile_ref, empty_offsets, 0, 0) for a zero-extent IAT."""
    info = _block_gather_analyze(iat)
    if info is None:
        raise ValueError("IAT does not qualify for block-gather fast path")
    ...
    tile_ref = iat.parent_ref.to_tile_ref()
    if n_points == 0:
        return None, tile_ref, np.empty(0, dtype=np.int64), 0, 0
    idx_values_map, idx_sticks = _block_gather_read_idx(...)   # now validates (3.4)
    offsets = _block_gather_offsets(...)
    mgr = _MemAccessor(...)
    unique_sticks = _MemAccessor.count_sticks(...)   # reuse, see 3.6
    return mgr, tile_ref, offsets, unique_sticks, idx_sticks
```

`_block_gather_load`/`_block_gather_store` become ~8 lines each. This also fixes the `index_unique_sticks=None` bug for free: the zero-extent branch returns `idx_sticks=0` through the same variable both call sites already assign to `result.index_unique_sticks`, so there's no longer a code path that forgets to set it — there's only one place that decides the value, for both empty and non-empty cases.

### 3.6 Stick counting: call `_MemAccessor.count_sticks`, don't re-derive the formula

`_MemAccessor.count_sticks(memory_space, byte_addresses)` already exists as "single source of truth for stick counting." The fast path currently inlines `np.unique((tile_ref.base_ptr + offsets * bpe) // stick_bytes).size` twice instead of calling it. `count_sticks` takes an `Iterable[int]` of byte addresses — either give it a vectorized sibling (`count_sticks_vectorized(memory_space, base_ptr, offsets, bpe)` that does the same `// STICK_BYTES` dedup via numpy instead of a Python set) or just call `count_sticks(memory_space, (tile_ref.base_ptr + offsets * bpe).tolist())` if the list-conversion cost is acceptable at this cardinality. Either way, one formula, one place to fix if stick-boundary math ever changes (e.g. multi-byte elements spanning a stick edge).

### 3.7 `_is_block_gather` / `_block_gather_analyze` double-call: cache, don't recompute

`indirect_load`/`indirect_store` call `_is_block_gather(iat)` (which internally calls `_block_gather_analyze(iat)` and discards the result) purely to decide whether to dispatch; `_block_gather_load`/`_store` then call `_block_gather_analyze(iat)` again from scratch. Simplest fix: have `indirect_load`/`indirect_store` call `_block_gather_analyze(iat)` once and pass the `info` tuple (or `None`) into `_block_gather_load`/`_store`, deleting `_is_block_gather` as a separate entry point:

```python
info = _block_gather_analyze(iat)
if info is not None:
    return _block_gather_load(context, iat, info, result_shape)
```

## 4. Net effect of the reuse pass

- `_gather_from`, `_scatter_into`, `_place_in_lx`, and the double `_block_gather_analyze` call disappear.
- `_block_gather_read_idx` shrinks to a thin wrapper over `_resolve_idx_reads`.
- The negative-index guard becomes one function called from both the scalar and vectorized coordinate paths — closing the correctness gap found in review, not by adding a parallel check, but by removing the fork that let it go missing.
- `_block_gather_load`/`_block_gather_store` shrink from ~45 lines combined boilerplate to ~15, with the empty-tile bug structurally impossible rather than fixed by remembering an extra line.
- The only code that remains genuinely new (and should remain new) is the broadcast/meshgrid offset math in `_block_gather_offsets` — the part that actually delivers the claimed speedup.

yuhaohaoyu:
@fabianlim I will give it a spin when with cycle. meanwhile, if you can give it a deeper look through. See if some comments are communicable ? Agent is great at diverging, we love to converge. :) Probably no need. Overall it's a refactoring/de-dup check.

yuhaohaoyu:
@fabianlim 

# PR #147 Review Feedback: Categorization & Alignment

## Lars' review (tactical — things wrong or dangerous NOW)

| # | Issue | Status |
|---|---|---|
| Major 1 | `_place_in_lx` aliasing | Deferred (separate PR) |
| Major 2 | None guard on load/store | Fixed |
| Major 3 | Read-modify-write in store (scatter) | Fixed |
| Major 4 | Dead `sub_dep_vars` | Fixed |
| Major 5 | Fallback gate mid-loop | Fixed |
| Minor 1 | Private imports coupling | Fixed (comment) |
| Minor 2 | Assert message | Fixed |

## Fabianlim's agent (strategic — forked implementation causes invariant drift)

| # | Issue | Aligns with Lars? |
|---|---|---|
| 3.1 | `_gather_from`/`_scatter_into` bypass `_read_flat`/`_write_flat` bounds-check | Extends Lars Major 3 — we added scatter, agent says it should live inside `_read_flat` |
| 3.2 | `_place_in_lx` bypasses `_write_flat` safety | = Lars Major 1, different angle |
| 3.3 | `_block_gather_read_idx` duplicates `_resolve_idx_reads` | New observation |
| 3.4 | **Negative-index guard missing** | New — Lars didn't catch explicitly |
| 3.5 | Load/store boilerplate + empty-tile asymmetry | Loosely related to Lars Major 2 |
| 3.6 | Stick counting inlined, not calling `count_sticks` | New observation |
| 3.7 | Double-call of `_block_gather_analyze` | New observation |

## Categorization

### Major (correctness — something is wrong)

- **Negative-index guard missing** (3.4) — only real bug. A fast-path IAT with negative index silently wraps instead of raising. One line fix: `if (idx_values_arr < 0).any(): raise IndexError(...)`.

### Minor (code quality — nothing is wrong today, but drift risk)

- Everything else (3.1–3.3, 3.5–3.7). They're all variations of: "you built a parallel path instead of parameterizing the existing one, so safety invariants now have two maintenance sites."

## The cloud

Lars saw *symptoms* (dead code, missing guard, wrong dispatch point). Fabianlim's agent sees the *structural cause*: the fast path is a **fork** of the general path, not a **specialization** within it. The fork is why Lars' symptoms appeared — when you maintain two parallel implementations, invariants slip through.

But: **the fork exists for a reason**. The general path is O(total_points) Python loops. The fast path is numpy broadcast. These are fundamentally different computational strategies — you can't just "parameterize" a Python loop into numpy broadcasting. The agent's suggestion to route through `_resolve_idx_reads` and `_build_indirect_coords` would reintroduce the per-point Python loops that the fast path exists to eliminate.

**What's actionable now:** Add the negative-index check (one line, real bug). The rest is future refactoring that should be weighed against the performance motivation — sharing validation doesn't require sharing computation.

yuhaohaoyu:
@fabianlim the design decision not to factor the algorithms for 'blocked-indirection' treatment into the 7 steps of 'element-wise indirection' was by design, not to tangle too much with the existing bottlenecked path when introducing new and complex algorithmic solution path. So we chose 'front-gated separate path'. 

After this PR merges, we will have a close-formed task to refactor, de-dup. 

It's fairly common in implementation of complex algorithms: lift the implementation then refactor.

fabianlim:
Ok thanks for the feedback. Maybe hold up the code work for abit and let me
take a look at your write up

fabianlim:
@yuhaohaoyu @lasch Ran an experiment to make the "reuse vs. fork" idea concrete: [compare](https://github.com/yuhaohaoyu/ktir-cpu/compare/indirect_access_memcpy_3_to_1..fabianlim:ktir-cpu:block-gather-reuse-validation?expand=1)

**Both paths do the same 3 stages: read index tensors → build coordinates → gather/scatter.** The only real difference between "general" and "fast" is *how many points* get enumerated in stage 1 — everything after that is the same code today (offset math), it just got duplicated instead of parameterized.

```
                    GENERAL PATH                         FAST PATH (dep-indices only)
stage 1 (read):     _resolve_idx_reads(iat)               _resolve_idx_reads(iat, points=dep_points)
                     loops over ALL points                 loops over just the dep-var subset (small)
                     ↓                                     ↓
stage 2 (validate    _validate_nonneg_idx(...)             _validate_nonneg_idx(...)   ← SAME call
+ coords):           _build_indirect_coords(...)           _block_gather_broadcast_coords(...)  ← NEW,
                     per-point Python loop                  numpy broadcast, no Python loop
                     ↓                                     ↓
stage 3 (move data): MemoryOps.load(coords=...)            MemoryOps.load(coords=...)  ← SAME call
                     → _flat_memory_offsets + mgr.gather   → _flat_memory_offsets + mgr.gather
```

So: **one new function** (`_block_gather_broadcast_coords` — the meshgrid/broadcast math, the actual perf win) plus **one new parameter** (`_resolve_idx_reads(..., points=)`) plus **one extracted helper** (`_validate_nonneg_idx`, called by both sides — this is what closes the negative-index gap). `_block_gather_analyze` (gating) stays as-is.

**What that lets us delete**, since neither path ever needs to talk to the simulators directly anymore:
- `_block_gather_read_idx`, `_block_gather_offsets`, `_block_gather_offsets_fallback` — replaced by reusing `_resolve_idx_reads`/`_build_indirect_coords` with a smaller point set
- `HBMSimulator.gather`/`.scatter`, `LXScratchpad.gather`/`.scatter`, `_gather_from`, `_scatter_into` — `memory.py` needs zero changes, both paths already terminate at `MemoryOps.load`/`.store`
- `_place_in_lx` — `_write_to_lx` already covers it

Validated against your test suite unmodified: 46/49 `test_block_gather_fast_path.py` pass (3 failures call the now-deleted `gather`/`scatter` directly — expected), 21/21 `test_indirect_access.py` pass, full suite clean. Confirmed the negative-index fix concretely: negative value in an index tensor on a fast-path-qualifying IAT now raises `IndexError`, same as the general path.

**NOTE: I didnt check the benchmarks. I could have missed the whole point of the optimization, so need to check this thinking**

**NOTE 2:  I did not move the helpers from their original position (to make it easy to compare), but would be good to move the gather helpers to a different utils file to keep memory_ops clean**

yuhaohaoyu:
1. PR rebased, it'd be mergeable.

2. assigned issue #171 to myself and rationalized the understanding the implication of not to tackle the LX-aliasing now. 

@lasch @fabianlim @WarningRan @nwang-ibm

PR #149, opened 2026-07-03
fix: materialize constant outs in inplace_outs ops (straggling #140 case)

fabianlim:
## Summary

Fixes a case missed by #140: `arith.constant` used **directly** as the `outs` operand of `linalg.matmul` / `linalg.batch_matmul` inside a loop body — not via `scf.for iter_arg`.

PR #140 fixed the iter_arg path by copying constant inits in `_materialize_iter_inits`. But when a constant is passed as `outs` directly (no enclosing `scf.for`), that copy never happens. The in-place `acc.data += result` corrupts the shared literal on the first iteration; subsequent iterations start from the dirty value instead of zero.

## Fix

`_accumulate_inplace` helper (mirrors `_materialize_iter_inits`): checks `context._uncharged_tiles` before mutating, copies the literal if needed. Applied to both `linalg.matmul` and `linalg.batch_matmul`.

## Test plan

- `TestConstantDirectOutsMutation::test_constant_outs_not_corrupted_across_loop_iterations` — regression for the direct-outs pattern (4 iterations, each must produce 2.0, not 2/4/6/8)
- `TestKtdp::test_load_respects_non_contiguous_middle_stride*` — confirms strided `ktdp.load` is already correct (valid finding from the debugging session)
- Existing `TestConstantIterArgMutation` suite — all pass, no regression on the iter_arg path

### DIFF
--- a/ktir_cpu/dialects/linalg_ops.py
+++ b/ktir_cpu/dialects/linalg_ops.py
@@ -297,6 +297,23 @@ def linalg__broadcast(op, context, env):
     return Tile(result, inp.dtype, out_shape)
 
 
+def _accumulate_inplace(acc: "Tile", result: "Tile", context) -> "Tile":
+    """Accumulate *result* into *acc* in place and return *acc*.
+
+    Mirrors the logic of ``_materialize_iter_inits`` in ``scf_ops.py``:
+    if *acc* is a 0-LX literal (an ``arith.constant`` bound with
+    ``charge=False``), copy it into a fresh working buffer first so the
+    shared literal is never mutated.  Without this guard, a constant used
+    as the ``outs`` operand of ``linalg.matmul`` inside a loop is corrupted
+    on the first iteration; every subsequent iteration starts from the dirty
+    value instead of the intended initial value (typically zero).
+    """
+    if id(acc) in context._uncharged_tiles:
+        acc = acc.copy()
+    acc.data += result.data
+    return acc
+
+
 @register("linalg.matmul", latency_category=LC.COMPUTE_MATMUL, inplace_outs=True)
 def linalg__matmul(op, context, env):
     """Execute linalg.matmul: result = outs + ins[0] @ ins[1].
@@ -316,15 +333,10 @@ def linalg__matmul(op, context, env):
     tile_a = context.get_value(op.operands[0])  # ins[0] = A
     tile_b = context.get_value(op.operands[1])  # ins[1] = B
     result = ArithOps.matmul(tile_a, tile_b)    # A @ B
-    # Accumulate into outs (operands[2] = C) when present.
-    # In-place update: mirrors hardware behavior where the accumulator is a
-    # physical buffer written in-place. Returning the same Tile object means
-    # alias_dedup sees the same id() and doesn't double-charge LX.
     if len(op.operands) > 2:
         acc = context.get_value(op.operands[2])
         if isinstance(acc, Tile):
-            acc.data += result.data
-            return acc
+            return _accumulate_inplace(acc, result, context)
     return result
 
 
@@ -346,8 +358,7 @@ def linalg__batch_matmul(op, context, env):
     if len(op.operands) > 2:
         acc = context.get_value(op.operands[2])
         if isinstance(acc, Tile):
-            acc.data += result.data
-            return acc
+            return _accumulate_inplace(acc, result, context)
     return result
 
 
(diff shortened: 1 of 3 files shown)


ohadeytan:
Tested PR #149 head (`74fe4b8b`) against the case in #157 (a hoisted `arith.constant dense<0.0>` used directly as `linalg.matmul outs` inside an `scf.for`). **The copy-on-use fix is correct, but it currently trips the structural assertion added in #136, so it fails hard instead of producing the right answer when run through the interpreter.**

`_accumulate_inplace` copies the uncharged literal and returns the **copy** (a new `Tile`, new `id()`). But `interpreter.py` (~line 236) still asserts the handler returned the *same* object as `outs`:

```
assert id(result) == id(outs_tile), (
    f"{op.op_type}: handler returned new Tile instead of mutating outs {outs_name!r}")
```

`outs_tile` is re-fetched as the original `%cst`, so `id(copy) != id(%cst)` → `AssertionError: linalg.matmul: handler returned new Tile instead of mutating outs '%z'` on the first matmul.

The reason #149's own unit test (`TestConstantDirectOutsMutation`) passes is that it calls `dispatch("linalg.matmul")(op, ctx, None)` **directly**, bypassing `_execute_op` where the assertion lives — so the assertion never runs in that test.

### Reproduction through the real interpreter

Minimal KTIR + driver in #157. Results on `74fe4b8b`:

- **Assertions on** (normal `python`): `AssertionError` at `interpreter.py:236` on the first `linalg.matmul` — nothing executes.
- **Assertions off** (`python -O`, strips `assert`): **correct** — `Out[0] = A[0]@B`, `Out[1] = A[1]@B`. Confirms the copy fix itself is right; only the assertion blocks it.

On our real flash-attention KTIR the same assertion fires immediately, so all tests error out (including cases that previously passed).

### Suggestion

The structural assertion needs to allow the copied-literal case. Either:
- relax it to accept a returned copy when the `outs` was an uncharged 0-LX literal (i.e. exactly when `_accumulate_inplace` chose to copy), or
- have `_accumulate_inplace` rebind the copy under the `outs` SSA name so the assertion's re-fetch sees the same object.

Happy to test a follow-up commit.

fabianlim:
@ohadeytan thanks for catching this — the copy-on-use fix was tripping the structural assertion from #136 exactly as you diagnosed. Fixed in the latest commit: the assertion now skips the identity check when `outs` was an uncharged (0-LX literal) tile, since that's precisely the case where `_accumulate_inplace` returns a copy by design. Added a regression test that exercises this through `_execute_op` (not just direct handler dispatch) so the assertion path is covered too.

ohadeytan:
Thanks @fabianlim, that fixed it!

fabianlim:
Added regression tests closing the coverage gaps flagged in review:

- **f16 variant** of the direct-outs corruption test — matches the dtype of the real repros (spyre_chained_scratchpad, #157).
- **`linalg.batch_matmul` direct-outs test** — previously only covered via the iter_arg path; this hits `_accumulate_inplace`'s batch_matmul branch directly.
- **LX-accounting assertions** for `_accumulate_inplace` — copy is distinct/chargeable, original `%cst` binding and `_uncharged_tiles` membership stay untouched (mirrors the existing iter_arg accounting tests).
- **Shared-literal multi-use** — two matmuls consuming the same `%cst` in one scope get independent copies; neither touches the shared literal.

Full suite: 1064 passed, 1 skipped, 13 xfailed — no regressions.

fabianlim:
@lasch sorry the review got dismissed, would appreciate another approval

ktir_cpu/dialects/linalg_ops.py
```diff
@@ -297,6 +297,23 @@ def linalg__broadcast(op, context, env):
...
+    shared literal is never mutated.  Without this guard, a constant used
+    as the ``outs`` operand of ``linalg.matmul`` inside a loop is corrupted
+    on the first iteration; every subsequent iteration starts from the dirty
+    value instead of the intended initial value (typically zero).
+    """
+    if id(acc) in context._uncharged_tiles:
+        acc = acc.copy()
+    acc.data += result.data
+    return acc
+
```
yuhaohaoyu, line 315: executor/interpreter side code, it's 'running time', using python runtime func id() make sense.

fabianlim: yes the use of `id` is to detect different python instances during runtime.

fabianlim:
@lasch 

> The follow-up test_constant_outs_via_execute_op... closes the assertion path, but still hand-builds Operation objects. 

Yes in quite alot of the tests we dont introduce a mlir fixture file, but we just construct the operation objects to exhibit a certain behavior. There is a trade off, having alot of mlir fixtures that are negative examples can be quite confusing. Currently we just use the mlir fixtures as a bare-minimum set of e2e tests. So i feel until we change that stance, its ok to not add a new mlir fixture. However its true that it let through that bug. So maybe we need to consider what stance we want and the trade offs

@lasch @yuhaohaoyu it seems you guys are ok with it, so maybe an approve to merge?

PR #152, opened 2026-07-07
[latency] Surface cross-core communication view in latency_demo.ipynb

### LINKED ISSUE #126: latency analysis: surface cross-core communication (comm) cost

WarningRan:
Spyre's chip has many cores; some kernels need cores to exchange and combine partial results over the on-chip ring interconnect (a "cross-core reduce"). That exchange costs time, but the latency tool could never show it. This change surfaces it.

**What.** Add a cross-core **communication view** to `latency_demo.ipynb` — a per-core compute/memory/comm breakdown plus, on the critical (wall-clock) core, `comm_fraction` / `comm_time` / `comm_bytes`.

**Why.** The latency model already costs communication (folded into `total_cycles`), but the notebook never showed it — so *"how much latency does a cross-core reduce add, and is comm ever the bottleneck?"* was unanswerable.

Driven on the in-repo `examples/ktir/ring_reduce*.mlir` all-reduce kernels, the comm segment now shows as a distinct, non-zero slice (e.g. `ring_reduce`: comm = 0.85% of the critical core, memory-bound). **Rendered charts + tables in the comment below.**

**Before → after**
- Before: the `comm` column prints integer cycles (sub-cycle comm rounds to `0`), and no communicating kernel is even loaded — comm is effectively invisible.
- After: a stacked-bar breakdown + critical-core comm metrics, computed in-cell from the existing `LatencyReport`.

**Metrics** (all on the critical-path core = `max(total_cycles)`)

| metric | formula |
|---|---|
| `comm_fraction` | `comm_cycles / total_cycles` |
| `comm_time` | `comm_cycles / (clock_ghz · 1e3)` (µs) |
| temporal breakdown | `{compute, memory, comm}_cycles / total_cycles` |
| `comm_bytes` | per-core ring bytes (Σ over that core's hops) |

`total_cycles = compute + memory + comm`; `comm_cycles = comm_bytes / ring_bytes_per_cycle` (already modelled).

**Scope.** Notebook-only; no `ktir_cpu` change; surfacing only. Part of #126 (report-API promotion and a composite cross-core kernel are follow-ups).

### DIFF
--- a/notebooks/latency_demo.ipynb
+++ b/notebooks/latency_demo.ipynb
@@ -662,6 +662,32 @@
   },
   {
    "cell_type": "markdown",
+   "source": "## 7. Cross-core communication (comm) view\n\nEvery kernel above is embarrassingly parallel, so the `comm` column reads `0`. Communication *is* costed by\nthe model (`ktdp.inter_tile_reduce` → `comm_cycles`, folded into `total_cycles`; `bottleneck` recognises\n`comm`) — it just isn't exercised or surfaced here.\n\nThis section loads the in-repo cross-core all-reduce examples (`examples/ktir/ring_reduce*.mlir`) and surfaces\ncommunication: a per-core `compute / memory / comm` breakdown, plus the **critical-path core**'s\n`comm_fraction`, `comm_time`, and `comm_bytes`. Metrics are taken on the critical core (`max(total_cycles)`)\n— the wall-clock core `kernel_time` comes from — so this is *exposed communication* (PyTorch **HTA** framing):\ncomm on a non-critical core is hidden behind that core's own work and adds no wall time.",
+   "metadata": {}
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


WarningRan:
CC maintainers @fabianlim @kiszk @lchu6 @nwang-ibm and @yuhaohaoyu for review.

WarningRan:
The PR adds one **"Cross-core communication view"** section to `latency_demo.ipynb`. **Notebook outputs are
cleared on commit** — the figures/numbers below are what its cells produce when run (default
`HardwareConfig()`), reproduced here so reviewers see the results without running the notebook.

**Cells added (which block produces what):**

- **`[md]` intro** — comm is already costed but was invisible; metrics are on the critical (wall-clock) core
  (exposed communication, HTA).
- **`[code]` comm helpers** — `run_comm_kernel(path, func, **kw)` (load + seed HBM + run → report),
  `comm_metrics(report)` (critical-core comm_fraction / comm_time / temporal / comm_bytes),
  `plot_comm_breakdown(report)` (per-core stacked bar), `comm_table(report)` (per-core table).
- **`[code]` ring_reduce cell** — loads `examples/ktir/ring_reduce.mlir`, runs → metrics + table + bar.
- **`[code]` ring_reduce_inner_loop cell** — loads `examples/ktir/ring_reduce_inner_loop.mlir` (`n_iters=3`),
  runs → metrics + table + bar.

Both `.mlir` are already in the repo — this PR only loads them; no kernel files added.

### `ring_reduce` — single-ring all-reduce (one top-level comm) · output of the **ring_reduce cell**
issue126-comm-ring_reduce

critical core = 0 · bottleneck = **memory** · comm_fraction = **0.85 %** · comm_time = 0.192 ns · comm_bytes = 768 (per-core)

| core | compute | memory | comm | total | comm % |
|---:|---:|---:|---:|---:|---:|
| 0 | 6.016 | 16.384 | 0.192 | 22.592 | 0.85% |
| 1 | 6.016 | 8.192 | 0.192 | 14.400 | 1.33% |
| 2 | 6.016 | 8.192 | 0.192 | 14.400 | 1.33% |
| 3 | 6.016 | 8.192 | 0.192 | 14.400 | 1.33% |

### `ring_reduce_inner_loop` — all-reduce inside `scf.for` (`n_iters=3`) · output of the **ring_reduce_inner_loop cell**
issue126-comm-ring_reduce_inner_loop


critical core = 0 · bottleneck = **compute** · comm_fraction = **1.41 %** · comm_time = 0.576 ns · comm_bytes = 2304 (per-core)

| core | compute | memory | comm | total | comm % |
|---:|---:|---:|---:|---:|---:|
| 0 | 24.016 | 16.384 | 0.576 | 40.976 | 1.41% |
| 1 | 24.016 | 8.192 | 0.576 | 32.784 | 1.76% |
| 2 | 24.016 | 8.192 | 0.576 | 32.784 | 1.76% |
| 3 | 24.016 | 8.192 | 0.576 | 32.784 | 1.76% |

### Reading

- The view surfaces comm as a **distinct, non-zero slice** that used to print `0`. On these examples
  memory/compute dominate, so comm is a minority share — enough to validate the surfacing end-to-end, not a
  real-workload comm share.
- Nesting the all-reduce **in a loop scales comm with iterations**: `n_iters=3` triples the ring traffic
  (`comm_bytes` 768 → 2304, `comm` cycles 0.192 → 0.576), lifting `comm_fraction` 0.85 % → 1.41 %; the extra
  in-loop combiner work also flips the critical core's bottleneck from **memory** to **compute**.
- The write-back core (core 0) carries an extra store, so it is the critical core; its larger denominator is
  why its `comm_fraction` reads *lower* than the other cores'. `comm_fraction` is exposed comm on the
  wall-clock core — read it against the per-core table, not as the max comm share.

WarningRan:
**Hand-calc check (default config)** — the surfaced numbers are verifiable by hand, not a black box.

Constants (default `HardwareConfig`): clock 1 GHz (1 cycle = 1 ns); `ring_bytes_per_cycle = 4000 B/cyc`; `hbm_bytes_per_cycle_per_core = 31.25 B/cyc`; `simd = 64 elem/cyc`; 4 cores, each row `1×128 f16 = 256 B`; ring all-reduce hops = N−1 = 3; core 0 = write-back (load+store), cores 1–3 = load only.

**Critical core (core 0) components**

| component | formula | ring_reduce | inner_loop (n=3) |
|---|---|---|---|
| compute | #adds × 128/64 | 3-hop add × (128/64) = 6.016 | 3 iters × (3 combiner + 1 accum) × (128/64) = 24.016 |
| memory | (load+store) 256 B ÷ 31.25 | 2 × 8.192 = 16.384 | same (outside loop) = 16.384 |
| comm | comm_bytes ÷ 4000 | 0.192 | 0.576 |
| **total** | sum | **22.592** | **40.976** |

**Metrics**

| metric | formula | ring_reduce | inner_loop (n=3) |
|---|---|---|---|
| `comm_bytes` (per core) | hops × tile_bytes (× n) | 3×256 = 768 B | 3×256×3 = 2304 B |
| `comm_cycles` | comm_bytes ÷ ring_bytes_per_cycle | 768/4000 = 0.192 | 2304/4000 = 0.576 |
| `comm_time` | comm_cycles ÷ clock_ghz | 0.192 ns | 0.576 ns |
| `comm_fraction` | comm_cycles ÷ total_cycles (core 0) | 0.192/22.592 = **0.85%** | 0.576/40.976 = **1.41%** |
| `bottleneck` | argmax(compute, memory, comm) | memory | compute |

The loop's ×3 lands on comm and compute; memory is unchanged (load/store are outside the loop), which flips the bottleneck memory→compute. `comm_fraction` is taken on the critical (slowest) core — exposed communication — so it reads *lower* on core 0 (0.85%) than on cores 1–3 (1.33%, which lack the store); read it together with the per-core table.

yuhaohaoyu:
Two initial notes for the notebook text :

1. **Units not labeled in the per-core table.** The summary line shows `comm_time=0.576 ns` but the table columns (`compute`, `memory`, `comm`, `total`) are in cycles with no header annotation. At the default 1 GHz clock the numbers happen to match (1 cycle = 1 ns), which makes the output misleadingly consistent — a reader won't notice they're looking at different units until they change `clock_ghz`. 
   -  Fix: adding "(cycles)" to the table header or a brief note.

3. **Serial-sum assumption not stated.** The breakdown relies on `total_cycles = compute_cycles + memory_cycles + comm_cycles` (non-overlapping, fully serialized). This is documented in the PR description and hand-calc comment, but the notebook itself never states it. A reader seeing `comm_fraction=0.85%` has no way to know whether the three components are assumed serial (fractions sum to 1) or can overlap. 
   - Add a one-liner in the section 7 markdown, e.g. "The model assumes compute, memory, and comm are fully serialized (non-overlapping); `total_cycles = compute + memory + comm`."
   
   
I am staring the notebook now, will come up more.

